PR #44, opened 2024-04-16
Tat 117 amend git history that having api token

amybiyuliu:
### Description
Amend git history on main branch that having api token hardcoded

In [this commit](https://github.com/buildkite/test-splitter/pull/37/commits/15abe1a441b6dfd6a2fa1d1edeeb1f0b6582a2d2) I hardcoded a suite token BUILDKITE_SUITE_TOKEN
In a [later commit](https://github.com/buildkite/test-splitter/pull/37/commits/15ddf95f772231f92dadc68cb46fa592546a5815), I started to use ENV to replace the hardcoded token.

This PR is using interactive rebase approach to squash the impacted commits so that the hardcoded token won't show in the main branch git commit history.

* The test suite having the token has been deleted. The token is not valid anymore. 

### Context

https://linear.app/buildkite/issue/TAT-117/amend-git-history-that-having-api-token
### Changes

Amend git history on main such that hardcoded token can't be found in commit history


amybiyuliu:
^ the commits above looks odd as I become co-authoring all commits. I asked in #chat-engineering channel and #team-test-splitting channel. https://buildkite-corp.slack.com/archives/C05NAGJREJU/p1713248974384419
The suggestion is that we don't amend the git history as the token is already revoked.

PR #46, opened 2024-04-18
Implement test discovery and exclusion using glob pattern

nprizal:
### Description

Currently, we discover the test file by traversing inside the `spec` directory and perform operation on each file underneath it to check whether the file name ends with `_spec.rb`. If it matches, we append the file to an array, and return that array. The directory name and pattern are currently **hard coded** inside the function.

We need to implement a feature to exclude certain test files, to support the current usage of Buildkite/Buildkite pipeline to exclude test files within the `spec/features` directory. Additionally, we want to be able to configure the pattern used to find the test files (i.e. to override Rspec default pattern).

The implementation details are discussed in this [Basecamp post](https://3.basecamp.com/3453178/buckets/33707278/messages/7181426988).

### Context

https://linear.app/buildkite/issue/TAT-93/implement-test-discovery-and-exclusion-using-glob-pattern

### Changes

1. create `discoverTestFiles` function that use `zzglob` to discover test files using a glob pattern and exclude files that match exclude pattern.
1. add `discoveryPattern` function to `Rspec` struct that read environment variables `BUILDKITE_SPLITTER_PATTERN` & `BUILDKITE_SPLITTER_EXCLUDE_PATTERN`. If `BUILDKITE_SPLITTER_PATTERN` is missing, use fallback to the value of `spec/**/*_spec.rb`.
2. modify Rspec.GetFile to use `discoverTestFiles` using the pattern returned by `Rspec.discoveryPattern`.

### Testing

- I have done a manual test against Buildkite Rspec to include and exclude test files.

### DIFF
--- a/internal/runner/discover.go
+++ b/internal/runner/discover.go
@@ -15,12 +15,12 @@ type DiscoveryPattern struct {
 func discoverTestFiles(pattern DiscoveryPattern) ([]string, error) {
 	parsedPattern, err := zzglob.Parse(pattern.IncludePattern)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing pattern %q", pattern)
+		return nil, fmt.Errorf("error parsing test file pattern %q", pattern)
 	}
 
 	parsedExcludePattern, err := zzglob.Parse(pattern.ExcludePattern)
 	if err != nil {
-		return nil, fmt.Errorf("error parsing exclude pattern %q", pattern.ExcludePattern)
+		return nil, fmt.Errorf("error parsing test file exclude pattern %q", pattern.ExcludePattern)
 	}
 
 	discoveredFiles := []string{}
--- a/internal/runner/discover.go
+++ b/internal/runner/discover.go
@@ -1,0 +1,32 @@
+package runner
+
+import (
+	"fmt"
+	"io/fs"
+
+	"github.com/DrJosh9000/zzglob"
+)
+
+func discoverTestFiles(pattern string) []string {
+	parsedPattern, err := zzglob.Parse(pattern)
+	if err != nil {
+		fmt.Printf("Error parsing pattern: %v\n", err)
+	}
+
+	discoveredFiles := []string{}
+
+	// Use the Glob function to traverse the directory recursively
+	// and append the matched file paths to the discoveredFiles slice
+	parsedPattern.Glob(func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			fmt.Printf("Error walking: %v\n", err)
+		}
+		if d.IsDir() {
+			return nil
+		}
+		discoveredFiles = append(discoveredFiles, path)
+		return nil
+	}, nil)
+
+	return discoveredFiles
+}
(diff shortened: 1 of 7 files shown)


internal/runner/discover.go
```diff
@@ -0,0 +1,48 @@
...
+
+	discoveredFiles := []string{}
+
+	// Use the Glob function to traverse the directory recursively
+	// and append the matched file paths to the discoveredFiles slice
+	parsedPattern.Glob(func(path string, d fs.DirEntry, err error) error {
+		if err != nil {
+			fmt.Printf("Error walking: %v\n", err)
+		}
+		if d.IsDir() {
```
amybiyuliu, line 34: non-blocking question: what this does? is this one checking the directory doesn't have test files?

amybiyuliu (commented):
Looking good ✨ I'm wondering if we want to add the inculde/exclude pattern in README

internal/runner/discover.go
DrJosh9000: `Glob` works similarly to a filepath walk, which calls the callback both with matching files _and_ the intermediate directories. This provides a means to skip a particular directory for whatever reason.

DrJosh9000 (approved):
Really nice work!

internal/runner/discover.go
```diff
@@ -0,0 +1,48 @@
...
+
+type DiscoveryPattern struct {
+	IncludePattern string
+	ExcludePattern string
+}
+
+func discoverTestFiles(pattern DiscoveryPattern) ([]string, error) {
+	parsedPattern, err := zzglob.Parse(pattern.IncludePattern)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing pattern %q", pattern)
```
niceking, line 18:
```suggestion
		return nil, fmt.Errorf("error parsing test file pattern %q", pattern)
```

internal/runner/discover.go
```diff
@@ -0,0 +1,48 @@
...
+	}
+
+	parsedExcludePattern, err := zzglob.Parse(pattern.ExcludePattern)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing exclude pattern %q", pattern.ExcludePattern)
```
niceking, line 23:
```suggestion
		return nil, fmt.Errorf("error parsing test file exclude pattern %q", pattern.ExcludePattern)
```

niceking (commented):
Implementation looks good, just a few minor naming things. 

After reading this PR, I think the ENV vars should be called BUILDKITE_SPLITTER_TEST_FILE_PATTERN and BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN. It's a bit more verbose, but without the "test file" bit it's not really clear what the pattern is for, and while this is quite obvious to us as we're quite deep in the product, I don't think the less specific name will be as intuitive to users. What do you think?

My other comment is +1 to @amybiyuliu's suggestion that you add these ENV vars to the readme.

nprizal:
I've added both env variables to the README.

niceking (approved):
🚢

PR #47, opened 2024-04-18
Cleanup env vars

niceking:
### Description

Making changes to these ENV vars:

BUILDKITE_SUITE_TOKEN
A more appropriate name for this is probably BUILDKITE_SPLITTER_SUITE_TOKEN  and we should default this value to BUILDKITE_ANALYTICS_TOKEN as that's what we use in the collectors. 

BUILDKITE_BUILD_ID
We're using this as a unique identifier for test plans. We should call this value BUILDKITE_SPLITTER_IDENTIFIER instead so people can set their own if they wish, and default its value to BUILDKITE_BUILD_ID

Will also follow up with a PR to add the new BUILDKITE_SPLITTER_SUITE_TOKEN to the bk/bk pipeline definition. 

### Testing

Will test after this is released with PR with new ENV vars on bk/bk

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 The latest version of Buildkite Test Splitter can be downloaded from https://github.com/buildkite/test-splitter/releases
 
 ### Supported OS/Architecture
-ARM and AMD architecture for linux and darwin 
+ARM and AMD architecture for linux and darwin
 
 The available Go binaries
 - test-splitter-darwin-amd64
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -19,14 +19,18 @@ Please ensure that these default Buildkite ENV variables are available in the en
 - BUILDKITE_PARALLEL_JOB_COUNT
 - BUILDKITE_PARALLEL_JOB
 
-Additionally, we need the API token for the Test Suite that has the test data for the build. This will be available in the settings page for the Test Suite. We are expecting this key to be available as BUILDKITE_SUITE_TOKEN.
+Additionally, we need the API token for the Test Suite that has the test data for the build. If your pipeline only has one Test Suite, we can automatically detect this variable as BUILDKITE_ANALYTICS_TOKEN.
+
+However, if you have multiple test suites in your pipeline, you'll need to disambiguate these tokens by setting different values of BUILDKITE_SPLITTER_SUITE_TOKEN for each splitting job for each suite.
```
nprizal, line 24: I don't think this is accurate. If a pipeline has multiple test suites, they still need to be configured separately. I'm not sure if having `BUILDKITE_SPLITTER_SUITE_TOKEN` will make any different.

```
steps:
  - name: Unit test
    command: ./test-splitter
    env:
      - BUILDKITE_ANALYTICS_TOKEN: unit-test-suite-token-abc
      - BUILDKITE_SPLITTER_SUITE_TOKEN: unit-test-suite-token-abc
  - name: Feature test
    command: ./test-splitter
    env:
      - BUILDKITE_ANALYTICS_TOKEN: feature-test-suite-token-xyz
      - BUILDKITE_SPLITTER_SUITE_TOKEN: feature-test-suite-token-xyz
```

niceking: I'm not sure what you mean? In your example BUILDKITE_SPLITTER_SUITE_TOKEN is working as intended, as there are different values of the token for the features-on and features-off. It's what we've been done on the bk/bk pipeline for ages - it's simply renaming it from BUILDKITE_SUITE_TOKEN to BUILDKITE_SPLITTER_SUITE_TOKEN

The other part here is that we now default BUILDKITE_SPLITTER_SUITE_TOKEN to BUILDKITE_ANALYTICS_TOKEN. That means that for pipelines with only a single suite (not us), test splitter will work for them out of the box without them setting any additional env vars.

Let me know if you don't think this is clear in the readme or from my description?

niceking: PS @nprizal this is the customer feedback [here](https://buildkite-corp.slack.com/archives/C05NKQQTSGM/p1713322344820239) that led to this default value of BUILDKITE_ANALYTICS_TOKEN, it's being added for customers with pipelines much simpler than us, and nothing changes for Buildkite other than the rename of the env var, which I've got PR up for [here](https://github.com/buildkite/buildkite/pull/16329)

nprizal: Sorry, my comment was more about the README instead of renaming `BUILDKITE_SUITE_TOKEN` to `BUILDKITE_SPLITTER_SUITE_TOKEN`.

What I was saying is, even if a pipeline use multiple use suite tokens for the collectors, they need to be configured as `BUILDKITE_ANALYTICS_TOKEN` separately. So technically, it will still work out of the box without having to manually set the `BUILDKITE_SPLITTER_SUITE_TOKEN`.

This pipeline config will work as we expected, notice the token is different for `Unit test` and `Feature test`
```
steps:
  - name: Unit test
    command: ./test-splitter
    env:
      - BUILDKITE_ANALYTICS_TOKEN: unit-test-suite-token-abc
  - name: Feature test
    command: ./test-splitter
    env:
      - BUILDKITE_ANALYTICS_TOKEN: feature-test-suite-token-xyz
  ```

Comment (niceking) [reply README.md:24]: Yup, so that config above would work, and also this config would work:
```
steps:
  - name: Unit test
    command: ./test-splitter
    env:
      - BUILDKITE_SPLITTER_SUITE_TOKEN: unit-test-suite-token-abc
  - name: Feature test
    command: ./test-splitter
    env:
      - BUILDKITE_SPLITTER_SUITE_TOKEN: feature-test-suite-token-xyz
 ```
 
 I'm being deliberately vague in the README because I'm not presuming how they've set up their pipeline. On bk/bk we don't have a value set for BUILDKITE_ANALYTICS_TOKEN so I'm trying to make the docs not prescriptive to accomodate other pipeline set ups

niceking: Also - this is still in closed beta testing at the moment. I think its more valuable to ship the instructions and ask customers if they make sense rather than guess! There's still many stages before we decide on what the best copy for this is :)

nprizal: I saw your PR for the pipeline, and I think I understand. In our pipeline, if we don't set `BUILDKITE_SPLITTER_SUITE_TOKEN` it will not get the token, because we don't have `BUILDKITE_ANALYTICS_TOKEN` in the pipeline. Instead, we read `BUILDKITE_ANALYTICS_RSPEC_TOKEN` and set the token for the collector in `spec_helper.rb`.

So, I think `BUILDKITE_SPLITTER_SUITE_TOKEN` is useful when you don't set the token for collectors via `BUILDKITE_ANALYTICS_TOKEN` (like what we are doing). In my opinion, it has nothing to do with having multiple suites in a single pipeline.

niceking: Or is your comment more around we should *only* use BUILDKITE_ANALYTICS_TOKEN 🤔 My gut is to have the option to keep these two different as we evolve permissions. I don't know how long we're going to keep this pattern of BUILDKITE_ANALYTICS_TOKEN == access to both write and read from a suite.

nprizal: I think it's good that we have 2 options. As you said, not all pipeline will have `BUILDKITE_ANALYTICS_TOKEN` set up.

nprizal (approved):

PR #50, opened 2024-04-23
Allow configuration of Rspec command to be executed

amybiyuliu:
### Description
Currently, the Rspec command that will be executed by the test splitter is hard-coded as: 
`bin/rspec --options rspec.ci $test_files`
This command works well for bk/bk pipeline, however, beta customers might run Rspec with different commands and we should make the command configurable.

**Decisions:**

- We pass the whole test command as a new ENV variable to test-splitter
- We support interpolation of the test examples in the Rspec command. Example:
  
`BUILDKITE_TEST_SPLITTER_CMD=bundle exec rspec {{testExamples}}`

- If no test command is provided to the test-splitter, fallback to `bundle exec rspec $testExamples`

### Context
linear
https://linear.app/buildkite/issue/TAT-94/allow-configuration-of-rspec-command-to-be-executed

basecamp discussion
https://3.basecamp.com/3453178/buckets/33707278/messages/7318595635

### Testing
I tested the change locally. Here are the scenarios.
1. Provide test command without interpolation placeholder
`BUILDKITE_TEST_SPLITTER_CMD='bin/rspec --options rspec.ci' ./test-splitter ` ✅ tests running with customised command

2. Provide test command with interpolation placeholder
BUILDKITE_TEST_SPLITTER_CMD='bin/rspec --options rspec.ci {{testExamples}}' ./test-splitter ✅ tests running with customised command

3. Do not provide test command
./test-splitter ✅ tests running with default test command

4. Provide invalid test command ✅ error is handled 
```
BUILDKITE_TEST_SPLITTER_CMD='bin/rspecabdsdf --options rspec.ci' ./test-splitter
"bin/rspecabdsdf --options rspec.ci spec/features/clusters/sidenav_single_cluster_spec.rb spec/models/billing/plan/features_spec.rb spec/models/scm/pipeline_settings_spec.rb spec/workers/s3_copy_object_worker_spec.rb"
2024/05/01 15:49:01 Couldn't start tests: fork/exec bin/rspecabdsdf: no such file or directory
```
5. Provide invalid test command with single quote  ✅ error is handled 
```
BUILDKITE_TEST_SPLITTER_CMD="bin/rspec --options rspec.ci 'a  {{testExamples}}" ./test-splitter
Couldn't process test command "bin/rspec --options rspec.ci 'a  {{testExamples}}": Unterminated single-quoted string%
```

### DIFF
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -36,8 +36,7 @@ func (r Rspec) GetFiles() ([]string, error) {
 // Command returns an exec.Cmd that will run the rspec command
 func (r Rspec) Command(testCases []string) *exec.Cmd {
 	commandName, commandArgs := r.commandNameAndArgs(testCases)
-
-	fmt.Println(commandName, strings.Join(commandArgs, " "))
+	fmt.Printf("%q\n", commandName+" "+strings.Join(commandArgs, " "))
 
 	cmd := exec.Command(commandName, commandArgs...)
 	cmd.Stderr = os.Stderr
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -33,25 +33,12 @@ func (r Rspec) GetFiles() ([]string, error) {
 }
 
 // Command returns an exec.Cmd that will run the rspec command
-func (Rspec) Command(testCases []string, testCommandArgs []string) *exec.Cmd {
-	testCommand := "bundle exec rspec"
-	testArgs := testCases
+func (r Rspec) Command(testCases []string, commandLineArgs []string) *exec.Cmd {
+	commandName, commandArgs := r.commandNameAndArgs(testCases, commandLineArgs)
 
-	if len(testCommandArgs) > 0 {
-		index := -1
-		for i, item := range testCommandArgs {
-			if item == "{{testExample}}" {
-				index = i
-				break
-			}
-		}
-		testCommand = strings.Join(testCommandArgs[:index], " ")
-		testArgs = append(testArgs, testCommandArgs[index+1:]...)
-	}
+	fmt.Println(commandName, strings.Join(commandArgs, " "))
 
-	fmt.Println(testCommand, strings.Join(testArgs, " "))
-
-	cmd := exec.Command(testCommand, testArgs...)
+	cmd := exec.Command(commandName, commandArgs...)
 	cmd.Stderr = os.Stderr
 	cmd.Stdout = os.Stdout
 	return cmd
@@ -73,4 +60,40 @@ func (Rspec) discoveryPattern() DiscoveryPattern {
 		IncludePattern: includePattern,
 		ExcludePattern: excludePattern,
 	}
+}
+
+// commandNameAndArgs returns the command name and arguments to run the Rspec tests
+func (Rspec) commandNameAndArgs(testCases []string, commandLineArgs []string) (string, []string) {
+	commandName := ""
+	commandArgs := []string{}
+	// if commandLineArgs is not empty, using customized test command
+	if len(commandLineArgs) > 0 {
+		index := -1
+		for i, item := range commandLineArgs {
+			if strings.Compare(item, "{{testExamples}}") == 0 {
+				index = i
+				break
+			}
+		}
+		// Command name is the first element of the command line args
+		commandName = commandLineArgs[0]
+		// The rest of the command line args are the arguments of the test command
+		// if {{testExamples}} is found, replace it with testCases
+		// otherwise, append testCases to the end of the command args
+		if index != -1 {
+			commandArgs = append(commandArgs, commandLineArgs[1:index]...)
+			commandArgs = append(commandArgs, testCases...)
+			commandArgs = append(commandArgs, commandLineArgs[index+1:]...)
+		} else {
+			commandArgs = append(commandArgs, commandLineArgs[1:]...)
+			commandArgs = append(commandArgs, testCases...)
+		}
+	} else {
+		// The default Rspec command `bundle exec rspec ${testCases}`
+		commandName = "bundle"
+		commandArgs = append(commandArgs, "exec", "rspec")
+		commandArgs = append(commandArgs, testCases...)
+	}
+
+	return commandName, commandArgs
 }
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
+	"slices"
 	"strings"
 
 	"github.com/kballard/go-shellquote"
@@ -68,18 +69,15 @@ func (Rspec) discoveryPattern() DiscoveryPattern {
 
 // commandNameAndArgs returns the command name and arguments to run the Rspec tests
 func (Rspec) commandNameAndArgs(testCases []string, testCommand string) (string, []string, error) {
-	completedTestCommand := ""
-	if strings.Contains(testCommand, "{{testExamples}}") {
-		completedTestCommand = strings.ReplaceAll(testCommand, "{{testExamples}}", shellquote.Join(testCases...))
-	} else {
-		completedTestCommand = testCommand + " " + shellquote.Join(testCases...)
-	}
-
-	testCommandFields, err := shellquote.Split(completedTestCommand)
-
+	words, err := shellquote.Split(testCommand)
 	if err != nil {
 		return "", []string{}, err
 	}
-
-	return testCommandFields[0], testCommandFields[1:], nil
+	idx := slices.Index(words, "{{testExamples}}")
+	if idx < 0 {
+		words = append(words, testCases...)
+		return words[0], words[1:], nil
+	}
+	words = slices.Replace(words, idx, idx+1, testCases...)
+	return words[0], words[1:], nil
 }
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"slices"
 	"strings"
 )
 
@@ -34,8 +33,8 @@ func (r Rspec) GetFiles() ([]string, error) {
 }
 
 // Command returns an exec.Cmd that will run the rspec command
-func (r Rspec) Command(testCases []string) *exec.Cmd {
-	commandName, commandArgs := r.commandNameAndArgs(testCases)
+func (r Rspec) Command(testCases []string, testCommand string) *exec.Cmd {
+	commandName, commandArgs := r.commandNameAndArgs(testCases, testCommand)
 	fmt.Printf("%q\n", commandName+" "+strings.Join(commandArgs, " "))
 
 	cmd := exec.Command(commandName, commandArgs...)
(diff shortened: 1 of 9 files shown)


main.go
```diff
@@ -24,7 +24,8 @@ import (
 func main() {
 	// TODO: detect test runner and use appropriate runner
 	testRunner := runner.Rspec{}
-
+	// Get the arguments without the program name
+	argsWithoutProg := os.Args[1:]
```
DrJosh9000, line 28: This gets all args, which includes flags (e.g. if someone uses the `--files` flag then this will contain `"--files"` somewhere). `flag.Args()` returns the program args not including flags, which you will need to call after `flag.Parse()`.

amybiyuliu: Hey @DrJosh9000 thanks for calling this out
We are going to stop supporting the `--files` flags as we are not using it anymore. 
I'm wondering if we remove the support of `--files` flag, do I still need to use `flag.Args()` for the program args?

DrJosh9000: I think you shouldn't mix and match, and you should either:

* use `flag.Args()` in this PR, and replace it with `os.Args` when you get rid of `flag` in a later PR

or

* get rid of `flag` and change to `os.Args` in this PR.

amybiyuliu: Thanks Josh, I'll get rid of flag and change to os.Args in this PR. I'll also update the PR description

niceking:
Hey @amybiyuliu I think there's been some confusion about what you were supposed to be building? Perhaps it wasn't clear from the [basecamp post](https://3.basecamp.com/3453178/buckets/33707278/messages/7318595635) 😅 

The intention of the feature is to introduce a new environment variable e.g. BUILDKITE_TEST_SPLITTER_CMD or similar name, which contains a string that we'll use to execute the test command. If none is provided, we fallback to "bundle exec rspec {{testCases}}".

It might clearer to think about how this feature will be used? To take bk/bk as an example - once this feature is built, we'll need to add in our [pipeline.yml](https://github.com/buildkite/buildkite/blob/6d2ba6e856ff066247e4ffd88bf5a0aae71cc7a1/.buildkite/pipeline.yml#L233) a new env var called BUILDKITE_TEST_SPLITTER_CMD="bin/rspec --options .rspec.ci {{testCases}}". Or possibly export that env var at the top of the [splitter spec](https://github.com/buildkite/buildkite/blob/6d2ba6e856ff066247e4ffd88bf5a0aae71cc7a1/.buildkite/steps/splitter_specs#L3) file.

Let me know if this isn't clear and we can have a zoom!

niceking:
So the configurable rspec cmd isn't supposed to be a command line argument to the splitter client, it's supposed to be an ENV variable :)

internal/runner/rspec.go
```diff
@@ -33,14 +33,12 @@ func (r Rspec) GetFiles() ([]string, error) {
 }
 
 // Command returns an exec.Cmd that will run the rspec command
-func (Rspec) Command(testCases []string) *exec.Cmd {
-	args := []string{"--options", ".rspec.ci"}
+func (r Rspec) Command(testCases []string, commandLineArgs []string) *exec.Cmd {
+	commandName, commandArgs := r.commandNameAndArgs(testCases, commandLineArgs)
 
-	args = append(args, testCases...)
+	fmt.Println(commandName, strings.Join(commandArgs, " "))
```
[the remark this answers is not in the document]
amybiyuliu: I've updated this line. Let me know if I'm not doing the right thing!

internal/runner/rspec.go
```diff
@@ -63,3 +61,36 @@ func (Rspec) discoveryPattern() DiscoveryPattern {
 		ExcludePattern: excludePattern,
 	}
 }
+
+// commandNameAndArgs returns the command name and arguments to run the Rspec tests
+func (Rspec) commandNameAndArgs(testCases []string) (string, []string) {
```
niceking, line 66: At first read, this implementation was much more complicated than I thought that it would be. Two questions: 

1. Is this slicing necessary? Could we do a straight up string substitution instead? i.e. something like
`cmd = string.ReplaceAll(testCommand, "{{testCases}}", testCases...`

2. If this slicing *is* necessary, I'm finding the naming of the arguments a little confusing. The command seems unnecessarily split into "commandName" and "commandArgs"? Especially when it doesn't really make sense to think of "bundle" as the "commandName" and "exec rspec" as the "commandArgs". I think "cmdPt1" and "cmdPt2" might serve as better names? 

But I think its possible to just pass a single command argument into exec.Command - is there a reason we're not just doing that instead? 

cc @DrJosh9000

niceking:
I had a chat with @mitchbne in the morning which made me think we shouldn't delete the "files" flag just yet. Could you remove that deletion from this PR until we decide what to do with that feature?

niceking:
Good testing on this PR!

internal/runner/rspec.go
amybiyuliu: > Could we do a straight up string substitution instead?

Good call, I think we should be able to simplify the replacement of {{testCases}} with ReplaceAll.

> But also I think its possible to just pass a single command argument into exec.Command - is there a reason we're not just doing that instead

I also thought about this and attempted to pass just one arg `exec.Command(args...)`. It shows error that I'm not providing enough arguments. It looks like I have to follow the definition `func Command(name string, arg ...string) *Cmd`

niceking: Ahhhh I see, and I also had a more thorough read of the docs for the [command](https://pkg.go.dev/os/exec#Command) function. 

For the command, do we have to be strictly observant of which parts are the "command" and what parts are the "args"? Because strictly speaking, for Rspec the command is "bundle exec" or does it not matter as long as both parts add up to a sensible full command? Just wondering because there's a note in the docs about parsing compatibility with Windows.

In that case, perhaps everything before {{testCases}} is the command? And {{testCases}} and whatever follows is the argument? Which is not too different from your original implementation 😂 Just a case of whether the first word is the command or not 🤔

amybiyuliu: I'm not sure how exec.command works in go. But I tried to use `bundle exec` as the command name and Go isn't able to resolve this path. 
If we have an rspec command `bundle exec rspec test_spec.rb`, I think it needs to look like this for 
 go to understand `exec.Command("bundle", "exec" , "rspec", "test_spec.rb")`

niceking: Ok sweet thanks for explaining that to me and also running that code 😁

niceking (commented):
Looks good - pending removal of the removal of the "files" flag from the PR!

niceking (approved):
Nice, this LGTM but I think worth waiting for a review from @DrJosh9000 in case I completely steered you wrong 😂

DrJosh9000 (changes requested):

internal/runner/rspec.go
```diff
@@ -63,3 +62,22 @@ func (Rspec) discoveryPattern() DiscoveryPattern {
...
+		completedTestCommand = strings.ReplaceAll(testCommand, "{{testExamples}}", strings.Join(testCases, " "))
+	} else {
+		completedTestCommand = testCommand + " " + shellquote.Join(testCases...)
+	}
+
+	testCommandFields, err := shellquote.Split(completedTestCommand)
+
+	if err != nil {
+		fmt.Printf("Couldn't process test command %q: %v", testCommand, err)
+		os.Exit(3)
```
amybiyuliu, line 79: We want to rerun rspec tests when having rspec failure (Rspec failure will exit with exit code 1). I'm using error code 3 here so we can differentiate rspec failure and splitter client failure.

internal/runner/rspec.go
```diff
@@ -63,3 +65,21 @@ func (Rspec) discoveryPattern() DiscoveryPattern {
...
+// commandNameAndArgs returns the command name and arguments to run the Rspec tests
+func (Rspec) commandNameAndArgs(testCases []string, testCommand string) (string, []string, error) {
+	completedTestCommand := ""
+	if strings.Contains(testCommand, "{{testExamples}}") {
+		completedTestCommand = strings.ReplaceAll(testCommand, "{{testExamples}}", shellquote.Join(testCases...))
+	} else {
+		completedTestCommand = testCommand + " " + shellquote.Join(testCases...)
+	}
+
+	testCommandFields, err := shellquote.Split(completedTestCommand)
```
DrJosh9000, line 78: Joining the command together to search for the string only to then split it again seems unnecessary and possibly still error-prone to me. One of the words from `shellquote.Split` should be `{{testExamples}}`, because it takes care of de-quoting each word, whether the user wrote the test command with `{{testExamples}}` or `'{{testExamples}}'` or `"{{testExamples}}"`.

What about:

```suggestion
	words, err := shellquote.Split(completedTestCommand)
	if err != nil {
		return "", []string{}, err
	}
	idx := slices.Index(words, "{{testExamples}}") 
	if idx < 0 {
		words = append(words, testCases...)
		return words[0], words[1:], nil		
	}
	words = slices.Replace(words, idx, idx+1, testCases...)
	return words[0], words[1:], nil	
```

amybiyuliu: Good call! Joining the string and splitting the results does feel unnecessary and error-prone! 😅

internal/runner/rspec_test.go
```diff
@@ -1,12 +1,71 @@
...
 	"github.com/google/go-cmp/cmp"
+	"github.com/kballard/go-shellquote"
 )
 
+func TestCommandNameAndArgs_WithInterpolationPlaceholder(t *testing.T) {
+	rspec := Rspec{}
+	testCases := []string{"spec/models/user_spec.rb", "spec/models/billing_spec.rb"}
+	testCommand := "bin/rspec --options {{testExamples}} --format"
+
+	gotName, gotArgs, _ := rspec.commandNameAndArgs(testCases, testCommand)
```
DrJosh9000, line 17: Check the error return (mandatory), and ideally log the args to the function (highly recommended). This will help cut down investigation time when diagnosing possible test failures.

```suggestion
	gotName, gotArgs, err := rspec.commandNameAndArgs(testCases, testCommand)
	if err != nil {
		t.Errorf("Rspec.commandNameAndArgs(%q, %q) error = %v", testCases, testCommand, err)
	}
```

internal/runner/rspec_test.go
```diff
@@ -1,12 +1,71 @@
...
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
+	}
+}
+
+func TestCommandNameAndArgs_WithoutInterpolationPlaceholder(t *testing.T) {
+	rspec := Rspec{}
+	testCases := []string{"spec/models/user_spec.rb", "spec/models/billing_spec.rb"}
+	testCommand := "bin/rspec --options --format"
+
+	gotName, gotArgs, _ := rspec.commandNameAndArgs(testCases, testCommand)
```
DrJosh9000, line 35: As above

```suggestion
	gotName, gotArgs, err := rspec.commandNameAndArgs(testCases, testCommand)
	if err != nil {
		t.Errorf("Rspec.commandNameAndArgs(%q, %q) error = %v", testCases, testCommand, err)
	}
```

internal/runner/rspec_test.go
```diff
@@ -1,12 +1,71 @@
...
+
+	wantName := "bin/rspec"
+	wantArgs := []string{"--options", "spec/models/user_spec.rb", "spec/models/billing_spec.rb", "--format"}
+
+	if diff := cmp.Diff(gotName, wantName); diff != "" {
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
+	}
+	if diff := cmp.Diff(gotArgs, wantArgs); diff != "" {
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
```
DrJosh9000, line 26:
```suggestion
		t.Errorf("Rspec.commandNameAndArgs(%q, %q) diff (-got +want):\n%s", testCases, testCommand, diff)
	}
	if diff := cmp.Diff(gotArgs, wantArgs); diff != "" {
		t.Errorf("Rspec.commandNameAndArgs(%q, %q) diff (-got +want):\n%s", testCases, testCommand, diff)
```

internal/runner/rspec_test.go
```diff
@@ -1,12 +1,71 @@
...
+	gotName, gotArgs, _ := rspec.commandNameAndArgs(testCases, testCommand)
+
+	wantName := "bin/rspec"
+	wantArgs := []string{"--options", "--format", "spec/models/user_spec.rb", "spec/models/billing_spec.rb"}
+
+	if diff := cmp.Diff(gotName, wantName); diff != "" {
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
+	}
+	if diff := cmp.Diff(gotArgs, wantArgs); diff != "" {
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
```
DrJosh9000, line 44:
```suggestion
		t.Errorf("Rspec.commandNameAndArgs(%q, %q) diff (-got +want):\n%s", testCases, testCommand, diff)
	}
	if diff := cmp.Diff(gotArgs, wantArgs); diff != "" {
		t.Errorf("Rspec.commandNameAndArgs(%q, %q) diff (-got +want):\n%s", testCases, testCommand, diff)
```

internal/runner/rspec_test.go
```diff
@@ -1,12 +1,71 @@
...
+
+	wantName := ""
+	wantArgs := []string{}
+
+	if diff := cmp.Diff(gotName, wantName); diff != "" {
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
+	}
+	if diff := cmp.Diff(gotArgs, wantArgs); diff != "" {
+		t.Errorf("Rspec.commandNameAndArgs() diff (-got +want):\n%s", diff)
+	}
```
DrJosh9000, line 63: If the intention is to disregard the other return values when `err != nil`, then there's not much value in testing that the other return values are empty. But this is fine.

DrJosh9000 (approved):
Good work! Some final comments from me for you.

internal/runner/rspec.go
```diff
@@ -33,17 +36,17 @@ func (r Rspec) GetFiles() ([]string, error) {
...
-
-	args = append(args, testCases...)
-
-	fmt.Println("bin/rspec", strings.Join(args, " "))
+func (r Rspec) Command(testCases []string, testCommand string) (*exec.Cmd, error) {
+	commandName, commandArgs, err := r.commandNameAndArgs(testCases, testCommand)
+	if err != nil {
+		return nil, err
+	}
+	fmt.Printf("%q\n", commandName+" "+strings.Join(commandArgs, " "))
```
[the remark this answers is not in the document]
amybiyuliu: Good catch! We can totally use shellquote.Join here!!

internal/runner/rspec_test.go
amybiyuliu: ohh wow I totally forgot to check error return for those tests 😅 Thanks for calling it out! I'll apply the changes!

PR #53, opened 2024-05-03
Return specific exit code from Test Splitter Client

amybiyuliu:
### Description
Currently the test splitter exit with exit code 1 when there's a Rspec test failure. The test splitter also exits with exit code 1 when there are other errors (eg. config error). 

In the bk/bk pipeline setup, we rerun failed tests when the test splitter exits with non-zero exit code. As both Rspec failure and other test splitter failures exit with 1, we can't differentiate them. 

In this PR, the test splitter will return exit code 3 and prints error message if it's not Rspec test failure

### Context
Linear ticket:
https://linear.app/buildkite/issue/TAT-109/return-specific-exit-code-from-test-splitter-client
Additional context on slack:
TLDR: It's suggested we pick a number from 3-31 for the exit code. I checked bk/bk pipeline setup and 3 is not being used
https://buildkite-corp.slack.com/archives/C05NKQQTSGM/p1714633465870209

### Changes
Adding a helper function to print error message and exit with custom exit code

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,5 +41,9 @@ Otherwise, your script for executing specs may look something like:
 chmod +x test-splitter
 ./test-splitter # fetches the test plan for this node, and then executes the rspec tests
 ```
-
-
+### Exit code
+| Exit code | Description |
+| ---- | ---- |
+| 0 | All tests pass |
+| 1 | At least one test fails |
+| 3 | Test Splitter failure (eg. config error) |
--- a/main.go
+++ b/main.go
@@ -7,7 +7,6 @@ import (
 	"errors"
 	"flag"
 	"fmt"
-	"log"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -35,7 +34,7 @@ func main() {
 	} else {
 		fs, err := testRunner.GetFiles()
 		if err != nil {
-			log.Fatalf("Couldn't get files: %v", err)
+			logErrorAndExit(3, "Couldn't get files: %v", err)
 		}
 		files = fs
 	}
@@ -43,7 +42,7 @@ func main() {
 	// get config
 	cfg, err := config.New()
 	if err != nil {
-		log.Fatal("Invalid configuration: ", err)
+		logErrorAndExit(3, "Invalid configuration: %v", err)
 	}
 
 	// get plan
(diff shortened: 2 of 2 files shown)


wooly:
Do we need to update the docs to document the error code being not-one?

DrJosh9000 (approved):
LGTM for code

wooly (approved):

README.md
```diff
@@ -41,5 +41,9 @@ Otherwise, your script for executing specs may look something like:
 chmod +x test-splitter
 ./test-splitter # fetches the test plan for this node, and then executes the rspec tests
 ```
-
-
+### Exit code
+| Exit code | Description |
+| ---- | ---- |
+| 0 | All tests pass |
+| 1 | At least one test fails |
```
pda, line 48: Do `0`/`1` specifically mean exactly these things? Or do they mean whatever the underlying test runner means?

Could it be expressed like this?

- `0`: success (passed through from test runner)
- `1`: failure (passed through from test runner)
- `3`: Test Splitter failure (eg. config error)
- `*`: other errors (passed through from the test runner)

(I guess for now we have a good understanding of RSpec and its exit statuses, but that might be quite as true for future supported test runners)

Given that the above list is basically “exit codes are always passed through from the test runner, but there's at least one special one that we generate for internal problems”, I wonder if we should reserve a range of not-commonly-used exit statuses for our internal use, e.g. 16–23.

pda (commented):
👍🏼

README.md
wooly: This comment sounds reasonable, yep. The 16–23 exit code reservation would include changing the current `3` code to `16` is that right?

amybiyuliu: Yes, we run the rspec tests in a subprocess and exit the main program the same code returned by the subprocess. 0, 1 and other errors are passed through from the test runner. The only special case is the test splitter internal errors (currently 3).

I'll create a PR to improve the README, and update the exit code for internal failure to 16. 

I'm also curious why we pick 16-23? I'm assuming exit codes below 16 could be used commonly for other exit reasons?

wooly: @pda any reasoning behind the 16–23?

Issue #52, opened 2024-05-02
Exclude pattern parameter doesn't seem to work as documented


TJC:
Hi,
I'm trying to use the test-splitter on a project that requires some unit tests to run on a different set of agents.

These use Ruby Rspec, and I could potentially use either rspec tags, or simple file paths, to pick them out. I'm trying to use the `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` parameter to achieve this, but it doesn't seem to be working for me.

For example, I am trying this command, but finding it still runs rspecs from within that directory.

`BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN=spec/acceptance/** test-splitter`

Is this the right way to do this?

Do I need to change the include pattern to be very minimal, perhaps?

TJC:
Update: I think we're running an older version of the test-splitter. I'll update and re-test, and re-open this issue if needs be.

amybiyuliu:
Hi @TJC I'm Amy from Buildkite 👋 Thanks for raising the issue, and please let us know if you have more questions setting up the test splitter!

niceking:
Hey @TJC! Perhaps bumping the version of the Test Splitter will help 😂 If not, it might be how you're specifying your exclude pattern? We expect the exclude pattern to take the form of a file glob, not a directory path as you're probably used to doing with Rspec! 

If you want to exclude all the acceptance specs, you'll need to do something like:
`BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN=spec/acceptance/**/*_spec.rb`

TJC:
@niceking You mentioned that you expect the pattern to take the form of a file glob, not a directory. However the documentation actually states otherwise:

> Optional, glob pattern to use for excluding test files or directory.

niceking:
Thanks for the feedback @TJC, I've made a PR for this here: https://github.com/buildkite/test-splitter/pull/54

Does that help clarify what we expect? Also, you might have already seen this, but we also link out to the glob library that we use to interpret the given globs: 
<img width="833" alt="Screenshot 2024-05-06 at 1 26 18 PM" src="https://github.com/buildkite/test-splitter/assets/4241125/ca7a6d3a-a6b3-4c32-95a4-8edb5fc0bf42">

Let me me know here or on the PR if you think the wording could be clearer 😄

PR #54, opened 2024-05-06
Make glob requirements clearer

niceking:
### Description

As raised by @TJC in https://github.com/buildkite/test-splitter/issues/52, the phrasing of what we expect the glob to be could be clearer

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,8 +21,8 @@ The available Go binaries
 |  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
 | `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
 |  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID` | Required, use default unless you have a specific use case |
-| `BUILDKITE_SPLITTER_TEST_FILE_PATTERN` | `spec/**/*_spec.rb` | Optional, glob pattern for discovering test files that need to be executed. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library*. |
-| `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` | - | Optional, glob pattern to use for excluding test files or directory. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library.* |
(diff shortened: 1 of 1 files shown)


wooly (approved):

amybiyuliu (approved):

README.md
```diff
@@ -21,8 +21,8 @@ The available Go binaries
...
-| `BUILDKITE_SPLITTER_TEST_FILE_PATTERN` | `spec/**/*_spec.rb` | Optional, glob pattern for discovering test files that need to be executed. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library*. |
-| `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` | - | Optional, glob pattern to use for excluding test files or directory. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library.* |
+| `BUILDKITE_SPLITTER_TEST_FILE_PATTERN` | `spec/**/*_spec.rb` | Optional, file glob pattern for discovering test files that need to be executed. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library*. |
+| `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` | - | Optional, file glob pattern to use for excluding test files or directory. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library.* |
```
TJC, line 25: It says `for excluding test files or directory.` -- so can I use the pattern to exclude a whole directory tree? You suggested in #52 that it had to match files.

nprizal: Hi @TJC, yes you can use `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` to exclude a whole directory, providing you use the correct globing pattern syntax. The issue you were having in #52 seems to be caused by the issue in the identifier that you raised in #55.

Can you confirm that the exclude pattern is now working on your side?

TJC: I was referring to the comment in https://github.com/buildkite/test-splitter/issues/52#issuecomment-2091992853 which said "We expect the exclude pattern to take the form of a file glob, not a directory path as you're probably used to doing with Rspec!"

The description you're changing in the PR now says "file glob pattern to use for excluding test files or directory." 
That seems to imply you could pass a glob that matched a directory.

So perhaps you can see where this documentation isn't actually that clear about whether it accepts a directory, or if you actually need to be creating a glob that matches all files within a directory. Based on our discussions I think the answer is "a glob that matches all files within a directory", but to a fresh reader who comes to this project, I don't think that would be clear.

TJC: There's also been #58 merged since this PR was opened, which I think has changed how it works?

wooly:
I think this has been rendered unnecessary by the change that @DrJosh9000 made to the zzglob library and subsequent 0.4.0 release of the splitter, so I'm going to close this.

PR #58, opened 2024-05-13
Support exclude pattern that is a specific directory

DrJosh9000:
### Description

The exclusion pattern can now be a specific directory (or glob that only matches directories).

### Context

Relates to #54 and #46

### Changes

- Update zzglob to v0.2.0 to support new option
- Pass the new option
- Change the callback logic to return `fs.SkipDir` when the exclusion pattern matches a directory
- Add a test

### Testing

See new unit test.

### DIFF
--- a/internal/runner/discover_test.go
+++ b/internal/runner/discover_test.go
@@ -52,3 +52,26 @@ func TestDiscoverTestFiles_WithExcludePattern(t *testing.T) {
 		t.Errorf("discoverTestFiles(%q, %q) diff (-got +want):\n%s", pattern, excludePattern, diff)
 	}
 }
+
+func TestDiscoverTestFiles_WithExcludeDirectory(t *testing.T) {
+	pattern := "fixtures/**/*_test"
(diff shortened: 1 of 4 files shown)


wooly:
LGTM!

wooly (approved):

internal/runner/discover_test.go
```diff
@@ -52,3 +52,26 @@ func TestDiscoverTestFiles_WithExcludePattern(t *testing.T) {
 		t.Errorf("discoverTestFiles(%q, %q) diff (-got +want):\n%s", pattern, excludePattern, diff)
 	}
 }
+
+func TestDiscoverTestFiles_WithExcludeDirectory(t *testing.T) {
+	pattern := "fixtures/**/*_test"
+	excludePattern := "fixtures/animals"
```
nprizal, line 58: 🏆🏆🏆

nprizal (approved):
This is neat!

PR #59, opened 2024-05-13
Update IDENTIFIER default to BUILD_ID + STEP_ID

amybiyuliu:
### Description
Current default BUILDKITE_SPLITTER_IDENTIFIER is the BUILDKITE_BUILD_ID. With this default setup, the test splitter doesn't work correctly when there are multiple test groups in one build. 

For more context: 
A beta customer try to apply different exclude/include patterns at different steps with the same identifier (BUILDKITE_BUILD_ID), and released the exclude/include patterns wasn't applied. This is due to we cache plan for same identifier. https://github.com/buildkite/test-splitter/issues/55

This PR updates the default identifier to BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID, and update the README.

### Context

https://linear.app/buildkite/issue/TAT-138/update-buildkite-splitter-identifier-default-to-build-id-step-id

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,10 +20,10 @@ The available Go binaries
 |  `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism |
 |  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
 | `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
-|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Required, use default unless you have a specific use case |
+|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Optional. Test Splitter uses the identifier to store and fetch the test plan and must be unique for each build and steps group. By default it will use a composite of `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, but it can be overridden by specifying the `BUILDKITE_SPLITTER_IDENTIFIER`. `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` must be accessible by the client when using the default. |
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -20,7 +20,7 @@ The available Go binaries
 |  `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism |
 |  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
 | `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
-|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID` | Required, use default unless you have a specific use case |
+|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Required, use default unless you have a specific use case |
```
wooly, line 23: By making this default to a compound identifier based on the Build ID and Step ID, I don't think this is 'Required' any more as suggested by the description, so we perhaps need to wordsmith a better description for this one.

There are two things that stand out to me:
1. The use of the `BUILDKITE_SPLITTER_IDENTIFIER` can be used to override the identifier used by the test-splitter.
2. If the `BUILDKITE_SPLITTER_IDENTIFIER` environment variable is not overridden, the test splitter _must_ have access to the `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` environment variables to create the compound identifier.

wooly:
Code looks good, one comment on docs :+1:

README.md
nprizal: This is a good point 💯. The identifier was never "required" by the client because it will set a default value, but the user has to pass the `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`. We might also need to explain the usage, and how the test splitter use the identifier. 

Maybe something along these lines?
> Optional. Test Splitter use identifier to store and fetch the test plan and must be unique for each build and steps group. By default it will use a composite of `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, but it can be overriden by specifying the `BUILDKITE_SPLITTER_IDENTIFIER`. `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` must be accessible by the client when using the default.

amybiyuliu: Good catch! It makes sense to me that the identifier shouldn't be required. 
I've updated the description based on @nprizal's suggestion.

wooly: Yeah looks good :+1:

wooly (approved):

PR #60, opened 2024-05-13
Refactor API package

nprizal:
### Description

We will be implementing split by example that includes a change on how the client communicate with the server.
With the new changes, the client will call the server's new endpoint to fetch the test plan first, and make a subsequent request to create a test plan if the test plan hasn't been created. The pseudocode for the changes looks something like this:

```
// main.go
func fetchOrCreatePlan() {
  plan := api.FetchPlan()

  if plan != nil {
    return plan
  }

  if plan != nil {
    return plan
  }
}
```

Currently, we have `FetchTestPlan` function in the `api` package that creates the test plan. Since we want to handle fetch and create separately, we need to rename the existing function into `CreateTestPlan`. Also, because fetch and create function will share the same concerns (e.g. url, token, organization slug), I created a `client` struct in the `api` that will hold the information so we don't need to pass it twice.

This PR is a prep for further work to separate fetch and create of test plan.

### Context

https://linear.app/buildkite/issue/TAT-143/fetch-plan-from-the-cache-before-making-a-post-request-to-create-one

### Changes

- Rename `FetchTestPlan` to `CreateTestPlan`
- Create `client` struct
- Move `FetchTestPlan` into a function for the `client` struct

### DIFF
--- a/internal/api/create_test_plan_test.go
+++ b/internal/api/create_test_plan_test.go
@@ -15,7 +15,7 @@ import (
 
 func ptr[T any](t T) *T { return &t }
 
-func TestFetchTestPlan(t *testing.T) {
+func TestCreateTestPlan(t *testing.T) {
 	svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
 		fmt.Fprint(w, `{
 	"tasks": {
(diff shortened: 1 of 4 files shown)


internal/api/create_test_plan_test.go
```diff
@@ -97,15 +104,19 @@ func TestFetchTestPlan_Timeout(t *testing.T) {
 	defer cancel()
 
 	params := TestPlanParams{}
-	got, err := FetchTestPlan(fetchCtx, svr.URL, params)
+	apiClient := client{
+		ServerBaseUrl: svr.URL,
```
nprizal, line 108: Later, we will add API access token, organization slug, etc

PR #63, opened 2024-05-14
Use new API endpoint to fetch and create the test plan

nprizal:
### ⚠️ BREAKING CHANGES
These changes introduced 3 required configs
- `BUILDKITE_API_ACCESS_TOKEN`. You now need Buildkite API access token that has `read_suites`, `read_test_plan`, and `write_test_plan` scopes.
- `BUILDKITE_ORGANIZATION_SLUG`. This is available in your Buildkite Pipeline environment, so you don't need to set it manually.
- `BUILDKITE_SPLITTER_SUITE_SLUG`. This replaces `BUILDKITE_SPLITTER_SUITE_TOKEN` and need to be set manually.

### Description

As part of split by example implementation, we want to change how the client fetch and create test plan from the server. The new flow will look like this
```
// main.go
func fetchOrCreatePlan() {
  plan := api.FetchPlan()

  if plan != nil {
    return plan
  }

  if plan != nil {
    return plan
  }
}
```

This PR also replaced the endpoint to the new endpoint.

### Changes

- Add 3 new configs
- Add `httpClient` to `api.client` struct that has Transport to attach access token to each request
- Implement `FetchTestPlan` in the `api.client` struct
- Replace endpoint in `CreateTestPlan` to the new endpoint
- Removed `BUILDKITE_SPLITTER_SUITE_TOKEN` from config

### Testing

As this is a breaking change, I suggest testing it manually against a local Buildkite server (or Prod). I tested it manually against my local Buildkite server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,8 +17,8 @@ The available Go binaries
 
 | Environment Variable | Default Value | Description |
 | ---- | ---- | ----------- |
-| `BUILDKITE_API_ACCESS_TOKEN ` | - | Required, Buildkite API access token with `read_suites`, `read_test_plan`, and `write_test_plan` scopes. |
-| `BUILDKITE_ORGANIZATION_SLUG` | - | Required, the slug of your Buildkite organization. |
+| `BUILDKITE_API_ACCESS_TOKEN ` | - | Required, Buildkite API access token with `read_suites`, `read_test_plan`, and `write_test_plan` scopes. You can create access token from [Personal Settings](https://buildkite.com/user/api-access-tokens) in Buildkite |
+| `BUILDKITE_ORGANIZATION_SLUG` | - | Required, the slug of your Buildkite organization. This is available in your pipeline environment, so you don't need to set it manually |
 | `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism. |
 | `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
 | `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Optional. Test Splitter uses the identifier to store and fetch the test plan and must be unique for each build and steps group. By default it will use a composite of `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, but it can be overridden by specifying the `BUILDKITE_SPLITTER_IDENTIFIER`. `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` must be accessible by the client when using the default. |
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ The available Go binaries
 
 For most use cases, Test Splitter should work out of the box due to the default values available from your Buildkite environment.
 
-However, you'll need to set `BUILDKITE_SPLITTER_SUITE_TOKEN` if your test collector doesn't use the `BUILDKITE_ANALYTICS_TOKEN` value, or your pipeline has multiple suites.
+However, you'll have to set `BUILDKITE_API_ACCESS_TOKEN` and `BUILDKITE_SPLITTER_SUITE_SLUG`.
 
 You can also set the `BUILDKITE_SPLITTER_TEST_FILE_PATTERN` or `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` if you need to filter the tests selected for execution.
 
--- a/README.md
+++ b/README.md
@@ -17,12 +17,14 @@ The available Go binaries
 
 | Environment Variable | Default Value | Description |
 | ---- | ---- | ----------- |
-|  `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism |
-|  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
-| `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
-|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Optional. Test Splitter uses the identifier to store and fetch the test plan and must be unique for each build and steps group. By default it will use a composite of `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, but it can be overridden by specifying the `BUILDKITE_SPLITTER_IDENTIFIER`. `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` must be accessible by the client when using the default. |
+| `BUILDKITE_API_ACCESS_TOKEN ` | - | Required, Buildkite API access token with `read_suites`, `read_test_plan`, and `write_test_plan` scopes. |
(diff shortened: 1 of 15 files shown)


README.md
```diff
@@ -17,12 +17,14 @@ The available Go binaries
 
 | Environment Variable | Default Value | Description |
 | ---- | ---- | ----------- |
-|  `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism |
-|  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
-| `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
-|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Optional. Test Splitter uses the identifier to store and fetch the test plan and must be unique for each build and steps group. By default it will use a composite of `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, but it can be overridden by specifying the `BUILDKITE_SPLITTER_IDENTIFIER`. `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` must be accessible by the client when using the default. |
+| `BUILDKITE_API_ACCESS_TOKEN ` | - | Required, Buildkite API access token with `read_suites`, `read_test_plan`, and `write_test_plan` scopes. |
+| `BUILDKITE_ORGANIZATION_SLUG` | - | Required, the slug of your Buildkite organization. |
```
nprizal, line 21:
```suggestion
| `BUILDKITE_ORGANIZATION_SLUG` | - | Required, the slug of your Buildkite organization. This is available in your pipeline environment, so you don't need to set it manually |
```

README.md
```diff
@@ -17,12 +17,14 @@ The available Go binaries
 
 | Environment Variable | Default Value | Description |
 | ---- | ---- | ----------- |
-|  `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism |
-|  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
-| `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
-|  `BUILDKITE_SPLITTER_IDENTIFIER` | `BUILDKITE_BUILD_ID/BUILDKITE_STEP_ID` | Optional. Test Splitter uses the identifier to store and fetch the test plan and must be unique for each build and steps group. By default it will use a composite of `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, but it can be overridden by specifying the `BUILDKITE_SPLITTER_IDENTIFIER`. `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID` must be accessible by the client when using the default. |
+| `BUILDKITE_API_ACCESS_TOKEN ` | - | Required, Buildkite API access token with `read_suites`, `read_test_plan`, and `write_test_plan` scopes. You can create access token from [Personal Settings](https://buildkite.com/user/api-access-tokens) in Buildkite |
```
amybiyuliu, line 20: To clarify the `read_suites` scope is required when we make request to the get_files endpoint right?

internal/api/client.go
```diff
@@ -1,17 +1,47 @@
...
+	ServerBaseUrl    string
+}
+
+// authTransport is a middleware for the HTTP client.
+type authTransport struct {
+	accessToken string
 }
 
+// RoundTrip adds the Authorization header to all requests made by the HTTP client.
+func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
```
amybiyuliu, line 28: Wow How this worked? I didn't see you explicitly call this method. I looked at your testing file and it seems when you call `c.httpClient.Get(svr.URL)` the auth header will just be added.

amybiyuliu (commented):
Nice work! I just left a couple of questions. 

Also wondering if we need the foo_spec.rb for this PR?

README.md
wooly: Yep, that's right

nprizal: I'm torn between putting `read_suites` token on the doc now, or when we implement the get_files. But decided to do it now.

internal/api/client.go
nprizal: We set `authTransport` as the `Transport` for `http.Client` when we create it within `api.NewClient` function. 
Every time we make a request using that http client, it will call the `RoundTrip` function.

README.md
```diff
@@ -17,12 +17,14 @@ The available Go binaries
...
 | `BUILDKITE_SPLITTER_TEST_FILE_PATTERN` | `spec/**/*_spec.rb` | Optional, glob pattern for discovering test files that need to be executed. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library*. |
 | `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` | - | Optional, glob pattern to use for excluding test files or directory. </br> *It accepts pattern syntax supported by [zzglob](https://github.com/DrJosh9000/zzglob?tab=readme-ov-file#pattern-syntax) library.* |
+| `BUILDKITE_SPLITTER_SUITE_SLUG` | - | Required, the slug of your test suite. |
 | `BUILDKITE_TEST_SPLITTER_CMD` | `bundle exec rspec {{testExamples}}` | Optional, test command for running your tests. Test splitter will fill in the `{{testExamples}}` placeholder with the test splitting results |
 
 For most use cases, Test Splitter should work out of the box due to the default values available from your Buildkite environment.
```
wooly, line 30: I don't think the authentication settings below this are valid any more, is that correct?

internal/api/fetch_test_plan.go
```diff
@@ -0,0 +1,58 @@
...
+	case resp.StatusCode >= 400 && resp.StatusCode <= 403:
+		// treat 400-403 as an error because the request is invalid
+		responseBody, _ := io.ReadAll(resp.Body)
+		var errorResp errorResponse
+		json.Unmarshal(responseBody, &errorResp)
+		return nil, fmt.Errorf(errorResp.Message)
+	default:
+		// ignore other errors and treat them as cache miss
+		return nil, nil
+	}
```
nprizal, line 42: @wooly @niceking I'm ignoring server error, except for 400 ~ 403, so the client can proceed and attempt to make POST request. Do you think this make sense? Or should we implement a retry mechanism for this as well?

The reason behind handling 400 ~ 403 is that we want to terminate the program if there is bad request (miss config), unauthorized, or not permitted (token doesn't have required scope)

wooly: I haven't fully thought through the implications of falling back to cache-miss yet, but it probably makes sense to document the individual status codes that we respond to in the splitter and what they mean. At least to have as a point of reference, even if the code groups them all together.

amybiyuliu (approved):
@nprizal and I reviewed the changes in this PR at today's team time. The changes look good to me

README.md
nprizal: Good catch! I have updated the readme, but don't want to change it too much for now. What do you think?

internal/api/fetch_test_plan.go
nprizal: Good suggestion. I have improve the readability and the test. Let me know what you think.

amybiyuliu (approved):

wooly:
Removing the DO NOT MERGE tag from this as it's now suitable to be merged into the `0.6.0` branch.

PR #66, opened 2024-05-20
Retry failed tests with a max retry count

amybiyuliu:
### Description
bk/bk pipeline rerun failed rspec tests outside of test splitter. We should support retry failing tests in the test splitter.

### Context

On basecamp, we decided to implement retry failing tests in two steps. This PR is part of step 1

### Changes

- Read from `BUILDKITE_SPLITTER_RETRY_COUNT` for max retry attempts 
- Create retry command from the original test command: Append `--only-failures` to the original test command, and remove all occurrences of `{{testExamples}}`
- Retry failed tests if retry is enabled (`BUILDKITE_SPLITTER_RETRY_COUNT` > 0)

### Testing

Testing the retry feature locally

A screenshot of failing tests passed after retry: 

Screenshot 2024-05-21 at 4 39 54 pm

A screenshot of failing tests still failed after used all retry attempts: 
Screenshot 2024-05-22 at 9 49 26 am

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,6 +116,7 @@ func main() {
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
 			exitCode := exitError.ExitCode()
 			if cfg.MaxRetries == 0 {
+				// If retry is disabled, we exit immediately with the same exit code from the test runner
 				logErrorAndExit(exitCode, "Rspec exited with error %d", err)
 			} else {
 				retryFailedTests(testRunner, cfg.MaxRetries)
@@ -147,10 +148,12 @@ func retryFailedTests(testRunner runner.Rspec, maxRetries int) {
 			if exitError := new(exec.ExitError); errors.As(err, &exitError) {
 				exitCode := exitError.ExitCode()
 				if retries >= maxRetries {
+					// If the command exits with an error and we've reached the maximum number of retries, we exit.
 					logErrorAndExit(exitCode, "Rspec exited with error %d", err)
 				}
 			}
 		} else {
+			// If the command exits successfully, we break the loop.
 			break
 		}
 	}
--- a/main.go
+++ b/main.go
@@ -119,7 +119,12 @@ func main() {
 				// If retry is disabled, we exit immediately with the same exit code from the test runner
 				logErrorAndExit(exitCode, "Rspec exited with error %v", err)
 			} else {
-				retryFailedTests(testRunner, cfg.MaxRetries)
+				retryExitCode := retryFailedTests(testRunner, cfg.MaxRetries)
+				if retryExitCode == 0 {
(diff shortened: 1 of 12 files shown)


main.go
```diff
@@ -105,7 +105,11 @@ func main() {
 	if err := cmd.Wait(); err != nil {
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
 			exitCode := exitError.ExitCode()
-			logErrorAndExit(exitCode, "Rspec exited with error %d", err)
+			if cfg.MaxRetries == 0 {
+				logErrorAndExit(exitCode, "Rspec exited with error %d", err)
+			} else {
+				retryFailedTests(testRunner, cfg)
```
wooly, line 111: Do we want to add some logging here like 'Retrying, attempt X of Y' or something to show that a) retries are configured, and b) how many there are going to be?

wooly (commented):
I wonder if the environment variable should just be `BUILDKITE_SPLITTER_RETRY_COUNT`?

What do you think @amybiyuliu @nprizal?

amybiyuliu:
> I wonder if the environment variable should just be `BUILDKITE_SPLITTER_RETRY_COUNT`?
> 
> What do you think @amybiyuliu @nprizal?

Yeah, I can make the env name shorter. In Captain's config file, the max retry value is just called `retries.attempt`. I think we can document it only retry the failed tests in the README file to make it clear to the user, and use a shorter env name.

main.go
```diff
@@ -124,6 +134,39 @@ func main() {
...
 }
 
+func retryFailedTests(testRunner runner.Rspec, maxRetries int) int {
+	// Retry failed tests
+	retries := 0
+	for retries < maxRetries {
+		retries++
+		fmt.Printf("Attempt %d of %d to retry failing tests\n", retries, maxRetries)
+
+		cmd, err := testRunner.RetryCommand()
```
amybiyuliu, line 144: @nprizal and I talked about unit testing this function by providing a simple command(i.e 'cat banana') to the test runner such that it would be easy to test this func. 
However, the `RetryCommand` will append `--only-failures` to the test command we provide and `cat banana --only-failures` isn't a valid command to run. I didn't add unit tests for this but have done manual testing.

wooly: You can use `true` as the command rather than `cat banana`, as `true --only-failures` will exit with 0:

```
~/version-control/buildkite/buildkite
❯ true --only-failures

~/version-control/buildkite/buildkite
❯ echo $?
0
```

amybiyuliu: Thanks Steve that's exactly what i'm looking for! I have added the unit tests for this retry logic

wooly (approved):
Looks good! Step 1 tick :)

wooly:
:tada:

PR #69, opened 2024-05-21
Refactor test command default handling

nprizal:
### Description
Currently, the default test command is set inside the `config` package using Rspec default command. Since we will be adding support to other test runners (frameworks), we need to move this inside the `Rspec` struct.

I also store the `testCommand` as the property of Rspec struct so all functions can have access the value without having to pass the command every time.

### Testing

It's worth testing test-splitter with and without custom test command.

### DIFF
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -14,6 +14,17 @@ import (
 // For now, Rspec provides rspec specific behaviour to execute
 // and report on tests in the Rspec framework.
 type Rspec struct {
+	TestCommand string
+}
+
+func NewRspec(testCommand string) Rspec {
+	if testCommand == "" {
+		testCommand = "bundle exec rspec {{testExamples}}"
+	}
+
+	return Rspec{
+		TestCommand: testCommand,
+	}
 }
 
 // GetFiles returns an array of file names, for files in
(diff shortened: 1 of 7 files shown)


wooly:
Looks good :+1: Will give it a test. I don't think this needs a changelog entry, since the API isn't changing, and if we get to a changelog as part of the `0.5.x` release cycle.

internal/runner/rspec.go
```diff
@@ -14,6 +14,17 @@ import (
 // For now, Rspec provides rspec specific behaviour to execute
 // and report on tests in the Rspec framework.
 type Rspec struct {
+	TestCommand string
```
nprizal, line 17: We can store Rspec related config here (e.g. test command, retry command, etc) for all functions to access.

internal/runner/rspec.go
```diff
@@ -14,6 +14,17 @@ import (
...
+}
+
+func NewRspec(testCommand string) Rspec {
+	if testCommand == "" {
```
nprizal, line 21: Set the default test command here instead of the top level `config` package.

wooly:
Can we add a breaking change to rename the test command environment variable to `BUILDKITE_SPLITTER_TEST_CMD`?

wooly (approved):
Have tested with two scenarios: test cmd absent and test cmd present in env, and it's working in both cases.

LGTM!

amybiyuliu (approved):
Nice work! Keen to have it merged so I can also use it in the retry PR 😜

nprizal:
> Can we add a breaking change to rename the test command environment variable from BUILDKITE_TEST_SPLITTER_CMD to BUILDKITE_SPLITTER_TEST_CMD?

This has been bothering me too. Let's do that!

PR #71, opened 2024-05-21
Use new API endpoint to fetch and create the test plan

nprizal:
### ⚠️ BREAKING CHANGES
These changes introduced 3 required configs
- `BUILDKITE_API_ACCESS_TOKEN`. You now need Buildkite API access token that has `read_suites`, `read_test_plan`, and `write_test_plan` scopes.
- `BUILDKITE_ORGANIZATION_SLUG`. This is available in your Buildkite Pipeline environment, so you don't need to set it manually.
- `BUILDKITE_SPLITTER_SUITE_SLUG`. This replaces `BUILDKITE_SPLITTER_SUITE_TOKEN` and need to be set manually.

### Description

As part of split by example implementation, we want to change how the client fetch and create test plan from the server. The new flow will look like this
```
// main.go
func fetchOrCreatePlan() {
  plan := api.FetchPlan()

  if plan != nil {
    return plan
  }

  if plan != nil {
    return plan
  }
}
```

This PR also replaced the endpoint to the new endpoint.

### Changes

- Add 3 new configs
- Add `httpClient` to `api.client` struct that has Transport to attach access token to each request
- Implement `FetchTestPlan` in the `api.client` struct
- Replace endpoint in `CreateTestPlan` to the new endpoint
- Removed `BUILDKITE_SPLITTER_SUITE_TOKEN` from config

### Testing

As this is a breaking change, I suggest testing it manually against a local Buildkite server (or Prod). I tested it manually against my local Buildkite server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,13 +17,15 @@ The available Go binaries
 
 | Environment Variable | Default Value | Description |
 | ---- | ---- | ----------- |
-|  `BUILDKITE_PARALLEL_JOB_COUNT` | - | Required, total number of parallelism |
-|  `BUILDKITE_PARALLEL_JOB` | - | Required, test plan for specific node |
-| `BUILDKITE_SPLITTER_SUITE_TOKEN` | `BUILDKITE_ANALYTICS_TOKEN` | Required, unique token for Test Suite that is being parallelised |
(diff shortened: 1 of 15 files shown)


amybiyuliu (approved):
Looks good. I reviewed and approved the original PR

PR #72, opened 2024-05-22
Update test plan schema

nprizal:
### Description
Update the test plan schema to support Split by Example.

New request schema
```
{
  "identifier": "fruits",
  "parallelism": "2",
  "tests": {
    "files": [
      {
        "path": "apple_spec.rb"
      }
    ],
    "examples": [
      {
        "path": "banana_spec.rb:1",
        "identifier": "something",
        "scope": "peel off",
        "name": "something"
      }
    ]
  }
}
```

New response schema
```
{
  "identifier": "fruits",
  "parallelism": "2",
  "tasks": {
    "0": {
      "node_number": 0,
      "tests": [
        {
          "path": "apple_spec.rb",
          "estimated_duration": 1,
          "format": "file"
        }
      ],
    },
    "1": {
      "node_number": 1,
      "tests": [
        {
          "path": "banana_spec.rb:1",
          "estimated_duration": 1,
          "identifier": "something",
          "scope": "peel off",
          "name": "something",
          "format": "example"
        }
      ]
    }
  }
}
```

### DIFF
--- a/internal/plan/type.go
+++ b/internal/plan/type.go
@@ -1,21 +1,23 @@
 package plan
 
+type TestCaseFormat string
+
+const (
+	TestCaseFormatFile    TestCaseFormat = "file"
+	TestCaseFormatExample TestCaseFormat = "example"
+)
+
 // TestCase represents a single test case.
 type TestCase struct {
-	Path              string `json:"path"`
-	EstimatedDuration *int   `json:"estimated_duration"`
-}
-
-// Tests represents a set of tests.
-type Tests struct {
-	Cases  []TestCase `json:"cases"`
-	Format string     `json:"format"`
+	Path              string         `json:"path"`
+	EstimatedDuration *int           `json:"estimated_duration"`
+	Format            TestCaseFormat `json:"format"`
 }
 
 // Task represents the task for the given node.
 type Task struct {
-	NodeNumber int   `json:"node_number"`
-	Tests      Tests `json:"tests"`
+	NodeNumber int        `json:"node_number"`
+	Tests      []TestCase `json:"tests"`
 }
 
 // TestPlan represents the entire test plan.
--- a/internal/plan/type.go
+++ b/internal/plan/type.go
@@ -10,8 +10,8 @@ const (
 // TestCase represents a single test case.
 type TestCase struct {
 	Path              string         `json:"path"`
-	EstimatedDuration *int           `json:"estimated_duration"`
-	Format            TestCaseFormat `json:"format"`
+	EstimatedDuration *int           `json:"estimated_duration,omitempty"`
+	Format            TestCaseFormat `json:"format,omitempty"`
 }
 
 // Task represents the task for the given node.
--- a/main.go
+++ b/main.go
@@ -154,8 +154,7 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
 	testCases := []plan.TestCase{}
 	for _, file := range files {
 		testCases = append(testCases, plan.TestCase{
-			Path:   file,
-			Format: plan.TestCaseFormatFile,
+			Path: file,
 		})
 	}
 
--- a/main.go
+++ b/main.go
@@ -72,7 +72,7 @@ func main() {
 
 	// execute tests
 	runnableTests := []string{}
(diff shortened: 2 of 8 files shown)


main.go
```diff
@@ -154,20 +154,18 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
 	testCases := []plan.TestCase{}
 	for _, file := range files {
 		testCases = append(testCases, plan.TestCase{
-			Path: file,
+			Path:   file,
+			Format: plan.TestCaseFormatFile,
```
amybiyuliu, line 158: Maybe we need to remove format when sending the request?

I'm also not sure if we allow the format field to be optional for `plan.TestCase`. Maybe we need to update that as well?

amybiyuliu (commented):
Nice work! I left a question about whether we stop sending `format` to the api. Other than that it looks great.

main.go
wooly: @amybiyuliu this was a deliberate decision to maintain consistency with the way that file tests and example tests are sent to the server:

```json
{
  "identifier": "fruits",
  "parallelism": "2",
  "tests": {
    "files": [
      {
        "path": "apple_spec.rb"
      }
    ],
    "examples": [
      {
        "path": "banana_spec.rb:1",
        "identifier": "something",
        "scope": "peel off",
        "name": "something"
      }
    ]
  }
}
```

It also allows us to add additional attributes to the file tests without having to change the schema in an intrusive way.

Another advantage is that objects tend to be easier to schema migrate than other types. Consider an API which sends the following data:

```json
{
  "paths": ["some_path1.rb", "some_path2.rb"]
}
```

API Consumer V1 consumes this API by mapping over the strings and doing something with each string.

Now, we want to change the API so that it sends the number of lines per file as well, so naturally we'd want to migrate to a schema that looks like this:

```json
{
  "paths": [{ "path": "some_path1.rb", "lines": 10 }, { "path": "some_path2.rb", "lines": 13 }]
}
```

However, we can't do this in a way that allows V1 consumers to still function, as they don't have the ability to distinguish between a path that was a string (in the old world) and a 'path' that is an object in the new world. Two solutions for this would be:

1) Change the new-world schema to use a different top-level key, so V1 consumers are none-the-wiser:

```json
{
  "paths": ["some_path1.rb", "some_path2.rb"],
  "pathsv2": [{ "path": "some_path1.rb", "lines": 10 }, { "path": "some_path2.rb", lines: 13 }]
}
```

_or_

2) Given this is a hypothetical, we design V1 of the API to allow these additional attributes to be added later, by changing the V1 API before it's in the wild to use objects instead of scalar values, allowing for easier migration later:

```json
{
  "paths": [{ "path": "some_path1.rb" }, { "path": "some_path2.rb" }]
}
```

Option 2 is what we are effectively doing here: making it easier for our future selves if we need to add additional fields to the File schema later.

nprizal: That's a great explanation @wooly 💚. But I think what @amybiyuliu meant was to remove the `format` in the request body now we separate `files` and `examples`. 
```
{
  tests: {
    files: [
      {
        path: "my_spec.rb",
        format: "file" <-- remove this
      }
    ]
}
```

wooly: Haha, what a dope! Sorry for the noise!

homer-simpson-animated-gif-7

internal/plan/type.go
```diff
@@ -1,21 +1,23 @@
...
-	Path              string `json:"path"`
-	EstimatedDuration *int   `json:"estimated_duration"`
-}
-
-// Tests represents a set of tests.
-type Tests struct {
-	Cases  []TestCase `json:"cases"`
-	Format string     `json:"format"`
+	Path              string         `json:"path"`
+	EstimatedDuration *int           `json:"estimated_duration,omitempty"`
```
nprizal, line 13: I added `omitempty` option to prevent sending the request with these empty attributes, as they are optional.

wooly (approved):

main.go
amybiyuliu: Thanks for the detailed explanation @wooly it's super useful to know how we came to the decision of the new schema.

amybiyuliu (approved):
Nice work! 🚀

wooly:
Nice!

PR #75, opened 2024-05-24
Add new line in each log

nprizal:
Some logs doesn't have a new line. This PR adds new line to all logs.

### DIFF
--- a/main.go
+++ b/main.go
@@ -169,8 +169,7 @@ func retryFailedTests(testRunner runner.Rspec, maxRetries int) int {
 
 // logErrorAndExit logs an error message and exits with the given exit code.
 func logErrorAndExit(exitCode int, format string, v ...any) {
-	fmt.Printf(format, v...)
-	fmt.Print("\n")
+	fmt.Printf(format+"\n", v...)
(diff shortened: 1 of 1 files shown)


wooly:
I _think_ we should change the base of this branch to 0.5.x

wooly (approved):

amybiyuliu (approved):
Thanks for adding this!

main.go
```diff
@@ -170,6 +170,7 @@ func retryFailedTests(testRunner runner.Rspec, maxRetries int) int {
 // logErrorAndExit logs an error message and exits with the given exit code.
 func logErrorAndExit(exitCode int, format string, v ...any) {
 	fmt.Printf(format, v...)
+	fmt.Print("\n")
```
DrJosh9000, line 173: Drive-by suggestion:

1. A shorter way to write this is `fmt.Println()`
2. Alternatively you could use `format+"\n"` in the line above

wooly: We love driveby suggestions, thanks @DrJosh9000! @nprizal want to make the change and ship?

PR #76, opened 2024-05-27
Create a function in rspec struct to execute --dry-run

nprizal:
### Description

To split test by example, the client needs to run Rspec with --dry-run to get the list of test examples within the slow files. We need to create a function that can execute Rspec in dry run mode, parse the output, and return the list of `TestCases`.

This PR also changes the `TestCase` struct to include other attributes, such as `scope` and `name`.


internal/runner/rspec_test.go
```diff
@@ -152,3 +153,34 @@ func TestRspecDiscoveryPattern_ExcludePattern(t *testing.T) {
...
+	rspec := NewRspec("./test/mock_rspec_dry_run")
+	files := []string{"spec/spells/expelliarmus_spec.rb"}
+	got, err := rspec.GetExamples(files)
+
+	want := []plan.TestCase{
+		{
+			Format:     plan.TestCaseFormatExample,
+			Identifier: "./spec/spells/expelliarmus_spec.rb[1:1]",
+			Name:       "disarms the opponent",
+			Path:       "./spec/spells/expelliarmus_spec.rb:3",
```
nprizal, line 167: Do we want to include a leading `./` to the file path? FYI, when the client discovers the test files using file globbing, it doesn't include the leading `./`.

internal/runner/rspec.go
```diff
@@ -112,3 +113,56 @@ func (r Rspec) commandNameAndArgs(testCases []string) (string, []string, error)
...
+}
+
+// GetExamples returns an array of test examples within the given files.
+func (r Rspec) GetExamples(files []string) ([]plan.TestCase, error) {
+	cmdName, cmdArgs, err := r.commandNameAndArgs(files)
+	if err != nil {
+		return nil, err
+	}
+
+	cmdArgs = append(cmdArgs, "--dry-run", "--format", "json")
```
amybiyuliu, line 142: This would use the user defined test command for the dry run right?
I don't know how `--dry-run` works with other rspec options (say if a user provided more option flags). Would the default rspec command `bundle exec rspec {files} --dry-run` be more robust?

nprizal: Yes, it will use the exact same command but with `--dry-run`. The dry run is basically similar to running Rspec with all options, but without actually executing the test. I would argue this is more robust because the dry run result will reflect the actual Rspec run. If, for instance, we use the default `bundle exec rspec {files} --dry-run`, but the customers use `--tag` to exclude some tests, thenm the excluded tests will be in the dry run result.

amybiyuliu: oh yeah! I forget about things like `tags`! Good call, we should keep using the user provided command

amybiyuliu (approved):
Nice work! 🥳

wooly:
Is this assuming that the output of the dry-run command will be in JSON format on a users machine? Wont that only be the case if the output format is set to JSON using an rspec flag?

nprizal:
> Is this assuming that the output of the dry-run command will be in JSON format on a users machine? Wont that only be the case if the output format is set to JSON using an rspec flag?

That's a good point! I added `--format json` so the result will be in `JSON`, however, I just tested that when user specifies other formatter, it will be in the output as well. So this isn't going to work. I think the solution is to explicitly tell rspec to save the output of json into a file `--dry-run --format json --out output.txt`

wooly:
Can we add --dry-run --format json to the command instead of just --dry-run? Will later occurrences of the same flag override previous ones, so if a customer specifies an execution command of `bin/rspec --options .rspec.ci --format xml {{files}}`, say, and we add `--dry-run --format json`, does the second format flag override the first?

wooly:
Do we definitely need to kick the results out to a file here?

nprizal:
As discussed. Running Rspec with multiple `--format` will output each formatter instead of overriding the previous `--format`. To get the output from specific `--format`, we need to tell Rspec to send the output to a file using `--out output.json`

internal/runner/rspec.go
```diff
@@ -112,3 +113,71 @@ func (r Rspec) commandNameAndArgs(testCases []string) (string, []string, error)
...
+		return nil, err
+	}
+
+	cmdArgs = append(cmdArgs, "--dry-run", "--format", "json", "--out", f.Name())
+
+	output, err := exec.Command(cmdName, cmdArgs...).CombinedOutput()
+
+	outputStr := string(output)
+	if err != nil {
+		return []plan.TestCase{}, fmt.Errorf("failed to run rspec dry run: %v", outputStr)
```
amybiyuliu, line 158: Do we want to wrap the `err` or `outputStr` with the message? 
I was thinking:
`fmt.Errorf("failed to run rspec dry run: %v", err)`

Do we need to move error handling to line 155?

nprizal: I did that initially, but the message from output is more descriptive.

```
fmt.Errorf("failed to run rspec dry run: %v", err)
> failed to run rspec dry run: exit status 10
```
vs
```
fmt.Errorf("failed to run rspec dry run: %v", outputStr)
> failed to run rspec dry run: Could not locate Gemfile or .bundle/ directory
```

nprizal: Also, `CombinedOutput` will return output from both `stdout` and `stderr`.

internal/runner/rspec_test.go
```diff
@@ -152,3 +153,65 @@ func TestRspecDiscoveryPattern_ExcludePattern(t *testing.T) {
...
+	if err != nil {
+		t.Errorf("Rspec.GetExamples(%q) error = %v", files, err)
+	}
+
+	if diff := cmp.Diff(got, want); diff != "" {
+		t.Errorf("Rspec.GetExamples(%q) diff (-got +want):\n%s", files, diff)
+	}
+}
+
+func TestRspecGetExamples_WithOtherFormatters(t *testing.T) {
```
wooly, line 188: Is it also worth adding a test for when a formatter with JSON is already specified. BKs `.rspec.ci` file will also add a JSON formatter with a different output location, so I'm interested in seeing how that will interact with ours: https://github.com/buildkite/buildkite/blob/0a0b32ac150fa1fcb5a92044d7a94e4c840a527c/.rspec.ci#L10

internal/runner/rspec_test.go
```diff
@@ -152,3 +153,71 @@ func TestRspecDiscoveryPattern_ExcludePattern(t *testing.T) {
 		t.Errorf("Rspec.discoveryPattern() diff (-got +want):\n%s", diff)
 	}
 }
+
+func TestRspecGetExamples(t *testing.T) {
+	rspec := NewRspec("rspec")
+	files := []string{"fixtures/spec/spells/expelliarmus_spec.rb"}
```
wooly, line 159: This will come from server-side with a leading `./`. Can we confirm this works?

nprizal: Good point! I have updated the test.

wooly: Ace!

wooly:
Looking good, I think!

internal/runner/rspec.go
amybiyuliu: Yeah, I think it gives more meaningful message if we wrap `outputStr`

Just curious, when dry run fails, does the program exit with `10` or `1`? Do we want do document the exit code in the README?

wooly (approved):

PR #80, opened 2024-05-30
Add debug mode to log the http request and response 

nprizal:
### Description

To help troubleshooting test splitter client, we want to log each request and response to/from the server. The logs should only be displayed when debug mode is enabled, and should redact any sensitive information (e.g. API token).

### Testing

Run `BUILDKITE_SPLITTER_DEBUG_ENABLED=true ./test-splitter`, and check if each http request/response is printed.

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -1,6 +1,9 @@
 package api
 
 import (
+	"bytes"
+	"fmt"
+	"io"
 	"net/http"
 )
 
@@ -17,17 +20,41 @@ type ClientConfig struct {
 	AccessToken      string
 	OrganizationSlug string
 	ServerBaseUrl    string
+	DebugEnabled     bool
 }
 
 // authTransport is a middleware for the HTTP client.
 type authTransport struct {
-	accessToken string
+	accessToken  string
+	debugEnabled bool
 }
 
 // RoundTrip adds the Authorization header to all requests made by the HTTP client.
 func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.debugEnabled {
+		fmt.Println("--- 🕵️ DEBUG")
+		fmt.Println("Request:")
+		fmt.Printf("%s %s\n\n", req.Method, req.URL.String())
+	}
+
+	// IMPORTANT: We need to set the token after printing the request to avoid leaking the token in logs.
 	req.Header.Set("Authorization", "Bearer "+t.accessToken)
-	return http.DefaultTransport.RoundTrip(req)
+
+	resp, err := http.DefaultTransport.RoundTrip(req)
+
+	if t.debugEnabled {
+		body, _ := io.ReadAll(resp.Body)
+		defer resp.Body.Close()
+
+		// We need to replace the body with a new reader, so it can be read again by the caller.
+		resp.Body = io.NopCloser(bytes.NewBuffer(body))
+
+		fmt.Println("Response:")
+		fmt.Println(resp.Status)
+		fmt.Println(string(body))
+	}
+
+	return resp, err
 }
 
 // NewClient creates a new client for the test splitter API with the given configuration.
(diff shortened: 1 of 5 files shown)


internal/api/client.go
```diff
@@ -17,25 +20,50 @@ type ClientConfig struct {
...
-	return http.DefaultTransport.RoundTrip(req)
+
+	resp, err := http.DefaultTransport.RoundTrip(req)
+
+	if t.debugEnabled {
+		body, _ := io.ReadAll(resp.Body)
+		defer resp.Body.Close()
+
+		// We need to replace the body with a new reader, so it can be read again by the caller.
+		resp.Body = io.NopCloser(bytes.NewBuffer(body))
```
nprizal, line 50: @DrJosh9000 Is there a better way to do this?

internal/api/client.go
```diff
@@ -17,25 +20,50 @@ type ClientConfig struct {
...
+	accessToken  string
+	debugEnabled bool
 }
 
 // RoundTrip adds the Authorization header to all requests made by the HTTP client.
 func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.debugEnabled {
+		fmt.Println("--- 🕵️ DEBUG")
+		fmt.Println("Request:")
+		fmt.Printf("%s %s\n\n", req.Method, req.URL.String())
```
DrJosh9000, line 37: You don't need to manually call the `String` method when using `fmt.Printf` - just use `%v`.

internal/api/client.go
```diff
@@ -17,25 +20,50 @@ type ClientConfig struct {
...
-	accessToken string
+	accessToken  string
+	debugEnabled bool
 }
 
 // RoundTrip adds the Authorization header to all requests made by the HTTP client.
 func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.debugEnabled {
+		fmt.Println("--- 🕵️ DEBUG")
+		fmt.Println("Request:")
```
DrJosh9000, line 36: Consider using [`httputil.DumpRequestOut`](https://pkg.go.dev/net/http/httputil#DumpRequestOut), which will give you a more complete picture of the outgoing request.

internal/api/client.go
```diff
@@ -17,25 +20,50 @@ type ClientConfig struct {
...
+	}
+
+	// IMPORTANT: We need to set the token after printing the request to avoid leaking the token in logs.
 	req.Header.Set("Authorization", "Bearer "+t.accessToken)
-	return http.DefaultTransport.RoundTrip(req)
+
+	resp, err := http.DefaultTransport.RoundTrip(req)
+
+	if t.debugEnabled {
+		body, _ := io.ReadAll(resp.Body)
```
DrJosh9000, line 46: You'll need to handle this error.

internal/api/client.go
DrJosh9000: So you can either do it this way (note that in this particular instance you should use `bytes.NewReader` (which is read-only) instead of `bytes.NewBuffer`) or you can use [`httputil.DumpResponse(resp, true)`](https://pkg.go.dev/net/http/httputil#DumpResponse) which does the "body swap" for you.

internal/api/client.go
```diff
@@ -17,25 +20,50 @@ type ClientConfig struct {
...
 type authTransport struct {
-	accessToken string
+	accessToken  string
+	debugEnabled bool
 }
 
 // RoundTrip adds the Authorization header to all requests made by the HTTP client.
 func (t *authTransport) RoundTrip(req *http.Request) (*http.Response, error) {
+	if t.debugEnabled {
+		fmt.Println("--- 🕵️ DEBUG")
```
DrJosh9000, line 35: One other thing: consider using `log.Print{,f,ln}` or `fmt.Fprint{,f,ln}(os.Stderr, ...)` to send debugging logs to a separate stream than the normal output.

PR #83, opened 2024-06-03
Implement split by example

nprizal:
### Description
This PR implements split by example that can be enabled by setting `BUILDKITE_SPLITTER_SPLIT_BY_EXAMPLE` environment variable to `true`.

When split by example is enabled, the test splitter client will fetch the timing data for each test files.
It then calls Rspec with `--dry-run` for slow files that have duration greater or equal to 3 minutes, and send the test examples within those files along with the rest of the test files to the server to get the test plan.

A failure when fetching the timing data or executing Rspec `--dry-run` will terminate the test splitter client.

This PR also adds debugging around fetching and creating test plan.


internal/plan/fallback.go
```diff
@@ -8,10 +8,10 @@ import (
 
 // CreateFallbackPlan creates a fallback test plan for the given tests and parallelism.
 // It distributes test cases evenly accross the tasks using deterministic algorithm.
-func CreateFallbackPlan(testCases []TestCase, parallelism int) TestPlan {
+func CreateFallbackPlan(files []string, parallelism int) TestPlan {
```
nprizal, line 11: We don't need an array of a whole test struct to create a fallback plan, so I refactored it to takes an array of file name instead.

internal/runner/rspec.json
nprizal, line 1: Remove this file

main.go
```diff
@@ -194,39 +201,92 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
+		testPlan = plan.CreateFallbackPlan(files, cfg.Parallelism)
+	}
+
+	return testPlan, nil
+}
+
+// createRequestParam creates the request parameters for the test plan.
+// If SplitByExample is disabled (default), it will return the default params that contain all the files.
+// If SplitByExample is enabled, it will split the slow files into examples and return it along with the rest of the files.
+func createRequestParam(cfg config.Config, files []string, client api.Client, runner TestRunner) api.TestPlanParams {
```
nprizal, line 231: I extracted this as a separate function so we can easily test it.

test/spec/fruits/grap_spec.rb
nprizal, line 1: typo on the file name.

internal/api/client.go
```diff
@@ -6,7 +6,7 @@ import (
 
 // client is a client for the test splitter API.
 // It contains the organization slug, server base URL, and an HTTP client.
-type client struct {
+type Client struct {
```
nprizal, line 9: I make this public because I need to pass this type on one of the function in the `main` package.

main.go
```diff
@@ -194,39 +201,92 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
+
+	// The server can return an "error" plan indicated by an empty task list (i.e. `{"tasks": {}}`).
+	// In this case, we should create a fallback plan.
+	if len(testPlan.Tasks) == 0 {
+		fmt.Println("Test splitter server returned an error, using fallback mode. Your build may take longer than usual.")
+		testPlan = plan.CreateFallbackPlan(files, cfg.Parallelism)
+	}
+
+	return testPlan, nil
+}
```
nprizal, line 226: It looks like a lot of changes here, but this is an existing code. The only changes are the extracting params creation into a separate function.

main.go
```diff
@@ -194,39 +201,92 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
-	// The server can return an "error" plan indicated by an empty task list (i.e. `{"tasks": {}}`).
-	// In this case, we should create a fallback plan.
-	if len(testPlan.Tasks) == 0 {
-		fmt.Println("Test splitter server returned an error, using fallback mode. Your build may take longer than usual.")
-		testPlan = plan.CreateFallbackPlan(testCases, cfg.Parallelism)
+	// Get the examples for the slow files.
+	// If we can't get examples for the slow files, we should fallback to default (split by file).
+	examples, err := runner.GetExamples(slowFiles)
+	if err != nil {
+		fmt.Printf("Split by example failed, using default mode.")
```
nprizal, line 279: Any suggestion for the message?

nprizal: We agreed to terminate the client in this scenario.

internal/api/fetch_files_timing.go
```diff
@@ -0,0 +1,84 @@
...
+
+	requestBody, err := json.Marshal(fetchFilesTimingParams{
+		Paths: files,
+	})
+
+	if err != nil {
+		return nil, fmt.Errorf("converting params to JSON: %w", err)
+	}
+
+	req, err := http.NewRequest(http.MethodGet, url, bytes.NewBuffer(requestBody))
```
nprizal, line 35: This isn't going to work. Request body for `GET` will be ignored. The server currently expecting the files to be in query params, but it doesn't work either because the payload is too large to be a query params. We need to change the endpoint into a `POST` endpoint, or remove the files params altogether. @wooly

nprizal, line 35: This isn't going to work. Request body for GET will be ignored. The server currently expecting the files to be in query params, but it doesn't work either because the payload is too large to be a query params. We need to change the endpoint into a POST endpoint, or remove the files params altogether. @wooly

internal/api/fetch_files_timing.go
```diff
@@ -0,0 +1,84 @@
...
+	"slices"
+)
+
+type fetchFilesTimingParams struct {
+	Paths []string `json:"paths"`
+}
+
+type fileTiming struct {
+	Path     string
+	Duration int
```
amybiyuliu, line 18: I noticed we store durations as real in the daily file stats table. I'm wondering if int would work?

nprizal: Ah good point. So it's a float with value in second right?
@wooly

nprizal: Resolved

wooly: Should we change this back to int? I don't think sub-second timing will make a huge amount of difference for SBE, but will simplify our deployment process with a few customers already using the splitter.

internal/api/fetch_files_timing.go
```diff
@@ -0,0 +1,66 @@
...
+		return nil, fmt.Errorf("reading response body: %w", err)
+	}
+
+	if resp.StatusCode != http.StatusOK {
+		var errorResp errorResponse
+		json.Unmarshal(responseBody, &errorResp)
+		return nil, fmt.Errorf(errorResp.Message)
+	}
+
+	var filesTiming map[string]int
```
amybiyuliu, line 53: Would it be possible that the server returns the file timing as a float in string format, or the server always returns int? I don't know if Go would be unhappy when the type doesn't match?

nprizal: It's possible if we change the API, like what we had last week. The integration test by pact will make sure that API doesn't break its contract with the client.

amybiyuliu (approved):
Nice work! Looking good to me

internal/config/read.go
```diff
@@ -37,6 +38,8 @@ func (c *Config) readFromEnv() error {
 	c.Mode = getEnvWithDefault("BUILDKITE_SPLITTER_MODE", "static")
 	c.TestCommand = os.Getenv("BUILDKITE_SPLITTER_TEST_CMD")
 
+	c.SplitByExample = os.Getenv("BUILDKITE_SPLITTER_SPLIT_BY_EXAMPLE") == "true"
```
niceking, line 41: I wonder if we should make this check case insensitive?

internal/runner/rspec.go
```diff
@@ -151,6 +152,8 @@ func (r Rspec) GetExamples(files []string) ([]plan.TestCase, error) {
 
 	cmdArgs = append(cmdArgs, "--dry-run", "--format", "json", "--out", f.Name())
 
+	debug.Println("Running rspec --dry-run")
```
niceking, line 155: [nit] could put backticks around the cmd bit?

main.go
```diff
@@ -213,15 +221,111 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
+}
+
+// createRequestParam creates the request parameters for the test plan.
+// If SplitByExample is disabled (default), it will return the default params that contain all the files.
+// If SplitByExample is enabled, it will split the slow files into examples and return it along with the rest of the files.
+//
+// Error is returned if there is a failure to fetch test file timings or to get the test examples from test files when SplitByExample is enabled.
+func createRequestParam(cfg config.Config, files []string, client api.Client, runner TestRunner) (api.TestPlanParams, error) {
+	if !cfg.SplitByExample {
+		debug.Println("Splitting by file")
```
niceking, line 249: Does your debug package print the time as well as the log line?

main.go
```diff
@@ -213,15 +221,111 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
 	return testPlan, nil
 }
+
+type fileTiming struct {
+	Path     string
+	Duration time.Duration
+}
+
+// createRequestParam creates the request parameters for the test plan.
+// If SplitByExample is disabled (default), it will return the default params that contain all the files.
```
niceking, line 243: The include/exclude filters have already been applied at this point right? Might be worth a mention

main.go
```diff
@@ -213,15 +221,111 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
+	// If the duration is the same, sort by path.
+	slices.SortFunc(allFilesTiming, func(a, b fileTiming) int {
+		if a.Duration == b.Duration {
+			return strings.Compare(a.Path, b.Path)
+		}
+		return int(b.Duration - a.Duration)
+	})
+
+	// Split the timings into slow files and the rest of the files.
+	threshold := int(float64(len(allFilesTiming)) * 0.3)
```
niceking, line 298: Hmmm, I wonder if we should be a bit smarter than just taking the slowest 30%? 

Like consider these scenarios:
- the changeset add lots of new files which we won't have timings for. there are so many new files that some of them are sorted in the slowest 30%
- the test files are all really small. perhaps we should apply a minimum, say 60s for a file's execution time before it meets the slow threshold? 

Like instead of having to SBE the slowest 30%, maybe it would be better to split *up to* 30% of the slowest test files?

test/spec/fruits/apple_spec.rb
```diff
@@ -0,0 +1,5 @@
+RSpec.describe "Apple" do
```
niceking, line 1: Are these files supposed to be in here?

main.go
nprizal: Currently, it will print the time and debug message. Do you think we need to add the line number as well?

It looks like this currently
```
2024/06/17 10:04:05 DEBUG: Fetching test plan
2024/06/17 10:04:05 DEBUG: No test plan found, creating a new plan
2024/06/17 10:04:05 DEBUG: Splitting by example
2024/06/17 10:04:05 DEBUG: Fetching timings for 1748 files
...
```

main.go
nprizal: This is a good point. Maybe we can split the 30% of _known_ slow files instead? The `GET /test_files` endpoint only returns test files that have been executed before, so we can be confident that the new test files isn't included in the 30% slow files? What do you think?

test/spec/fruits/apple_spec.rb
nprizal: Yes, I've introduced an integration testing between the client and `rspec`, and these files are used as a fixtures to test that integration.

wooly:
We can uptick our pact contracts for the create test plan endpoint before we release?

main.go
niceking: Nope I was just checking that we had timestamps!

main.go
niceking: As per slack discussion - I reckon we change so that we only split files that take longer than 3min to execute!

nprizal: We discussed this in Slack, and decided to define "slow files" as files that have duration greater or equal to 3 minutes.

main.go
```diff
@@ -213,15 +219,104 @@ func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []strin
...
+	for _, file := range files {
+		allFilesTiming = append(allFilesTiming, fileTiming{
+			Path:     file,
+			Duration: timings[file],
+		})
+	}
+
+	// Get files that has duration greater or equal to the slow file threshold.
+	// Currently, the slow file threshold is set to 3 minutes which is roughly 70% of optimal 4 minutes node duration.
+	slowFileThreshold := 3 * time.Minute
```
niceking, line 289: 👍

niceking (approved):
Nice, looks good, have you integration tested this locally? It might be worth pulling out the `slowFileThreshold` into an undocumented configurable as it will make this easier to test, as we currently don't have any data on prod that would meet this threshold

nprizal:
> Nice, looks good, have you integration tested this locally? It might be worth pulling out the `slowFileThreshold` into an undocumented configurable as it will make this easier to test, as we currently don't have any data on prod that would meet this threshold

Good point! I have made the threshold configurable (undocumented) and tested the integration locally.
Screenshot 2024-06-18 at 2 14 10 PM

PR #85, opened 2024-06-06
Create debug package

nprizal:
Create debug package that has functions to log any message when it is enabled. 

The debugger is enabled globally using `debug.SetDebug(true)`. 
By default, it will output the message to `os.stdout`, but you can set the output to be any `io.Writer` by calling `debut.SetOutput(file)`

```
func main() {
  debug.SetDebug(true)
  debug.Println("Hello world")
}

> 2024/06/06 15:17:21 DEBUG: Hello world
```

### DIFF
--- a/internal/debug/debug.go
+++ b/internal/debug/debug.go
@@ -1,0 +1,32 @@
+package debug
+
+import (
+	"io"
+	"log"
+	"os"
+)
+
+var enabled = false
+var logger = log.New(os.Stderr, "DEBUG: ", log.LstdFlags|log.Lmsgprefix)
+
+// Printf works like log.Printf, but only when debugging is enabled.
+func Printf(format string, v ...interface{}) {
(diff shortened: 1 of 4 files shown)


internal/debug/debug.go
```diff
@@ -0,0 +1,32 @@
...
+	if enabled {
+		logger.Println(v...)
+	}
+}
+
+func SetDebug(b bool) {
+	enabled = b
+}
+
+func SetOutput(w io.Writer) {
```
nprizal, line 30: This is so that we can write log to files in the future. At the moment, it will output to stderr by default.

wooly: I think we should codify that outputting to stdout by default instead of stderr is an explicit decision to conform to 12-factor principles.

wooly (approved):

PR #89, opened 2024-06-12
First pass at Pact consumer tests

wooly:
### Description

To prevents the test splitter API from breaking its promise with the client, we are adding a contract testing using [Pact.io](https://docs.pact.io). This PR replaces the server mocking in `api` package tests with Pact mock provider using pre-defined contract. This contract comes in the form of interactions containing information such as the expected request, the expected response, and a 'state' that the provider needs to be in to honour the interaction. These interactions can be generated as `JSON` file that can be used by test splitter API to verify that its behaviour doesn't break the contracts.


.buildkite/Dockerfile-compile
```diff
@@ -1 +1 @@
-FROM public.ecr.aws/docker/library/golang:1.22.3@sha256:f43c6f049f04cbbaeb28f0aad3eea15274a7d0a7899a617d0037aec48d7ab010
+FROM public.ecr.aws/docker/library/golang:1.21.11
```
nprizal, line 1: pact-go doesn't work with go 1.22.x at the moment.
https://github.com/pact-foundation/pact-go/issues/402

> Linux x86_64 users
>
> Using GoLang 1.22.x ?
> Resolution: disable -race
> Need to use  -race?
> Resolution: downgrade to 1.21.x

nprizal (approved):

PR #92, opened 2024-06-18
Improve server side error handling

nprizal:
We need to improve the robustness of error handling when communicating to the server to fetch or create the test plan. The error handling below applied to all requests made by the client, including fetch test plan, create test plan, and fetch files timing.

- Request errors (redirection error, protocol error, time-out) should be retried using exponential back-off, and fall back to naive test plan after reaching retry time-out.
- 5xx responses should be retried using exponential back-off.
- 429 responses should be retried after rate limit reset. 
- 4xx responses should terminate the client.
- Reading and parsing response errors should terminate the client.

For no 1, 2, 3 the requests should be retried in a loop until it reaches 130 seconds, and fall back to native plan if it's still not successful.

### Implementation
This PR creates a new function called `DoWithRetry` within the `api.Client` struct. The function takes a `ctx context.Context`, `req api.httpRequest` that includes request method, URL, and body, and `v interface{}` which is a pointer to store the response in case of successful request. This function also encodes the request body, creates the http request, and decodes the response body.

`FetchTestPlan`, `CreateTestPlan`, and `FetchFilesTiming` functions were refactored to call `DoWithRetry` function instead of calling `http.Client.Do` directly.

the `main` function is refactored to handle the error cases described above.

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -132,7 +132,7 @@ func (c *Client) DoWithRetry(ctx context.Context, req httpRequest, v interface{}
 
 		debug.Printf("Response code %d", resp.StatusCode)
 
-		// If we get a 429, we should return and rety after the rate limit resets.
+		// If we get a 429, we should return and retry after the rate limit resets.
 		if resp.StatusCode == http.StatusTooManyRequests {
 			if rateLimitReset, err := strconv.Atoi(resp.Header.Get("RateLimit-Reset")); err == nil {
 				r.SetNextInterval(time.Duration(rateLimitReset) * time.Second)
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net/http"
 	"runtime"
 	"strconv"
@@ -14,8 +15,6 @@ import (
 	"github.com/buildkite/roko"
 	"github.com/buildkite/test-splitter/internal/debug"
 )
-
-var ErrRetryTimeout = errors.New("request retry timeout")
 
 // client is a client for the test splitter API.
 // It contains the organization slug, server base URL, and an HTTP client.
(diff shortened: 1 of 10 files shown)


internal/api/client.go
```diff
@@ -52,3 +59,71 @@ func NewClient(cfg ClientConfig) *Client {
...
+	r := roko.NewRetrier(
+		roko.TryForever(),
+		roko.WithStrategy(roko.ExponentialSubsecond(initialDelay)),
+		roko.WithJitter(),
+	)
+
+	retryContext, cancel := context.WithTimeout(ctx, retryTimeout)
+	defer cancel()
+
+	// By default, httpclient read and close the body when the request is sent.
```
nprizal, line 81: @DrJosh9000 Do you have any suggestion for a better way to do this?

internal/api/client.go
```diff
@@ -52,3 +61,120 @@ func NewClient(cfg ClientConfig) *Client {
...
+		// which means there is a network error (e.g. protocol error, timeout),
+		// we should return and retry.
+		if err != nil {
+			debug.Printf("Error sending request: %v", err)
+			return nil, err
+		}
+
+		debug.Printf("Response code %d", resp.StatusCode)
+
+		// If we get a 429, we should return and rety after the rate limit resets.
```
niceking, line 135: typo: retry

niceking (approved):
Looks good! 

With the 130s time limit, that means worst case scenario, a 429 response will be retried twice, is that right? I.e. first request -> 429 with 59s reset -> second request -> 429 with 59 reset -> third request

Does the time limit interrupt whatever request is in progress?

nprizal:
> With the 130s time limit, that means worst case scenario, a 429 response will be retried twice, is that right? I.e. first request -> 429 with 59s reset -> second request -> 429 with 59 reset -> third request

Yup, that's right, at most, `429` response will be retried twice.

> Does the time limit interrupt whatever request is in progress?

Yes, we pass the `context` with 130s timeout to the request, which I believe will cancel the ongoing request. This is the documentation for `NewRequestWithContext` that is being used here.

> // For an outgoing client request, the context
> // controls the entire lifetime of a request and its response:
> // obtaining a connection, sending the request, and reading the
> // response headers and body.

internal/api/client.go
```diff
@@ -52,3 +61,120 @@ func NewClient(cfg ClientConfig) *Client {
...
+		if err != nil {
+			debug.Printf("Error sending request: %v", err)
+			return nil, err
+		}
+
+		debug.Printf("Response code %d", resp.StatusCode)
+
+		// If we get a 429, we should return and retry after the rate limit resets.
+		if resp.StatusCode == http.StatusTooManyRequests {
+			if rateLimitReset, err := strconv.Atoi(resp.Header.Get("RateLimit-Reset")); err == nil {
```
nprizal, line 137: @wooly can you confirm that this is what the API will return in the header when API rate limit is reached.

wooly: Yep, that's it.

PR #93, opened 2024-06-19
Add pact interaction for create test plan with split by example disabled

nprizal:
Create test plan has 2 different scenarios, when split by example is enabled and disabled.
We need to add pact interactions to test that both scenarios are met by the API.

### DIFF
--- a/internal/api/create_test_plan.go
+++ b/internal/api/create_test_plan.go
@@ -20,7 +20,7 @@ var (
 
 type TestPlanParamsTest struct {
 	Files    []plan.TestCase `json:"files"`
-	Examples []plan.TestCase `json:"examples"`
+	Examples []plan.TestCase `json:"examples,omitempty"`
 }
 
 // TestPlanParams represents the config params sent when fetching a test plan.
--- a/internal/api/create_test_plan_test.go
+++ b/internal/api/create_test_plan_test.go
@@ -32,6 +32,108 @@ func TestCreateTestPlan(t *testing.T) {
 			Files: []plan.TestCase{
 				{Path: "sky_spec.rb"},
 			},
+		},
+	}
+
+	err = mockProvider.
+		AddInteraction().
+		Given("A test plan doesn't exist").
+		UponReceiving("A request to create test plan with identifier abc123").
+		WithRequest("POST", "/v2/analytics/organizations/buildkite/suites/rspec/test_plan", func(b *consumer.V2RequestBuilder) {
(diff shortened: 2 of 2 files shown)


wooly:
Want to change the base branch for this?

wooly:
Thanks!

internal/api/create_test_plan_test.go
```diff
@@ -25,6 +25,108 @@ func TestCreateTestPlan(t *testing.T) {
...
+			Files: []plan.TestCase{
+				{Path: "sky_spec.rb"},
+			},
+		},
+	}
+
+	err = mockProvider.
+		AddInteraction().
+		Given("A test plan doesn't exist").
+		UponReceiving("A request to create test plan with identifier abc123").
```
wooly, line 41: Can you add split by example disabled here too for consistency?

internal/api/create_test_plan_test.go
```diff
@@ -25,6 +25,108 @@ func TestCreateTestPlan(t *testing.T) {
 		t.Error("Error mocking provider", err)
 	}
 
+	params := TestPlanParams{
+		Identifier:  "abc123",
```
niceking, line 29: Just a thought: we don't mock the headers in this request so the Pact contracts don't cover permissions? Or check that a token is sent? This is probably fine, just an observation

niceking (approved):

internal/api/create_test_plan.go
```diff
@@ -20,7 +20,7 @@ var (
 
 type TestPlanParamsTest struct {
 	Files    []plan.TestCase `json:"files"`
-	Examples []plan.TestCase `json:"examples"`
+	Examples []plan.TestCase `json:"examples,omitempty"`
```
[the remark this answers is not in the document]
wooly: It changes the payload from:
```
{
  "files": [...],
  "examples": null,
}
```

to 
```
{
  "files: [...],
}
```

So we're not sending a null value for the examples test cases if split by examples is disabled.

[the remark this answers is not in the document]
nprizal: It will omit the `examples` when converted to json.
So instead of `{"files": [...], "examples": null}`, we will be sending `{"files": [...]}`

internal/api/create_test_plan_test.go
nprizal: The interaction expects the token is included in the req header, see line 43.

niceking: Oh nice lol, bad reading on my part

PR #98, opened 2024-06-23
Record Rspec executions and send metadata to Test Analytics

nprizal:
### Description

We are building a UI in Test Analytics to display the test splitting span for each parallel job. To do this, we need the data of each rspec execution's start and finish, including the retry. This PR records each rspec execution's start/finish and send them along with env variables and version to Test Analytics.

The error raised when sending the metadata is suppressed, to prevent the build from failing.

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -164,9 +164,11 @@ func (c *Client) DoWithRetry(ctx context.Context, req httpRequest, v interface{}
 		}
 
 		// parse response
-		err = json.Unmarshal(responseBody, v)
-		if err != nil {
-			return nil, fmt.Errorf("parsing response: %w", err)
+		if v != nil {
+			err = json.Unmarshal(responseBody, v)
+			if err != nil {
+				return nil, fmt.Errorf("parsing response: %w", err)
+			}
 		}
 
 		return resp, nil
--- a/internal/config/env.go
+++ b/internal/config/env.go
@@ -33,3 +33,33 @@ func getIntEnvWithDefault(key string, defaultValue int) (int, error) {
 	// Return the value if it's successfully converted to int.
 	return valueInt, nil
 }
+
+func (c Config) DumpEnv() map[string]string {
+	keys := []string{
+		"BUILDKITE_BUILD_ID",
+		"BUILDKITE_JOB_ID",
+		"BUILDKITE_ORGANIZATION_SLUG",
+		"BUILDKITE_PARALLEL_JOB_COUNT",
+		"BUILDKITE_PARALLEL_JOB",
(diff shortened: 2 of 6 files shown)


main.go
```diff
@@ -172,14 +217,8 @@ func logErrorAndExit(exitCode int, format string, v ...any) {
 
 // fetchOrCreateTestPlan fetches a test plan from the server, or creates a
 // fallback plan if the server is unavailable or returns an error plan.
-func fetchOrCreateTestPlan(ctx context.Context, cfg config.Config, files []string, testRunner TestRunner) (plan.TestPlan, error) {
+func fetchOrCreateTestPlan(ctx context.Context, apiClient *api.Client, cfg config.Config, files []string, testRunner TestRunner) (plan.TestPlan, error) {
 	debug.Println("Fetching test plan")
-	apiClient := api.NewClient(api.ClientConfig{
```
nprizal, line 177: I moved the `apiClient` creation outside this function so we can reuse it in `sendMetadata`

internal/api/client.go
```diff
@@ -164,9 +164,11 @@ func (c *Client) DoWithRetry(ctx context.Context, req httpRequest, v interface{}
 		}
 
 		// parse response
-		err = json.Unmarshal(responseBody, v)
-		if err != nil {
-			return nil, fmt.Errorf("parsing response: %w", err)
+		if v != nil {
```
nprizal, line 167: Some requests might not have a useful response body, so I made this optional

main.go
```diff
@@ -58,8 +58,14 @@ func main() {
 
 	// get plan
 	ctx := context.Background()
+	apiClient := api.NewClient(api.ClientConfig{
+		ServerBaseUrl:    cfg.ServerBaseUrl,
+		AccessToken:      cfg.AccessToken,
+		OrganizationSlug: cfg.OrganizationSlug,
+		Version:          cfg.Version,
```
amybiyuliu, line 65: I saw you removed `version` from config in your other PR https://github.com/buildkite/test-splitter/pull/99/files. Do we need to apply the same change here?

nprizal: Good point!

internal/config/env.go
```diff
@@ -33,3 +33,33 @@ func getIntEnvWithDefault(key string, defaultValue int) (int, error) {
 	// Return the value if it's successfully converted to int.
 	return valueInt, nil
 }
+
+func (c Config) DumpEnv() map[string]string {
+	keys := []string{
+		"BUILDKITE_BUILD_ID",
+		"BUILDKITE_JOB_ID",
```
amybiyuliu, line 40: Do we want to send `BUILDKITE_STEP_ID` as well?

amybiyuliu (approved):

PR #99, opened 2024-06-24
Pass the correct version to api.Client

nprizal:
### Description

The client version is still missing in the user agent after `0.6.0`. This PR fixes the issue by passing the correct version to the `api.Client`


amybiyuliu (approved):
🚀

PR #100, opened 2024-06-24
Release 0.6.2

nprizal:
### Changelog
- Fixed issue where the client version is not set in the user agent


wooly:
Do we need to bump the version and add to the changelog as part of this?

nprizal:
> Do we need to bump the version and add to the changelog as part of this?

just did

amybiyuliu (approved):
🚢

PR #102, opened 2024-06-24
Remove configurability of identifier through `BUILDKITE_SPLITTER_IDENTIFIER` env var

nprizal:
### Description

Test Splitter by default uses `${BUILDKITE_BUILD_ID}/${BUILDKITE_STEP_ID}` as identifier. The identifier can be configured by setting `BUILDKITE_SPLITTER_IDENTIFIER`. Because we are using an identifier in the server as key for the test plan, and to associate the test plan with run/build, we want to remove the ability for the user to override this value, which might cause an issue down the line.


wooly (approved):
I tested this using the integration testing repo here: https://github.com/buildkite/test-splitter-integration-testing

And it is working as expected.

PR #107, opened 2024-06-28
Fix --version for env-less run.

### LINKED ISSUE #106: `test-splitter --version` fails
I tried running `test-splitter --version` to be sure I had installed it right, but it outputs a bunch of errors, no version information, and exits non-zero:

image

wooly:
### Description

Running `--version` on a fresh install with no environment variables should just return the version number.

Fixes #106


nprizal (approved):
👍

sj26 (approved):
Thanks!

PR #108, opened 2024-07-01
Improve logging to show its coming from TSC 

niceking:
### Description

Added the prefix "Buildkite Test Collector" before all error exit messages, and before retries to test failures. Have set retries to always be expanded because we don't support any CI other than Buildkite atm  🤷‍♀️ 

### Context

https://buildkite-corp.slack.com/archives/C05NAGJREJU/p1719808219253619

### DIFF
--- a/main.go
+++ b/main.go
@@ -78,6 +78,7 @@ func main() {
 		runnableTests = append(runnableTests, testCase.Path)
 	}
 
+	fmt.Printf("+++ Buildkite Test Splitter: Running tests")
 	cmd, err := testRunner.Command(runnableTests)
 	if err != nil {
 		logErrorAndExit(16, "Couldn't process test command: %q, %v", testRunner.TestCommand, err)
--- a/main.go
+++ b/main.go
@@ -172,7 +172,8 @@ func retryFailedTests(testRunner TestRunner, maxRetries int, timeline *[]api.Tim
 	retries := 0
 	for retries < maxRetries {
 		retries++
-		fmt.Printf("Attempt %d of %d to retry failing tests\n", retries, maxRetries)
+
+		fmt.Printf("+++ Buildkite Test Splitter: ♻️ Attempt %d of %d to retry failing tests\n", retries, maxRetries)
 
 		cmd, err := testRunner.RetryCommand()
 		if err != nil {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -172,7 +172,8 @@ func retryFailedTests(testRunner TestRunner, maxRetries int, timeline *[]api.Tim
 	retries := 0
 	for retries < maxRetries {
 		retries++
-		fmt.Printf("Attempt %d of %d to retry failing tests\n", retries, maxRetries)
+
+		fmt.Printf("+++ Buildkite Test Collector: ♻️ Attempt %d of %d to retry failing tests\n", retries, maxRetries)
```
wooly, line 176: Should this be `Buildkite Test Splitter` like the below?

wooly (approved):
General comment around Collector/Splitter language, but not blocking.

main.go
niceking: Oh lol

main.go
```diff
@@ -173,7 +172,8 @@ func retryFailedTests(testRunner TestRunner, maxRetries int, timeline *[]api.Tim
 	retries := 0
 	for retries < maxRetries {
 		retries++
-		fmt.Printf("Attempt %d of %d to retry failing tests\n", retries, maxRetries)
+
+		fmt.Printf("+++ Buildkite Test Splitter: ♻️ Attempt %d of %d to retry failing tests\n", retries, maxRetries)
```
wooly, line 176: Persnickety old man in the corner here but how come this one has `+++` at the start but the logErrorAndExit one doesn't?

niceking: My understanding of `+++` is that the group is automatically expanded? So all the re-run spec tests will automatically be expanded in the job log, rather than having to open the group to view the rerun specs. Whereas I think because the logErrorAndExit log lines are discrete and not part of a group they just get printed anyways? I could be very wrong though

wooly: Ah I see now. In that case do we want to have another group for the initial run? Or do you think it doesn't matter?

niceking: Added a group for this, the errors that might occur from the test cmd being wrong will be printed after this group

wooly:
LGTM!

PR #109, opened 2024-07-02
Refactor discovery to be more runner-agnostic.

wooly:

### DIFF
--- a/internal/runner/discover.go
+++ b/internal/runner/discover.go
@@ -3,6 +3,7 @@ package runner
 import (
 	"fmt"
 	"io/fs"
+	"os"
 
 	"github.com/DrJosh9000/zzglob"
 )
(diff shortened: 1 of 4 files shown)


internal/runner/discover.go
```diff
@@ -12,7 +13,20 @@ type DiscoveryPattern struct {
 	ExcludePattern string
 }
 
-func discoverTestFiles(pattern DiscoveryPattern) ([]string, error) {
+func discoverTestFiles(defaultIncludePattern string) ([]string, error) {
+	includePattern := os.Getenv("BUILDKITE_SPLITTER_TEST_FILE_PATTERN")
```
nprizal, line 17: I have a refactor idea related to this part which I have captured in this ticket
https://linear.app/buildkite/issue/TAT-229/refactor-the-fetching-of-includeexclude-pattern-from-env-variables

PR #115, opened 2024-07-04
Update changelog link 

niceking:
We deleted the migration instructions from 0.4.0 cos it only applied to TC, but they noticed the link didn't work so make the changelog ref to this a permalink


niceking:
This doesn't need to go in a release cos its a documentation only change...can this just go straight on main? Or would that muck with the branch management?

nprizal (approved):

amybiyuliu (approved):

PR #122, opened 2024-07-15
Add background process to fetch test plan and cancel running

nprizal:
### Description
Currently, when the test plan is missing from the cache (i.e. `GET /test_plan` return `NotFound`), the client will attempt to create a new plan. The process to create a new plan might take long, especially when the Split By Example is enabled, because the client will execute dry run to get the list of examples. While the client is attempting to create a new plan, we want the client to fetch the test plan periodically in the background in case the other node successfully created the plan. If the plan is found in the background, the client should cancel all process to create a new plan.

### Changes
Fetch test plan in the background and cancel ongoing test plan creation process.

### DIFF
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -1,6 +1,7 @@
 package runner
 
 import (
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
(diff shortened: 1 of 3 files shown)


wooly:
:+1:

wooly:
Want to delete the branch too?

PR #125, opened 2024-07-16
Publish Docker image to Buildkite Packages and Dockerhub

nprizal:
### Description
We want to create a Docker image that contains `test-splitter` binary and publish it to Dockerhub during the release.
We also want to create a nightly build image and publish it to private Buildkite Packages repo for internal use.

This PR introduce a branching in the release pipeline that will upload different sets of job depending on release. phase
1. Nightly build. Triggered when the tag for the version is already exists in git
   - Push docker image to Buildkite Packages
   - The image tag will have version number and short commit sha, e.g. `0.7.7-aa0e23`.
1. Stable. Triggered when the tag for the version is **not** already exists in git
   - Push docker image to Buildkite Packages
   - Push docker image to Dockerhub
   - Create GitHub Release
   - The image tag will only have version number, e.g. `0.7.7`.

I also added a block step as a safety mechanism before running the job to create GitHub Release, publish to Dockerhub, and to other potential public distribution in the future.
 

Following graph shows the build flow for `Test Spliter Client - Release` pipeline, that will be triggered on _every_ commit in `main` branch.

image

### Verification
I triggered a build in `Test Splitter Client - Release` for this branch, and this is how the [build](https://buildkite.com/buildkite/test-splitter-client-release/builds/108) looks like for nightly build.

### DIFF
--- a/.buildkite/pipeline.release.stable.yml
+++ b/.buildkite/pipeline.release.stable.yml
@@ -1,0 +1,26 @@
+- name: ":buildkite: :package: Publish Docker image to Buildkite Packages"
+  command: ".buildkite/steps/release-buildkite-packages.sh"
+
+- wait 
+
+- block: ":rocket: Create stable release?"
+  key: create-release
+  blocked_state: "passed"
+
+- name: ":octocat: :rocket: Create Github Release"
+  command: ".buildkite/steps/release-github.sh"
+  depends_on: create-release
+  plugins:
+    - aws-assume-role-with-web-identity:
+        role-arn: arn:aws:iam::445615400570:role/pipeline-buildkite-test-splitter-client-release
+    - aws-ssm#v1.0.0:
+        parameters:
+          GH_TOKEN: /pipelines/buildkite/test-splitter-client-release/GH_TOKEN
+    - docker-compose#v4.14.0:
+        config: .buildkite/docker-compose.yml
+        cli-version: 2
+        run: release
+        mount-checkout: true
+        mount-buildkite-agent: true
+        env:
+          - GH_TOKEN
--- a/.buildkite/pipeline.release.stable.yml
+++ b/.buildkite/pipeline.release.stable.yml
@@ -24,3 +24,14 @@
         mount-buildkite-agent: true
         env:
           - GH_TOKEN
+
+- name: ":docker: :rocket: Push Docker image to Docker Hub"
+  command: ".buildkite/steps/release-dockerhub.sh"
+  depends_on: create-release
+  plugins:
+    - aws-assume-role-with-web-identity:
+        role-arn: arn:aws:iam::445615400570:role/pipeline-buildkite-test-splitter-client-release
+    - aws-ssm#v1.0.0:
+        parameters:
+          DOCKERHUB_USER: /pipelines/buildkite/test-splitter-client-release/dockerhub-user
+          DOCKERHUB_PASSWORD: /pipelines/buildkite/test-splitter-client-release/dockerhub-password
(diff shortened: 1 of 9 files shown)


.buildkite/steps/publish-docker-image-ecr.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+splitter_version="$(cat version/VERSION)"
+builder_name=$(docker buildx create --use)
+image_tag="445615400570.dkr.ecr.us-east-1.amazonaws.com/test-splitter:$splitter_version-$BUILD_NUMBER"
```
wooly, line 7: Will this be the build number of the release pipeline or the build number of the test splitter pipeline? I think it should be the test splitter pipeline build number.

nprizal: It's the build number of the test splitter pipeline, we pass this when we trigger the release pipeline.

version/VERSION
```diff
@@ -1 +1 @@
-0.7.2
+0.7.3-rc
```
wooly, line 1: This feels ever so slightly off. Is it possible to build it without updating this to add the -rc suffix?

.buildkite/pipeline.release.stable.yml
```diff
@@ -0,0 +1,18 @@
+- name: ":octocat: Create Github Release"
```
nprizal, line 1: TODO: add step to publish docker image to Dockerhub

version/VERSION
nprizal: My thinking is so we know whether we should do stable (public) or private release.
If we don't want `-rc` suffix, we could check if the tag for that version exists or not, if tag already exists, then we to private release, otherwise it's a stable release. But it means that changes that will be released in `0.7.3` will have a version `0.7.2-123` instead. Thoughts?

wooly: I think this is fine - this appears to be the way that the agent is doing it as well: https://github.com/buildkite/agent/blob/main/version/VERSION

.buildkite/pipeline.release.stable.yml
```diff
@@ -0,0 +1,37 @@
+- name: ":buildkite: :package: Publish Docker image to Buildkite Packages"
```
nprizal, line 1: We push the stable version to Buildkite Packages as well, so we can use it internally. And it will make it easier to switch between nightly build and stable version internally. We only need to change the version number.

.buildkite/steps/release-github.sh
```diff
@@ -2,13 +2,7 @@
 
 set -euo pipefail
 
-splitter_version="$(cat version/VERSION)" 
-
-# Skip release if version already exists
```
nprizal, line 7: This check has moved to the `release-prepare.sh`. We will not run this job if the tag is already exists.

.buildkite/steps/release-prepare.sh
```diff
@@ -0,0 +1,21 @@
...
+version="$(cat version/VERSION)"
+
+echo "--- :label: Checking git tag"
+
+# If the tag is already exists, we assume the release is for nightly build,
+# otherwise we assume it's a stable release
+if git ls-remote --tags origin | grep "refs/tags/v${version}" ; then
+  echo "v${version} already exists, running nightly release"
+  buildkite-agent meta-data set "release-phase" "nightly"
+  buildkite-agent meta-data set "version" "$version-$(git rev-parse --short HEAD)"
```
nprizal, line 14: @wooly we cannot trigger build in `Test Splitter Client - Release` pipeline from the `Test Splitter Client` pipeline, therefore we can not get the build number. I decided to tag it with the short commit sha instead. What is your thought? We can use the build number of the `Test Splitter Client - Release` pipeline, but seems a bit off to me.

PR #126, opened 2024-07-16
Hard-exit the test splitter if the runner exits abnormally, e.g. SIGABRT

wooly:
### Description

We have an exit status problem with the test splitter if the runner exits abnormally due to an OS-level signal (SIGSEGV or some such).

This PR adds signal detection to both the runner and the retries to exit more cleanly if the runner fails miserably to execute.

### Context

Linear: https://linear.app/buildkite/issue/TAT-268/nuclear-rspec-command-should-not-be-retried

### Changes

- Handle signals in the runner and exit cleanly from the test splitter with an exit status equal to 128 + the signal number.
- Update the README.md file to reflect these changes.

### Testing

Tested locally with `test-splitter-integration-testing` repository. Will cause bk/bk builds to fail on the coredump error instead of job-level retries as the exit codes are not clear. This will be handled in bk/bk before shipping the upgrade PR.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,12 +62,18 @@ steps:
     env:
       BUILDKITE_SPLITTER_SUITE_SLUG: my-suite
       BUILDKITE_SPLITTER_API_ACCESS_TOKEN: your-secret-token
-``` 
+```
 
-### Exit code
-| Exit code | Description |
-| ---- | ---- |
-| 0 | Success (passed through from test runner) |
-| 1 | Failure (passed through from test runner) |
-| 16 | Test Splitter failure (e.g., config error) |
-| * | Other errors (passed through from the test runner) |
+### Possible exit statuses
+
+The test-splitter client may exit with a variety of exit statuses, outlined below:
+
+- If there is a configuration error, the test-splitter client will exit with
+  status 16.
+- If the test-splitter runner (such as RSpec) exits cleanly, the exit status of
+  the runner is returned. This will likely be 0 for successful test runs, 1 for
+  failing test runs, but may be any other error status returned by the runner.
+- If the test-splitter runner raises an OS level signal, such as SIGSEGV or
+  SIGABRT, the exit status returned will be equal to 128 plus the signal number.
+  For example, if the runner raises a SIGSEGV, the exit status will be (128 +
+  11) = 139.
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ The test-splitter client may exit with a variety of exit statuses, outlined belo
 - If the test-splitter runner (such as RSpec) exits cleanly, the exit status of
   the runner is returned. This will likely be 0 for successful test runs, 1 for
   failing test runs, but may be any other error status returned by the runner.
-- If the test-splitter runner raises an OS level signal, such as SIGSEGV or
+- If the test-splitter runner is terminated by an OS level signal, such as SIGSEGV or
   SIGABRT, the exit status returned will be equal to 128 plus the signal number.
   For example, if the runner raises a SIGSEGV, the exit status will be (128 +
   11) = 139.
--- a/main.go
+++ b/main.go
@@ -27,6 +27,7 @@ type TestRunner interface {
 	GetExamples(files []string) ([]plan.TestCase, error)
 	GetFiles() ([]string, error)
 	RetryCommand() (*exec.Cmd, error)
+	Name() string
 }
 
 func main() {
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -129,15 +130,20 @@ func main() {
 	if err != nil {
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
 			exitCode := exitError.ExitCode()
+			if exitCode == -1 {
```
amybiyuliu, line 133: I'm thinking about we could check if Rspec exit code is not 1 (Rspec exits due to test failure) before entering the code block for retry. Maybe this would give us more confidence the retry command gets executed only when there's test failure?

```
if exitCode != 1 {
	logErrorAndExit(exitCode, "%s exited abnormally, cannot continue", testRunner.Name())
}
```

wooly: I've updated this so instead of returning if the exit code is `-1`, it will return early (it'll still exit with -1 though) if a signal terminated the runner process.

main.go
```diff
@@ -129,15 +131,21 @@ func main() {
 	if err != nil {
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
 			exitCode := exitError.ExitCode()
+
+			if status, ok := exitError.Sys().(syscall.WaitStatus); ok && status.Signaled() {
+				logErrorAndExit(-1, "%s was terminated with signal: %v", testRunner.Name(), status.Signal())
```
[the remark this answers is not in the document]
moskyb: also! i found that the [`sys/unix`](https://pkg.go.dev/golang.org/x/sys/unix#SignalName) package provides a way to get the short signal name, so you could probably do something like

```suggestion
				logErrorAndExit(-1, "%s was terminated with signal: %s (%v)", testRunner.Name(), unix.SignalName(status.Signal(), status.Signal())
```

which would make the error string something like:
```
Rspec was terminated with signal: SIGSEGV (segmentation fault)
```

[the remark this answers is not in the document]
moskyb: will any of this work on windows? great question.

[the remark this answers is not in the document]
wooly: Thanks massively for the input @moskyb. I've heard @pda mention the 128 + signal number thing before as well, so thanks Paul.

I've updated the exit codes and attempted to clarify things in the README.

wooly:
@moskyb would super duper appreciate if you can give this one last look!

README.md
```diff
@@ -62,12 +62,18 @@ steps:
...
+
+- If there is a configuration error, the test-splitter client will exit with
+  status 16.
+- If the test-splitter runner (such as RSpec) exits cleanly, the exit status of
+  the runner is returned. This will likely be 0 for successful test runs, 1 for
+  failing test runs, but may be any other error status returned by the runner.
+- If the test-splitter runner raises an OS level signal, such as SIGSEGV or
+  SIGABRT, the exit status returned will be equal to 128 plus the signal number.
+  For example, if the runner raises a SIGSEGV, the exit status will be (128 +
+  11) = 139.
```
moskyb, line 79: maybe this would be a bit clearer as:

```suggestion
- If the test-splitter runner is terminated by an OS level signal, such as SIGSEGV or
  SIGABRT, the exit status returned will be equal to 128 plus the signal number.
  For example, if the runner raises a SIGSEGV, the exit status will be (128 +
  11) = 139.
```

main.go
```diff
@@ -129,15 +132,21 @@ func main() {
 	if err != nil {
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
 			exitCode := exitError.ExitCode()
+
+			if status, ok := exitError.Sys().(syscall.WaitStatus); ok && status.Signaled() {
+				logSignalAndExit(testRunner.Name(), status.Signal())
+			}
+
```
moskyb, line 139: `exitError.Sys()` will return something castable to `syscall.WaitStatus` - ie, the value of `ok` will be `true` - only on unixes, which i suspect will mean that on windows (the only notable non-unix that we might be expected to ship the test splitter to), behaviour will be undefined.

i think that this is probably totally fine, but i figured i'd point it out anyway. it'll just mean that when something is terminated with a signal on windows (does that even happen? i don't know) we'll likely return an exit code of -1, which again, i think is probably fine.

moskyb (approved):
LGTM! i have a minor wording nit on the readme, and an informational thing about `syscall.WaitStatus`, but happy to ship as is :)

nprizal (approved):
Changes looks good to me 👍

main.go
wooly: Legend, thank you so much!

PR #127, opened 2024-07-17
Move test splitter client pipeline to open source cluster

nprizal:
### Description
To follow Buildkite best practice on managing an Open Source, we need to move the pipelines to Buildkite open source cluster. Currently, we have 2 pipelines for Test Splitter Client.
1. Test Splitter Client (public)
2. Test Splitter Client - Release (private)

`Test Splitter Client` pipeline should be moved to `Open Source` cluster, while the `Test Splitter Client - Release` should be moved to `Open Source Deploy` cluster and remain private.

Because we cannot trigger a build for pipeline on a different cluster, this changes how we trigger the release pipeline.

**Before**
  1. Commit on `main` branch triggers a build (via GitHub webhook) in `Test Splitter Client` pipeline that uploads `.buildkite/pipeline.yml`.
  1. The `yml` file includes steps to run the tests and compile the binaries
  1. After all steps in that build passed, it triggers another build (via [Trigger Step](https://buildkite.com/docs/pipelines/trigger-step)) in `Test Splitter Client - Release` pipeline that uploads `.buildkite/pipeline.release.yml`.
  1. The `yml` file includes steps to download the compiled binaries from triggering build (see no. 2)and create GitHub Release

**After**
1. Commits on `main` branch triggers a build (via GitHub webhook) in `Test Splitter Client - Release` pipeline that upload `.buildkite/pipeline.yml`
1. The `yml` file includes steps to run the tests and compile the binaries
1. After all the steps passed, it uploads `.buildkite/pipeline.release.yml` to the same build
3. The `yml` file includes steps to download the compiled binaries from the previous step and create GitHub Release

Note: commits to branches other than `main` branch will not trigger build in `Test Splitter Client - Release`

Sample builds:
[Test Splitter Client - Release](https://buildkite.com/buildkite/test-splitter-client-release/builds/87)
[Test Splitter Client](https://buildkite.com/buildkite/test-splitter-client/builds/664)

### Changes
For this new process to work, following changes are necessary
- Setup a webhook in GitHub to trigger build in `Test Splitter Client - Release`
Screenshot 2024-07-18 at 3 04 57 PM

- Update `Test Splitter Client - Release` pipeline config to upload `.buildkite/pipeline.yml` and `.buildkite/pipeline.release.yml`
Screenshot 2024-07-18 at 3 06 40 PM

- Add branch filtering on `Test Splitter Client - Release` pipeline to only build on `main`
Screenshot 2024-07-18 at 3 07 26 PM


niceking:
I _think_ this makes sense but a few questions and clarifications:

When do we create builds on `Test Splitter Client`? Has that changed? Are we enabling third party PRs?

Commits on main will trigger releases, that makes sense. Are the released indexed to tags? Do we need to make a new tag everytime we commit to main? Also, only Buildkite engineers can merge PRs right?

nprizal:
> When do we create builds on Test Splitter Client? Has that changed? Are we enabling third party PRs?

Nothing changed, every commit will trigger build on `Test Splitter Client`. We haven't enabled third party PRs at the moment.

> Commits on main will trigger releases, that makes sense. Are the released indexed to tags? Do we need to make a new tag everytime we commit to main?

Yes, the release is indexed to a git tag, this is the current behaviour. The GitHub Release step will create the tag, so we don't have to it manually. 

Also, to clarify, the release will only be created if the tag hasn't been created for that version. This behaviour is unchanged.

> Also, only Buildkite engineers can merge PRs right?

I believe, yes, we must've configured this on the repo setting.

nprizal:
For extra safety, we could also have a block step before the release step in the `Test Splitter Client - Release` pipeline.

sexydaikon:
image

niceking:
(i'm sexydaikon btw, was just testing that external users can't merge PRs in this repo)

amybiyuliu:
Thanks for write up the PR description! I think I followed the idea of the change.
I just have a question about the commits on main, would that still trigger a build on the `Test Splitter Client` pipeline? If it will still be trigger, I'd imagine it won't create the github release is that right?

nprizal:
> I just have a question about the commits on main, would that still trigger a build on the Test Splitter Client pipeline? If it will still be trigger, I'd imagine it won't create the github release is that right?

Yes, it will trigger the build on the `Test Splitter Client` pipeline as well, but it will not run the release steps. Also, that pipeline is inside the Open Source cluster that doesn't have any credentials, so it is safe.

niceking (approved):

PR #131, opened 2024-07-23
Create more sophisticated release pipeline for tsc

nprizal:
### Description
Use goreleaser to build binaries, build and push docker image, and create GitHub release.
This PR also enables us to release a _pre-release_ version.

There will be a "Prepare Release" step at the beginning of every build in `Test Splitter Client - Release` pipeline that will determine the next available versions for the release, and create a block step before eventually creating the release. The block step is optional, and skipping the step will not create a release.

Screenshot 2024-07-24 at 5 25 56 PM

After selecting the version, it will create the release using `goreleaser` that will do the followings:
- build binaries for all support os and architecture
- build and push Docker image to Buildkite packages including `latest` tag for both stable and pre-release version
- build and push Docker image to Dockerhub including `latest` tag for stable version only
- create GitHub release with the binaries artifact

**Caveat**
The GitHub release needs to be mark as latest manually, because `goreleaser` cannot set the GitHub release only for _stable_ version.

### Verification
This is a fully working [build](https://buildkite.com/buildkite/test-splitter-client-release/builds/131) that I triggered manually, and these are the releases that were created by that build.
- [GitHub release](https://github.com/buildkite/test-splitter/releases/tag/untagged-c2644dd78b96bc1ae256)
- [Dockerhub](https://hub.docker.com/r/buildkite/test-splitter/tags)
- [Buildkite Packages](https://buildkite.com/organizations/buildkite/packages/registries/test-splitter-docker)


.goreleaser.yaml
```diff
@@ -0,0 +1,57 @@
+version: 2
+
+release:
+  name_template: Test Splitter v{{.Version}}
+  draft: true
```
nprizal, line 5: We may consider publishing the GitHub release instead of creating a draft.
Thought? @wooly

.buildkite/steps/prepare-release.sh
```diff
@@ -0,0 +1,46 @@
...
+
+current=$(svu current)
+
+
+options() {
+  if [[ "$current" =~ "-rc" ]]; then
+    # if current version is 0.7.5-rc.1
+    # the options should be
+    # - 0.7.5
+    # - 0.7.5-rc.2
```
nprizal, line 11: Do you think this make sense?
I think if the current version is a pre-release, the option is either promote it as a _stable_ or increment the pre-release version.

.goreleaser.yaml
wooly: What's the difference between draft and when github marks it as pre-release, is there one?

.buildkite/steps/prepare-release.sh
wooly: Yeah I think that makes sense :+1:

.goreleaser.yaml
nprizal: Draft release is not accessible by the public, and git `tag` will not be created for that release.

nprizal: I think my question is if we want to publish draft manually, or automate that. We've been automatically publish the draft, so maybe we should keep that behavior?

.buildkite/steps/prepare-release.sh
```diff
@@ -0,0 +1,46 @@
...
+    cat <<YAML
+        - label: $(svu patch)
+          value: $(svu patch)
+        - label: $(svu prerelease)
+          value: $(svu prerelease)
+YAML
+  else
+    # if current version is 0.7.5
```
amybiyuliu, line 19: We currently don't need to worry about releasing major version right? We could add that before releasing the first major version

.goreleaser.yaml
wooly: Ah, I reckon we don't need to produce a draft release? We just unblock a new patch or patch rc when we want to produce a pre-release version

.buildkite/steps/prepare-release.sh
nprizal: Yes, I think the plan is to have a mechanism to override that options, maybe type the version number manually on special circumstance.

.buildkite/pipeline.yml
```diff
@@ -22,17 +20,21 @@ steps:
...
-      key: build-binary
-      artifact_paths: "pkg/*"
+      artifact_paths: "dist/test-splitter-*"
       plugins:
-        docker-compose#v4.14.0:
-          config: .buildkite/docker-compose.yml
-          cli-version: 2
-          run: build
+        docker#v5.11.0:
+          image: goreleaser/goreleaser:v2.1.0
```
jradtilbrook, line 29: Oh I see you already have a matrix here. How is this working? I thought goreleaser pro only supported target builds

nprizal: I use `--single-target` option that will build the binary only for specific GOOS and GOARCH. The non-pro version however doesn't support split & merge release.
By the way, this build is for the CI pipeline, we don't use this for the release because we have to rebuild the binaries with correct version.

.buildkite/pipeline.release.yml
```diff
@@ -1,17 +1,32 @@
...
-      - docker-compose#v4.14.0:
-          config: .buildkite/docker-compose.yml
-          cli-version: 2
-          run: release
-          mount-checkout: true
+            GITHUB_TOKEN: /pipelines/buildkite/test-splitter-client-release/GH_TOKEN
+            DOCKERHUB_USER: /pipelines/buildkite/test-splitter-client-release/dockerhub-user
+            DOCKERHUB_PASSWORD: /pipelines/buildkite/test-splitter-client-release/dockerhub-password
+      - docker#v5.11.0:
+          image: goreleaser/goreleaser:v2.1.0
```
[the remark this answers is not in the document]
nprizal: Yeah I see that buildkite/cli uses the pro version. I'll see if I can do the same here.

.buildkite/pipeline.yml
jradtilbrook: Oh right yes okay

.buildkite/pipeline.release.yml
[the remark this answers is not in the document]
jradtilbrook: You might get a benefit 🤷‍♂️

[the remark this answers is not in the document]
nprizal: I have an issue getting the `GORELEASER_KEY` from the agent secret. I will defer this, and create ticket in linear. It takes 2 minutes without the parallelization, so I don't think we have a rush to parallelize this release.

wooly (approved):

niceking:
This is really awesome work @nprizal 🥳

PR #135, opened 2024-07-25
If err != nil, don't discover that path

DrJosh9000:
### Description

To enable debugging things like permissions issues that prevent Glob from walking a directory that matches the pattern, Glob calls the callback with paths if there was an error. This currently happens even with file paths that don't fully match the pattern (a zzglob bug). 

Regardless, errors being passed into the callback should cause the path to not be discovered.

### Context

Bug report from beta user

### Changes

`return nil` within the error handling block.

### Testing

No additional testing


wooly:
Legend, I was able to reproduce this with my little go program when I changed the ownership of some files to root.

wooly (approved):

PR #137, opened 2024-07-25
Add more logging for file discovery when in debug mode

wooly:
The debug logs are a bit lacking for the file discovery phase, so this makes it a bit louder.

```
2024/07/26 10:25:35 DEBUG: Discovering test files with include pattern: spec/**/*_spec.rb exclude pattern:
2024/07/26 10:25:35 DEBUG: Discovered 6 files
```


nprizal (approved):
Nice!

PR #143, opened 2024-07-29
Refactor test file and exclude pattern handling

nprizal:
Currently, the `BUILDKITE_SPLITTER_TEST_FILE_PATTERN` and `BUILDKITE_SPLITTER_TEST_FILE_EXCLUDE_PATTERN` are fetched from runner/rspec.go. To be consistent with other config, I moved the fetching to the `config/read.go`.

The default value for include pattern still be defined in `runner/rspec.go`.

### DIFF
--- a/internal/api/client_test.go
+++ b/internal/api/client_test.go
@@ -185,7 +185,7 @@ func TestDoWithRetry_RequestError(t *testing.T) {
 
 func TestDoWithRetry_429(t *testing.T) {
 	originalTimeout := retryTimeout
-	retryTimeout = 2 * time.Second
+	retryTimeout = 1500 * time.Millisecond
 	t.Cleanup(func() {
 		retryTimeout = originalTimeout
 	})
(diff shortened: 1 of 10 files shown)


internal/api/client_test.go
```diff
@@ -185,7 +185,7 @@ func TestDoWithRetry_RequestError(t *testing.T) {
 
 func TestDoWithRetry_429(t *testing.T) {
 	originalTimeout := retryTimeout
-	retryTimeout = 2 * time.Second
```
nprizal, line 188: This fix the flaky test

amybiyuliu: Mind me ask why this test was flaky with retry timeout being 2s?

amybiyuliu (approved):
Nice!

internal/api/client_test.go
nprizal: Good question. The timeout is set to 2 second, and the retry interval (`RateLimit-Reset`) is set to 1 second. The test is asserting that the request is made exactly 2 times, but in some situation, it made the request 3 times. (I'm not 100% sure how this happens, maybe because its so quick that the request is made in second 0, 1, and 2).
Reducing the timeout to `1.5 seconds` ensure that the request only made twice.

PR #147, opened 2024-08-02
Add support for specifying the runner using env-var.

wooly:
### Description

We want to support multiple runners in the future (like Jest, Cypress, etc), but to do this we have to be able to tell the test splitter which runner we want to use. We could be smart and infer this from the TEST_CMD, but for now let's just KISS and make it configurable via an env var called `BUILDKITE_SPLITTER_TEST_RUNNER`. If the runner is set to anything but `rspec`, the splitter will exit with 16 and appropriate message.

### Changes

- Add `BUILDKITE_SPLITTER_TEST_RUNNER` env var, default `rspec`.

### Testing

Test splitter will still work


internal/runner/rspec.go
```diff
@@ -23,7 +23,7 @@ type Rspec struct {
 	RetryTestCommand       string
 }
 
-func NewRspec(r Rspec) Rspec {
+func NewRspec(r Rspec) *Rspec {
```
nprizal, line 26: 👍

wooly:
There's probably a bunch more tidy-up I can do WRT the shared interface or the runners, but this at least gets us a small step towards a jest-runner world.

niceking (approved):

niceking:
We've got some stuff server side that hardcodes the runner that could be cleaned up after this is released and used by bk/bk/bk https://github.com/buildkite/buildkite/blob/main/app/models/test_splitting/test_plan/params_transformer.rb#L16

wooly:
Oh nice, I might also update the test plan payload being sent to include the runner so we can remove that.

wooly:
In fact I've just realised that splitter versions prior to the one that sends the runner in the payload will still need a fallback case, so that code would have to stay in bk/bk, but read from the params if present.

wooly:
@DrJosh9000 @moskyb any chance one of you could give this a quick look to make sure I'm not doing anything I shouldn't be doing?

DrJosh9000:
LGTM, my only suggestion would be to add a `go generate` check to the pipeline similar to https://github.com/buildkite/agent/blob/c714bbfc6a2f58d6bbb5f446fbf8bdf4933254c9/.buildkite/steps/check-code-committed.sh#L25-L35

wooly:
Thanks @DrJosh9000, I'll do that now!

DrJosh9000 (approved):
Sweet

PR #155, opened 2024-08-05
Shuffle things around in preparation for Jest runner.

wooly:
### Description

This PR moves the `TestRunner` interface to the detector and updates the return type of the `DetectRunner` to the interface instead of a pointer to the [Rspec struct](https://github.com/buildkite/test-splitter/blob/main/internal/runner/rspec.go#L19-L24). This will hopefully allow me to add the Jest runner (which will also implement the `TestRunner` interface more easily.

I'm wondering if there's a more ergonomic way of doing this? Also wondering if there's a neater way to handle the aforementioned Rspec struct since I'll need a Jest one that's exactly the same.

Any pointers appreciated!

### DIFF
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -20,7 +20,7 @@ type Rspec struct {
 	cfg RunnerConfig
 }
 
-func NewRspec(cfg *RunnerConfig) *Rspec {
+func NewRspec(cfg RunnerConfig) Rspec {
 	if cfg.TestCommand == "" {
 		cfg.TestCommand = "bundle exec rspec {{testExamples}}"
 	}
@@ -40,8 +40,8 @@ func (r Rspec) Name() string {
 
 // GetFiles returns an array of file names using the discovery pattern.
 func (r Rspec) GetFiles() ([]string, error) {
-	debug.Println("Discovering test files with include pattern:", r.TestFilePattern, "exclude pattern:", r.TestFileExcludePattern)
-	files, err := discoverTestFiles(r.TestFilePattern, r.TestFileExcludePattern)
+	debug.Println("Discovering test files with include pattern:", r.cfg.TestFilePattern, "exclude pattern:", r.cfg.TestFileExcludePattern)
+	files, err := discoverTestFiles(r.cfg.TestFilePattern, r.cfg.TestFileExcludePattern)
 	debug.Println("Discovered", len(files), "files")
 
 	// rspec test in Test Analytics is stored with leading "./"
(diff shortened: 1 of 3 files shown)


nprizal (commented):
The common practice in golang is to define the interface in the package that consume the value, in this case, the `main` package. But I think it's fine in this situtation because we want `runner` package return specific value that implement the interface based on the input it receives.

> I'm wondering if there's a more ergonomic way of doing this? Also wondering if there's a neater way to handle the aforementioned Rspec struct since I'll need a Jest one that's exactly the same.

If we know for sure that the config for `rspec` is the same for `jest` or other runner, we can maybe create `RunnerConfig` struct to initizalize the runners.
```
type RunnerConfig struct {
  TestCommand string
  TestFilePattern string
  TestFileExcludePattern string
  RetryTestCommand string
}

func DetectRunner(runner string, cfg RunnerConfig) (TestRunner, error) {
  switch runner {
    case rspec:
      return NewRspec(cfg)
    case jest:
      return NewJest(cfg)
  }
}

// rspec
type Rspec struct {
  cfg RunnerConfig
}

func NewRspec(cfg *RunnerConfig) *Rspec {
  if cfg.TestCommand == "" {
     cfg.TestCommand = "bundle exec rspec {{testExample}}"
  }
  ...
  return Rspec{
    cfg: cfg
  }
  
}

type (r Rspec) SomeFunc() {
  ...
}

// jest
type Jest struct {
  cfg RunnerConfig
}

func NewJest(cfg RunnerConfig) {
  if cfg.TestCommand == "" {
     cfg.TestCommand = "jest {{testExample}}"
  }
  ...
  return Jest{
    cfg: cfg
  }
}

type (j Jest) SomeFunc() {
  ...
}
```

internal/runner/rspec.go
```diff
@@ -17,22 +17,21 @@ import (
...
 
-	if r.TestFilePattern == "" {
-		r.TestFilePattern = "spec/**/*_spec.rb"
+	if cfg.TestFilePattern == "" {
+		cfg.TestFilePattern = "spec/**/*_spec.rb"
 	}
 
-	return &r
+	return Rspec{
+		cfg: cfg
```
wooly, line 33: @nprizal looks like there's an issue with this line, but I'm not good enough at go to work it out. Do you have any ideas?

DrJosh9000:
```suggestion
		cfg: cfg,
```

wooly:
Superseded by #166

PR #156, opened 2024-08-06
Parse Rspec output and retry failed tests explicitly

nprizal:
### Description
Currently, Test Splitter Client doesn't know the result of the Rspec execution. When Rspec exits with non-zero code, it assumes that there is a failed test, and will retry with `--only-failure` option. This means that if Rspec exits with non-zero code _outside the test_ (e.g. configuration error, crash), TSC will attempt to retry with 0 tests. This cause the job to pass when it should not.

We need to change the behavior to explicitly retry failed tests. To do this, we first need to store and parse the Rspec output to get the failed tests, and execute rspec by passing the failed tests explicitly.

### Changes

- Extract the signal forwarding into `runAndForwardSignal` function, so we can reuse it for the retry command (currently, the signal forwarding only applied to the original test command). This will also allow us to unit test some behaviors, including when the command crash.
- Remove `Command` and `RetryCommand` from `Rspec` struct and introduce `Run` function that will create the command, executes the command, parse the result, and handle errors.
- Implement `runTestsWithRetry` in the `main` function that will call `testRunner.Run` and retry when there are failed tests.
- Change the default value of retry command to be the same as the test command.


internal/runner/rspec.go
```diff
@@ -63,64 +67,72 @@ func (r Rspec) GetFiles() ([]string, error) {
...
+	commandName, commandArgs, err := commandNameAndArgs(command, testCases)
+	if err != nil {
+		return TestResult{Status: TestStatusError}, fmt.Errorf("failed to build command: %w", err)
+	}
 
-// Command returns an exec.Cmd that will run the rspec command
-func (r Rspec) Command(testCases []string) (*exec.Cmd, error) {
-	commandName, commandArgs, err := r.commandNameAndArgs(testCases)
+	// Create a temporary file to store the JSON output of the rspec run.
+	// This is a temporary solution, until we have an option to specify the output file.
```
nprizal, line 93: Will implement the option to set the output, in a separate PR

internal/runner/command.go
```diff
@@ -0,0 +1,84 @@
...
+	"os/exec"
+	"os/signal"
+	"slices"
+	"syscall"
+
+	"github.com/kballard/go-shellquote"
+)
+
+// runAndForwardSignal runs the command and forwards any signals received to the command.
+func runAndForwardSignal(cmd *exec.Cmd) error {
```
nprizal, line 14: @DrJosh9000 do you think it is a sensible approach to extract this bit from `main` function into a separate function?

DrJosh9000: Yep, I think it makes sense as a function.

internal/runner/result.go
```diff
@@ -0,0 +1,14 @@
...
+
+type TestStatus string
+
+const (
+	TestStatusPassed TestStatus = "passed"
+	TestStatusFailed TestStatus = "failed"
+	TestStatusError  TestStatus = "error"
+)
+
+type TestResult struct {
```
amybiyuliu, line 11: nit: My initial impression about this struct is that it holds the result of a single test case. Perhaps we could name it differently to reflect it holds the result of executing the test command.

README.md
```diff
@@ -44,10 +44,10 @@ The following environment variables can be used optionally to configure your Tes
...
+| `BUILDKITE_SPLITTER_RETRY_COUNT` | `0` | The number of retries. Test Splitter runs the test command defined in `BUILDKITE_SPLITTER_TEST_CMD` and retries only the failed tests up to `BUILDKITE_SPLITTER_RETRY_COUNT` times, using the retry command defined in `BUILDKITE_SPLITTER_RETRY_CMD`. |
 | `BUILDKITE_SPLITTER_SPLIT_BY_EXAMPLE` | `false` | Flag to enable split by example. When this option is `true`, the Test Splitter will split the execution of slow test files over multiple partitions. |
-| `BUILDKITE_SPLITTER_TEST_CMD` | `bundle exec rspec {{testExamples}}` | Test command to run your tests. Test splitter will fill in the `{{testExamples}}` placeholder with the test splitting results |
+| `BUILDKITE_SPLITTER_TEST_CMD` | `bundle exec rspec --format progress {{testExamples}}` | Test command to run your tests. Test Splitter will fill in the `{{testExamples}}` placeholder with the test splitting results. |
```
amybiyuliu, line 50: I'm curious why we added `--format progress` to the default test command. Do we need to add `--format progress` to the test command if user is not using the default one?

niceking (approved):
This looks sound to me, I don't really understand most of the channel stuff tho so up to you if you want a better Go review before merging, or if you feel confident about the change. Happy for you to merge if you feel like its safe and thoroughly tested etc. 

Another passing thought I had was about the output of the rspec file, and how much the content varies between rspec versions? We probably don't need to investigate it too deeply right now, but at some point we'll need to figure out which versions of rspec the TSC supports

README.md
nprizal: Since we are adding `--format json --out xxx.json` to save the json file, no output will be printed if the users use the default command because the `json` format we are appending overrides Rspec default format which is `progress`. Therefore, I added `--format progress` to the default command, so by default it prints the progress to stdout and save the json to a file.

internal/runner/command.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// Create a channel that will be closed when the command finishes.
+	finishCh := make(chan struct{})
+	defer close(finishCh)
+
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+
+	// Start a goroutine to that waits for a signal or the command to finish.
```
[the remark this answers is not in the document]
nprizal: The channel stuff mostly extracted from the main, the only changes I made is to prevent signal sent by the sub process to the main process to be forwarded back to the sub process. @DrJosh9000 thinks that it is a sensible approach to extract this logic into a separate function.
@DrJosh9000 would you give another quick look, to see if I don't do anything stupid?

nprizal:
> This looks sound to me, I don't really understand most of the channel stuff tho so up to you if you want a better Go review before merging, or if you feel confident about the change. Happy for you to merge if you feel like its safe and thoroughly tested etc.

The channel stuff mostly extracted from the main, the only changes I made is to prevent signal sent by the sub process to the main process to be forwarded back to the sub process. @DrJosh9000 thinks that it is a sensible approach to extract this logic into a separate function.
@DrJosh9000 would you give another quick look, to see if I don't do anything stupid?

nprizal:
> This looks sound to me, I don't really understand most of the channel stuff tho so up to you if you want a better Go review before merging, or if you feel confident about the change. Happy for you to merge if you feel like its safe and thoroughly tested etc.
> 
> Another passing thought I had was about the output of the rspec file, and how much the content varies between rspec versions? We probably don't need to investigate it too deeply right now, but at some point we'll need to figure out which versions of rspec the TSC supports

This is a good point. I personally believe that the output will not change that much between rspec versions. If anything changes substantially, I think that might impact our test collector as well. But yes, this is something we need to keep in mind.

internal/runner/result.go
nprizal: Good point. Since this inside the `runner` package, and it refers to the status of the runner (e.g. Rspec, Jest) Maybe we can just call it `Result`? Let me know what you think, or if you have a better suggestion.

```
result, _ := testRunner.Run(...)
if result.Status == runner.StatusFailed {
  ...
}
```

wooly: `RunExecutionResult`? Since it's the result of calling `Run`?

README.md
amybiyuliu: Oh yeah I think I kinda followed.
So say if a user overwrites `BUILDKITE_SPLITTER_TEST_CMD` with something like `bin/rspec` do we need to append a `--format progress` to the `bin/rspec`? Or you already doing it 😅

nprizal: If the customer change the default to `bin/rspec` it won't have `--format progress`. I'll update the documentation to tell customer to set their own formatter when using the custom command.

wooly:
Will do a review for you shortly @nprizal :+1:

internal/runner/command_test.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+}
+
+func TestRunAndForwardSignal_CommandExitsWithNonZero(t *testing.T) {
+	cmd := exec.Command("false")
+
+	err := runAndForwardSignal(cmd)
+	exitError := new(exec.ExitError)
+	if errors.As(err, &exitError) && exitError.ExitCode() != 1 {
+		t.Errorf("Expected exit code 1, but got %d", exitError.ExitCode())
```
DrJosh9000, line 31: Would it not also be a problem if `err` wasn't an `*exec.ExitError` (or unwrap to one)?

```suggestion
	if !errors.As(err, &exitError) {
		t.Fatalf("runAndForwardSignal(cmd) error type = %T (%v), want  *exec.ExitError", err, err)
	}
	if exitError.ExitCode() != 1 {
		t.Errorf("exitError.ExitCode() = %d, want 1", exitError.ExitCode())
```

(Or similar. Note I used `t.Fatalf` in order to avoid calling `exitError.ExitCode` if `errors.As` fails, because that would probably panic.)

amybiyuliu (approved):

internal/runner/rspec.go
```diff
@@ -63,64 +67,72 @@ func (r Rspec) GetFiles() ([]string, error) {
...
+	f, err := os.CreateTemp("", "rspec-*.json")
 	if err != nil {
-		return nil, err
+		return RunResult{Status: RunStatusError}, fmt.Errorf("failed to create temporary file for rspec output: %v", err)
 	}
-	fmt.Println(shellquote.Join(append([]string{commandName}, commandArgs...)...))
+	defer f.Close()
+	defer os.Remove(f.Name())
```
DrJosh9000, line 99: Because `defer` statements are executed in reverse order, when the function returns this will remove the file and then close it. 

I don't think that causes problems on most Unix-like OSes, but something in the back of my head is telling me Windows might have problems with this order of operations...

You could either swap the order, or wrap them in the desired order within a single func:
```suggestion
	defer func() {
		f.Close()
		os.Remove(f.Name())
	}()
```

internal/runner/rspec.go
```diff
@@ -63,64 +67,72 @@ func (r Rspec) GetFiles() ([]string, error) {
...
-}
 
-// commandNameAndArgs returns the command name and arguments to run the Rspec tests
-func (r Rspec) commandNameAndArgs(testCases []string) (string, []string, error) {
-	words, err := shellquote.Split(r.TestCommand)
-	if err != nil {
-		return "", []string{}, err
+	err = runAndForwardSignal(cmd)
+
+	if err == nil {
```
DrJosh9000, line 107: Because indenting the non-error case is unusual, it might be worth an extra comment:

```suggestion
	if err == nil {  // note: returning success early
```

internal/runner/rspec.go
```diff
@@ -63,64 +67,72 @@ func (r Rspec) GetFiles() ([]string, error) {
...
+//
+// Error is returned if the command fails to run, exits prematurely, or if the
+// output cannot be parsed.
+//
+// Test failure is not considered an error, and is instead returned as a RunResult.
+func (r Rspec) Run(testCases []string, retry bool) (RunResult, error) {
+	var command string
+
+	if retry {
+		command = r.RetryTestCommand
```
DrJosh9000, line 82: (Very low-importance suggestion)

```suggestion
	command := r.TestCommand
	if retry {
		command = r.RetryTestCommand
	}
```

(no `else` needed!)

internal/runner/rspec.go
```diff
@@ -63,64 +67,72 @@ func (r Rspec) GetFiles() ([]string, error) {
...
-	fmt.Println(shellquote.Join(words...))
-
-	cmd := exec.Command(words[0], words[1:]...)
-	cmd.Stderr = os.Stderr
-	cmd.Stdout = os.Stdout
-	return cmd, nil
-}
+	commandName, commandArgs, err := commandNameAndArgs(command, testCases)
+	if err != nil {
+		return RunResult{Status: RunStatusError}, fmt.Errorf("failed to build command: %w", err)
```
DrJosh9000, line 89: Low priority comment about error strings:

`failed to` is kind of implied by it not being a `nil` error value. But it depends how the error string is used, so your mileage may vary.

wooly:
Have been testing this a bit. I was trying to ensure that subsequent retries after the first used the _new_ list of failed tests from the previous retry instead of the initial list of test failures. I'm fairly sure this is handled correctly but wanted to call it out as a thing.

LGTM!

wooly (approved):

internal/runner/command_test.go
```diff
@@ -0,0 +1,132 @@
...
+
+func TestRunAndForwardSignal_SignalReceivedInMainProcess(t *testing.T) {
+	cmd := exec.Command("sleep", "10")
+
+	// Send a SIGTERM signal to the main process.
+	go func() {
+		pid := os.Getpid()
+		process, err := os.FindProcess(pid)
+		if err != nil {
+			fmt.Println(err)
```
DrJosh9000, line 43:
```suggestion
			t.Errorf("os.FindProcess(%d) error = %v", err)
```
Four things here:

1. If `os.FindProcess` fails to find the PID of the current process, it's a bad time for the computer. So if this happens, the test is probably invalid and should be recorded as a failure.
2. Stdout (used by `fmt.Println`) is not preferable when it could be logged as part of the test output instead.
3. We can't use `t.Fatalf` because that calls `t.FailNow` which per documentation shouldn't be used outside of the main test goroutine.
4. Logging bare errors isn't good practice - aim to include some context in the log message. Since this is a test, that usually means the name and arguments of the function call returning the error.

internal/runner/command_test.go
```diff
@@ -0,0 +1,132 @@
...
+		}
+		time.Sleep(300 * time.Millisecond)
+		process.Signal(syscall.SIGTERM)
+	}()
+
+	err := runAndForwardSignal(cmd)
+
+	signalError := new(ProcessSignaledError)
+	if !errors.As(err, &signalError) {
+		t.Errorf("Expected ErrProcessSignaled, but got %v", err)
```
DrJosh9000, line 53: Is `Expected ... got...` a pattern that the rest of the project has fallen into? 
I recommend [got-before-want](https://google.github.io/styleguide/go/decisions#got-before-want), but if this is how you (as a team) prefer to do it then apply consistency and adjust my suggestions elsewhere.

DrJosh9000 (commented):
Gotta clock off for the day, but I can keep going tomorrow 😛

internal/runner/command.go
```diff
@@ -0,0 +1,84 @@
...
+		}
+		return err
+	}
+
+	return nil
+}
+
+// commandNameAndArgs replaces the "{{testExamples}}" placeholder in the test command with the test cases.
+// It returns the command name and arguments to run the tests.
+func commandNameAndArgs(cmd string, testCases []string) (string, []string, error) {
```
wooly, line 72: @nprizal this is going to be problematic for adding Jest support, since there are different interpolation variables required.

Can we move this back into the individual runners?

wooly (changes requested):

internal/runner/command_test.go
nprizal: That's a good point!

internal/runner/command_test.go
nprizal: Thanks! We use `got-before-want`. I just have this tendency to switch to `Expected ... got ...` When comparing things that are not the output of that function (e.g. error). But your suggestions in other comment makes more sense, and consistent with `got-before-want`.

wooly (approved):

PR #159, opened 2024-08-06
Downcase runner and update pact.

wooly:
### Description

Need to downcase the runner on the wire. Updates the pacts to include the runner and the branch in the payload for the create test plan endpoint.

### Testing

Have verified that the test splitter client sends the runner as downcase locally.


nprizal (approved):
Can we also patch the server to lowercase the runner, so it works with `0.8.1`?

amybiyuliu (approved):

PR #166, opened 2024-08-20
Add experimental jest support

wooly:
### Description

The test splitter client currently only supports RSpec as a runner. This PR adds Jest support, configured by setting the `BUILDKITE_SPLITTER_RUNNER` environment variable to `jest`.

**NB** This is a first cut of Jest support, so not all features are available. SBE isn't enabled for this pass.

#169 is lined up to go in first.

### Testing

- Local testing with test-splitter-integration-testing
- RC testing on bk/bk pipeline

### DIFF
--- a/internal/runner/detector.go
+++ b/internal/runner/detector.go
@@ -36,7 +36,9 @@ func DetectRunner(cfg config.Config) (TestRunner, error) {
 	switch cfg.TestRunner {
 	case "rspec":
 		return NewRspec(runnerConfig), nil
+	case "jest":
+		return NewJest(runnerConfig), nil
 	default:
-		return nil, errors.New("runner value is invalid, possible values are 'rspec'")
+		return nil, errors.New("runner value is invalid, possible values are 'rspec', 'jest'")
 	}
 }
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -1,0 +1,196 @@
+package runner
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"os/exec"
+	"slices"
+	"strings"
+
+	"github.com/buildkite/test-splitter/internal/debug"
+	"github.com/buildkite/test-splitter/internal/plan"
+	"github.com/kballard/go-shellquote"
+)
+
+var _ = TestRunner(Jest{})
+
+type Jest struct {
+	RunnerConfig
+}
+
+func NewJest(j RunnerConfig) Jest {
+	if j.TestCommand == "" {
+		j.TestCommand = "yarn test {{testExamples}} --json --testLocationInResults --outputFile {{outputFile}}"
+	}
+
+	if j.TestFilePattern == "" {
+		j.TestFilePattern = "**/{__tests__/**/*,*.spec,*.test}.{ts,js,tsx,jsx}"
+	}
+
+	if j.TestFileExcludePattern == "" {
+		j.TestFileExcludePattern = "node_modules"
+	}
+
+	if j.RetryTestCommand == "" {
+		j.RetryTestCommand = "yarn test --testNamePattern '{{testNamePattern}}' --json --testLocationInResults --outputFile {{outputFile}}"
+	}
+
+	return Jest{j}
+}
+
+func (j Jest) Name() string {
+	return "Jest"
+}
+
+// GetFiles returns an array of file names using the discovery pattern.
+func (j Jest) GetFiles() ([]string, error) {
+	debug.Println("Discovering test files with include pattern:", j.TestFilePattern, "exclude pattern:", j.TestFileExcludePattern)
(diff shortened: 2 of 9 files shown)


internal/runner/jest.go
```diff
@@ -0,0 +1,184 @@
...
+
+func (j Jest) retryCommandNameAndArgs(cmd string, testCases []string) (string, []string, error) {
+	words, err := shellquote.Split(cmd)
+	if err != nil {
+		return "", []string{}, err
+	}
+
+	idx := slices.Index(words, "{{testNamePattern}}")
+	if idx < 0 {
+		err := fmt.Errorf("Couldn't find '{{testNamePattern}}' sentinel in retry command, exiting.")
```
nprizal, line 172:
```suggestion
		err := fmt.Errorf("couldn't find '{{testNamePattern}}' sentinel in retry command")
```

[Error string styleguide](https://google.github.io/styleguide/go/decisions.html#error-strings)

In convention, error string in Golang should not be capitalized. Also, I think `exiting` message should be logged from the consumer (main fn), as it is the responsibility of consumer to exit the program.

internal/runner/detector.go
```diff
@@ -4,17 +4,38 @@ import (
...
-			TestCommand:            cfg.TestCommand,
-			TestFilePattern:        cfg.TestFilePattern,
-			TestFileExcludePattern: cfg.TestFileExcludePattern,
-			RetryTestCommand:       cfg.RetryCommand,
-		}), nil
+		return NewRspec(runnerConfig), nil
+	case "jest":
+		return NewJest(runnerConfig), nil
 	default:
 		return nil, errors.New("Runner value is invalid, possible values are 'rspec'")
```
nprizal, line 40:
```suggestion
		return nil, errors.New("runner value is invalid, possible values are 'rspec', 'jest'")
```

internal/runner/rspec.go
```diff
@@ -14,18 +14,13 @@ import (
 	"github.com/kballard/go-shellquote"
 )
 
-// In future, Rspec will implement an interface that defines
-// behaviour common to all test runners.
-// For now, Rspec provides rspec specific behaviour to execute
-// and report on tests in the Rspec framework.
+var _ = TestRunner(Rspec{})
```
nprizal, line 17: Can you explain the use of this? I've never seen this before

internal/runner/jest.go
```diff
@@ -0,0 +1,184 @@
...
+	return RunResult{Status: RunStatusError}, err
+}
+
+type JestExample struct {
+	Name   string `json:"fullName"`
+	Status string `json:"status"`
+}
+
+type JestReport struct {
+	NumFailedTests int `json:"numFailedTests"`
```
nprizal, line 130:
```suggestion
	NumFailedTests int
```
[nit] I think we can omit this because the json attribute name is the same with the struct field name. Or if we want to keep this, maybe add the json field name for `TestResults` and `AssertionResults` to be consistent?

internal/runner/rspec.go
wooly: This is a build-time check rather than a run-time check to ensure that the Rspec struct implements the TestRunner interface :+1:

niceking: oic TIL https://medium.com/@matryer/golang-tip-compile-time-checks-to-ensure-your-type-satisfies-an-interface-c167afed3aae

internal/runner/jest.go
```diff
@@ -0,0 +1,196 @@
...
+		if err != nil {
+			return RunResult{Status: RunStatusError}, fmt.Errorf("failed to build command: %w", err)
+		}
+
+		cmd = exec.Command(commandName, commandArgs...)
+	}
+
+	err = runAndForwardSignal(cmd)
+
+	if err == nil {
```
nprizal, line 88: returning early when error is nil is unusual in Go, maybe add comment. https://github.com/buildkite/test-splitter/pull/156#discussion_r1721323894
```suggestion
	if err == nil { // note: returning success early
```

internal/runner/jest.go
```diff
@@ -0,0 +1,196 @@
...
+		err := fmt.Errorf("couldn't find '{{outputFile}}' sentinel in retry command, exiting.")
+		return "", []string{}, err
+	}
+	slices.Replace(words, outputIdx, outputIdx+1, j.ResultPath)
+
+	return words[0], words[1:], err
+}
+
+func (j Jest) GetExamples(files []string) ([]plan.TestCase, error) {
+	return nil, nil
```
nprizal, line 195: Maybe return an error for now?
```suggestion
	return nil, fmt.Error("not supported in Jest")
```

wooly:
Have re-requested a review from @niceking and @nprizal and have tagged a couple of go-shaped people for a quick sanity check on whether or not I'm committing some Go crimes.

nprizal (approved):

PR #167, opened 2024-08-20
Add option to set output file

nprizal:
### Description
in #156, we changed how we trigger the retry by saving the output of Rspec run to detect failed tests. The output of Rspec run was saved in a temporary file that will be deleted at the end of operation.

This PR adds a new option called `BUILDKITE_SPLITTER_RESULT_PATH` that can be used to specify the path to save the output of Rspec run. If the configured path has `*`, it will be replaced with a UNIX timestamp, this is useful when the retry function is enabled, and we want to store the output for each run separately without overriding the previous run.

### Testing
I've done integration testing manually
- set `BUILDKITE_SPLITTER_RESULT_PATH=./tmp/my-rspec-result-*.json`
- run `./test-splitter`
- check if a json file exists in `$BUILDKITE_SPLITTER_RESULT_PATH` configured above


internal/runner/rspec.go
```diff
@@ -89,17 +92,28 @@ func (r Rspec) Run(testCases []string, retry bool) (RunResult, error) {
...
-	f, err := os.CreateTemp("", "rspec-*.json")
-	if err != nil {
-		return RunResult{Status: RunStatusError}, fmt.Errorf("failed to create temporary file for rspec output: %v", err)
+	var resultPath string
+
+	if r.ResultPath != "" {
+		resultPath = strings.ReplaceAll(r.ResultPath, "*", strconv.FormatInt(time.Now().Unix(), 10))
+	} else {
+		// Create a temporary file to store the JSON output of the rspec run.
+		// This is a temporary solution, until we have an option to specify the output file.
```
niceking, line 101: Is this comment out of date?

internal/runner/rspec.go
```diff
@@ -89,17 +92,28 @@ func (r Rspec) Run(testCases []string, retry bool) (RunResult, error) {
...
 	}
 
-	// Create a temporary file to store the JSON output of the rspec run.
-	// This is a temporary solution, until we have an option to specify the output file.
-	f, err := os.CreateTemp("", "rspec-*.json")
-	if err != nil {
-		return RunResult{Status: RunStatusError}, fmt.Errorf("failed to create temporary file for rspec output: %v", err)
+	var resultPath string
+
+	if r.ResultPath != "" {
```
niceking, line 97: Should this whole thing be wrapped in an `if retry`? Like we only have to write the file if the customer wants retries, right?

niceking (commented):
> If the configured path has *, it will be replaced with a UNIX timestamp, this is useful when the retry function is enabled, and we want to store the output for each run separately without overriding the previous run.

What happens if the configured path doesn't have * and retries are enabled? Does that mean it won't work properly?

internal/runner/rspec.go
nprizal: Good point. Currently, TSC will write the output even when retry is disabled, as I think it will be useful to save the output regardless. Do you think we should save the output only when the retry is enabled for now?

nprizal:
> What happens if the configured path doesn't have * and retries are enabled? Does that mean it won't work properly?

It still works properly, because we interpret the report before we execute the retry runs. It just means that the result from retries will override the result from previous run(s).

nprizal:
> What happens if the configured path doesn't have * and retries are enabled? Does that mean it won't work properly?

@niceking  It still works properly, because we interpret the report before we execute the retry runs. It just means that the result from retries will override the result from previous run(s).

internal/runner/rspec.go
```diff
@@ -89,17 +92,27 @@ func (r Rspec) Run(testCases []string, retry bool) (RunResult, error) {
...
+		resultPath = strings.ReplaceAll(r.ResultPath, "*", strconv.FormatInt(time.Now().Unix(), 10))
+	} else {
+		// Create a temporary file if no result path is provided.
+		f, err := os.CreateTemp("", "rspec-*.json")
```
amybiyuliu, line 101: I was thinking about "rspec-*.json" could be the default of `BUILDKITE_SPLITTER_RESULT_PATH` and we don't need the conditional check on 97 🤔 the problem is we won't be able to tell if we want to remove the result file at the end. 
Is there value to always keep the result file? or that's more problematic?

wooly: There is definitely value in keeping the results files around as part of the test splitting runner process. Currently, the `bk/bk` run group outputs N+1 versions of the results JSON files:

One for each run of the tests:

Screenshot 2024-08-22 at 9 32 46 AM

And one file used internally by the test splitter to identify which tests need to be retried. 

If we merged this PR, we would have support for persisting the rspec run results _within_ the test splitter, so we could remove the explicit behaviour to store the rspec output results. I think this is currently handled in .rspec.ci.

Anyways TLDR is a little bit simpler and probably valuable for other customers too, so we might as well do it!

internal/runner/rspec.go
niceking: I think it would possibly be a bit surprising if we created a file and didn't tell the user about it. I think I'd maybe design it like if the customer specifies a file name, then the file is saved, but if they don't, the file is deleted? I wouldn't want a side effect of using the TSC be that we clutter a host machine with lots of files that the user didn't know we were creating? What do you think?

wooly: I think we should save the output if the user manually specifies the file path for the output rather than if retry is enabled.

internal/runner/rspec.go
nprizal: > Is there value to always keep the result file? or that's more problematic?

I think this is related to @niceking comment

> I think it would possibly be a bit surprising if we created a file and didn't tell the user about it. I think I'd maybe design it like if the customer specifies a file name, then the file is saved, but if they don't, the file is deleted? I wouldn't want a side effect of using the TSC be that we clutter a host machine with lots of files that the user didn't know we were creating? What do you think?


So I think, we should persist the file only when customer specify it?

amybiyuliu: Yep, the comment Rose made actually answered my question. I agree we persist the file only when customer specify it.

internal/runner/rspec.go
niceking: Yup that sounds good!

niceking (approved):

internal/runner/rspec.go
nprizal: Yeah, that's a good point, I considered this option when I implemented these changes. Currently, it's redundant when retry is disabled and no file path is specified. However, this will be useful in the future when we want to do something with the output, for example to send the output to Test Analytics. Although it's more like a future problem, I'd like to keep it this way for the simplicity.

> Unless there's some unintended side effects of creating and then deleting this temp file? Like permissions or in the unlikely event that file creation/deletion crashes?

There is always a risk, but I'd say the risk is low. We are using go std library to create/delete the temp file, which I believe is carefully implemented and tested.

niceking: Awesome to hear you've though through all the trade-offs and future possibilities 👍

PR #169, opened 2024-08-26
Prepare for Jest support.

wooly:
### Description

There are a few things we can do in preparation for adding support for Jest.

- Tidy up folder structure to namespace fixtures into the rspec runner
- Extract RunnerConfig struct for generalising the configuration
- A few typos and things

### Rollback

Revert PR.


internal/runner/rspec.go
```diff
@@ -16,19 +16,13 @@ import (
...
-	TestCommand            string
-	TestFileExcludePattern string
-	TestFilePattern        string
-	RetryTestCommand       string
-	ResultPath             string
+	RunnerConfig
 }
 
-func NewRspec(r Rspec) *Rspec {
+func NewRspec(r RunnerConfig) Rspec {
```
niceking, line 25: I don't know much Go, but how come this has changed to return a struct rather than a pointer?

niceking (commented):
What's the relationship between this PR and PR  #166? Are you planning on merging this one first and then removing the duplication from PR #166?

wooly:
> What's the relationship between this PR and PR #166? Are you planning on merging this one first and then removing the duplication from PR #166?

Yeah I'll merge this one, rebase the other one onto main, then we can review that one separately.

internal/runner/rspec.go
wooly: I had a wee pairing session with @wolfeidau on this last week. My recollection (forgive me if this is wrong Mark!) is that since we're not mutating the returned value in any way and there is no shared state between Rspec (or jest) instances - since there is only one - it's neater to just return the value rather than the pointer so we can avoid de-referencing it everywhere we need to use it.

wolfeidau: @wooly correct, this is more of a "configuration" or immutable thing, not a stateful thing. 

As mentioned this can trip people up so most certainly something to keep an eye on.

niceking: Ah, okay cool!

niceking (approved):
seems good but I am not a Go expert

wooly:
I'll wait until @nprizal gets a look in since he wrote the original Rspec code.

internal/runner/rspec.go
```diff
@@ -16,19 +16,13 @@ import (
 	"github.com/kballard/go-shellquote"
 )
 
-// In future, Rspec will implement an interface that defines
-// behaviour common to all test runners.
-// For now, Rspec provides rspec specific behaviour to execute
-// and report on tests in the Rspec framework.
+var _ = TestRunner(Rspec{})
```
nprizal, line 19: I asked this on the other PR, what's the purpose of this?

main.go
```diff
@@ -55,7 +55,7 @@ func main() {
 
 	files, err := testRunner.GetFiles()
 	if err != nil {
-		logErrorAndExit(16, "Couldn't get files: %v", err)
+		logErrorAndExit(16, "couldn't get files: %v", err)
```
nprizal, line 58: I think this is fine with a Capital letter. The conventions of not using capital letter is for error _message_, rather than the _log_.

nprizal (approved):
Looks good to me. I have few non blocking comments

internal/runner/rspec.go
nprizal: oops, I must've missed your reply on the other PR. All good!

PR #170, opened 2024-08-26
Fix up printed configuration errors.

wooly:
### Description

The test-splitter client will perform validation on a selection of the environment variables to ensure that they are in the correct format, have valid values, or are within certain bounds. If the client encounters validation errors, an error message is printed and the client will exit with status code 16.

The issue with these errors is that they refer to the members of the struct config instead of the environment variables:

8918be43-bb34-44d2-932f-c35733fe9734

It would be a lot easier for the user if these said BUILDKITE_SPLITTER_SUITE_SLUG must not be blank etc.

This PR changes the format for these error messages to look like this:

```
❯ ./run-splitter
Buildkite Test Splitter: Invalid configuration
BUILDKITE_PARALLEL_JOB was "", must be a number
BUILDKITE_PARALLEL_JOB was 0, must not be greater than -1
BUILDKITE_PARALLEL_JOB_COUNT was 0, must be greater than or equal to 1
BUILDKITE_SPLITTER_API_ACCESS_TOKEN must not be blank
BUILDKITE_SPLITTER_SUITE_SLUG must not be blank
```

### DIFF
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -9,49 +9,41 @@ func (c *Config) validate() error {
 	var errs InvalidConfigError
 
 	if c.MaxRetries < 0 {
-		errs.appendFieldError("MaxRetries", "was %d, must be greater than or equal to 0", c.MaxRetries)
-	}
-
-	if c.Identifier == "" {
-		errs.appendFieldError("Identifier", "must not be blank")
-	}
-
-	if got, limit := len(c.Identifier), 1024; got > limit {
-		errs.appendFieldError("Identifier", "was %d bytes long, must not be longer than %d", got, limit)
+		errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
 	}
 
 	if got, min := c.Parallelism, 1; got < min {
-		errs.appendFieldError("Parallelism", "was %d, must be greater than or equal to %d", got, min)
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %d, must be greater than or equal to %d", got, min)
 	}
 
 	if got, max := c.Parallelism, 1000; got > max {
-		errs.appendFieldError("Parallelism", "was %d, must not be greater than %d", got, max)
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %d, must not be greater than %d", got, max)
 	}
 
 	if got, min := c.NodeIndex, 0; got < 0 {
-		errs.appendFieldError("NodeIndex", "was %d, must be greater than or equal to %d", got, min)
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %d, must be greater than or equal to %d", got, min)
 	}
 
 	if got, max := c.NodeIndex, c.Parallelism-1; got > max {
-		errs.appendFieldError("NodeIndex", "was %d, must not be greater than %d", got, max)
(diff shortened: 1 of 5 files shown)


internal/config/validate.go
```diff
@@ -9,49 +9,41 @@ func (c *Config) validate() error {
...
 	}
 
 	if got, min := c.NodeIndex, 0; got < 0 {
-		errs.appendFieldError("NodeIndex", "was %d, must be greater than or equal to %d", got, min)
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %d, must be greater than or equal to %d", got, min)
 	}
 
 	if got, max := c.NodeIndex, c.Parallelism-1; got > max {
-		errs.appendFieldError("NodeIndex", "was %d, must not be greater than %d", got, max)
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %d, must not be greater than %d", got, max)
```
niceking, line 28: In your PR description the error reads like this: `BUILDKITE_PARALLEL_JOB was 0, must not be greater than -1`

I don't think this is right?

wooly: Yeah definitely not right! Will take a look, thanks for noticing!

wooly:
I've had a look at this and can't get it to do what I want because my Go isn't good enough.

What I'm after is to print out only a single error message for each field. So this:
```
BUILDKITE_PARALLEL_JOB_COUNT was 0, must be greater than or equal to 1
```

Where the PARALLEL_JOB_COUNT isn't present in the environment would return an error like
```
BUILDKITE_PARALLEL_JOB_COUNT not set in environment
```

And then when it is set to a numeric value that's not valid, it would say:
```
BUILDKITE_PARALLEL_JOB_COUNT must be between 1 and 500
```

nprizal:
> I've had a look at this and can't get it to do what I want because my Go isn't good enough.
> 
> What I'm after is to print out only a single error message for each field. So this:
> 
> ```
> BUILDKITE_PARALLEL_JOB_COUNT was "", must be a number
> BUILDKITE_PARALLEL_JOB_COUNT was 0, must be greater than or equal to 1
> ```
> 
> Where the PARALLEL_JOB_COUNT isn't present in the environment would return an error like
> 
> ```
> BUILDKITE_PARALLEL_JOB_COUNT not set in environment
> ```
> 
> And then when it is set to a numeric value that's not valid, it would say:
> 
> ```
> BUILDKITE_PARALLEL_JOB_COUNT must be between 1 and 500
> ```

I think this is the problem with reading the config, which isn't the purpose of this PR. I'd say we just fix the config name in this PR and fix the reading in separate PR?

internal/config/validate.go
wooly: @niceking this is happening because the PARALLEL_JOB_COUNT is set to 0, and it's being dynamically calculated. Are you happy for me to ship this as is and raise another card to improve the configuration reading/validation as suggested by @nprizal below?

wooly: @niceking This ^^

niceking: Hmmm just had another read through of this and it seems like the actual validation check is wrong, but the string is right? Like the validation check is: 

```
if got, max := c.NodeIndex, c.Parallelism-1; got > max {
		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %d, must not be greater than %d", got, max)
```

So it's a little confusing that the max is `-1`? 

Anyways I think this needs to go in for v0.9.0, I don't really care if it goes in in this PR or another one, but this should be fixed. Will approve this and you can decide whether you wanna fix the validation in this PR or the next one

niceking (approved):
Up to you whether you want to fix the config in this PR or a separate one, it doesn't really matter to me either way but it does need to be done for v0.9.0

PR #172, opened 2024-08-27
Stop appending `--format json --out ...` to Rspec command

nprizal:
### Description

In [v0.9.0-rc.1](https://github.com/buildkite/test-splitter/releases/tag/v0.9.0-rc.1), we have added a feature that adds `--format json --out $BUILDKITE_SPLITTER_RESULT_PATH` to the Rspec command, so it can read the Rspec result. This caused an issue where the `--format` sets in Rspec config file being overridden, because `--format` that the Test Splitter appended to the command has higher specificity than the config file (see Rspec [doc](https://rspec.info/features/3-12/rspec-core/configuration/read-options-from-file/)).

This PR moves `--format json` the default Rspec command, and ask customer to explicitly set the formatter themselves if they want to customize the command. The customer also need to set `BUILDKITE_SPLITTER_RESULT_PATH` to tell Test Splitter where to look at the Rspec result.


README.md
```diff
@@ -37,18 +37,19 @@ In addition to the above variables, you must set the following environment varia
 | `BUILDKITE_SPLITTER_API_ACCESS_TOKEN ` | Buildkite API access token with `read_suites`, `read_test_plan`, and `write_test_plan` scopes. You can create an access token from [Personal Settings](https://buildkite.com/user/api-access-tokens) in Buildkite |
 | `BUILDKITE_SPLITTER_SUITE_SLUG` | The slug of your Buildkite Test Analytics test suite. You can find the suite slug in the url for your suite. For example, the slug for the url: https://buildkite.com/organizations/my-organization/analytics/suites/my-suite is `my-suite` |
 
+| `BUILDKITE_SPLITTER_RESULT_PATH` | Test Splitter needs to collect the result generated by the test runner. It will use this path to locate the file. Additionally, Test Splitter will replace the `{{resultPath}}` placeholder in the `BUILDKITE_SPLITTER_TEST_CMD` with this value. This ensures that the test runner saves the output to the specified path. |
```
wooly, line 40: How about just giving some examples instead, something like...

```suggestion
| `BUILDKITE_SPLITTER_RESULT_PATH` | Test Splitter uses the result path environment variable to tell the runner where to store test result summaries, for use in retries and verification. For RSpec, this would usually be set to `tmp/rspec.json`, while for Jest, this would usually be set to `tmp/jest.json`. |
```

wooly (approved):
Just one nit around language, but looks good.

PR #173, opened 2024-08-27
Rename {{outputFile}} to {{resultPath}}

wooly:
### Description

To simplify things a little bit, we should rename the outputFile interpolation sentinel to resultPath, to match the environment variable name more closely.


nprizal (approved):
Nice one

PR #174, opened 2024-08-28
Append --format progress to Rspec dry run every time

nprizal:
We append `--format progress` to Rspec dry run when getting examples only when debug mode is enabled. To provide more information to the customer when there is a failure during the dry run, we need to append `--format progress` to the command regardless of the debug mode status.

### DIFF
--- a/internal/runner/rspec.go
+++ b/internal/runner/rspec.go
@@ -202,10 +202,9 @@ func (r Rspec) GetExamples(files []string) ([]plan.TestCase, error) {
 
 	cmdArgs = append(cmdArgs, "--dry-run", "--format", "json", "--out", f.Name(), "--format", "progress")
 
-	debug.Println("Running `rspec --dry-run`")
+	debug.Printf("Running `%s %s` for dry run", cmdName, strings.Join(cmdArgs, " "))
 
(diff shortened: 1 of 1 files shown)


wooly (approved):

internal/runner/rspec.go
```diff
@@ -215,10 +215,7 @@ func (r Rspec) GetExamples(files []string) ([]plan.TestCase, error) {
 		return nil, err
 	}
 
-	cmdArgs = append(cmdArgs, "--dry-run", "--format", "json", "--out", f.Name())
-	if debug.Enabled {
-		cmdArgs = append(cmdArgs, "--format", "progress")
-	}
+	cmdArgs = append(cmdArgs, "--dry-run", "--format", "json", "--out", f.Name(), "--format", "progress")
 
 	debug.Println("Running `rspec --dry-run`")
```
wooly, line 220: There is maybe a logic error here now, but will need to confirm.

It feels like with debug mode on, the logs will print twice?

amybiyuliu (approved):
🚀

internal/runner/rspec.go
nprizal: Since we are now appending `--format progress` every time, and log the output when the dry run fails, I don't think we should print the output in debug mode. I'll remove it.

PR #175, opened 2024-08-28
Test splitter exits early if either STEP_ID or BUILD_ID are blank.

wooly:
### Description

Since 0.7.0, the test splitter has constructed the identifier internally. However, the validation was never changed to ensure that _both_ the BUILD_ID and STEP_ID are present in the environment to generate the identifier correctly.

This PR updates the test splitter so that it will exit with a configuration error (status 16) if either the BUILDKITE_BUILD_ID or BUILDKITE_STEP_ID environment variables are blank.

### DIFF
--- a/internal/config/read_test.go
+++ b/internal/config/read_test.go
@@ -98,3 +98,45 @@ func TestConfigReadFromEnv_NotInteger(t *testing.T) {
 		t.Errorf("config.readFromEnv() error length = %d, want 2", len(invConfigError))
 	}
 }
+
+func TestConfigReadFromEnv_MissingBuildId(t *testing.T) {
+	os.Setenv("BUILDKITE_SPLITTER_BASE_URL", "")
+	os.Setenv("BUILDKITE_SPLITTER_MODE", "")
+	os.Setenv("BUILDKITE_SPLITTER_TEST_CMD", "")
+	os.Setenv("BUILDKITE_SPLITTER_RETRY_COUNT", "")
+	os.Setenv("BUILDKITE_STEP_ID", "123")
+	defer os.Clearenv()
+
+	c := Config{}
+	err := c.readFromEnv()
+
+	var invConfigError InvalidConfigError
+	want := "BUILDKITE_BUILD_ID must not be blank"
+
+	if errors.As(err, &invConfigError) {
+		if got := invConfigError[0].Error(); got != want {
+			t.Errorf("config.readFromEnv() got = %v, want = %v", got, want)
+		}
+	}
+}
+
+func TestConfigReadFromEnv_MissingStepId(t *testing.T) {
+	os.Setenv("BUILDKITE_SPLITTER_BASE_URL", "")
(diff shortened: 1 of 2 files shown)


internal/config/read_test.go
```diff
@@ -95,6 +99,49 @@ func TestConfigReadFromEnv_NotInteger(t *testing.T) {
...
+	os.Setenv("BUILDKITE_STEP_ID", "123")
+	defer os.Clearenv()
+
+	c := Config{}
+	err := c.readFromEnv()
+
+	var invConfigError InvalidConfigError
+	want := "BUILDKITE_BUILD_ID must not be blank"
+
+	if errors.As(err, &invConfigError) {
```
nprizal, line 121: If the error is not an `InvalidConfigError`, this test will pass

wooly: How's that?

nprizal: Because if the `err` is not an `InvalidConfigError`, the `t.Errorf` within will not be called, and the test function will finish without throwing any failure.

Now that you've added the check above, you can remove the `if` condition here.

```
var invConfigError InvalidConfigError
if !errors.As(err, &invConfigError) {
	t.Fatalf("config.readFromEnv() error = %v, want InvalidConfigError", err)
}

if got := invConfigError[0].Error(); got != want {
	t.Errorf("config.readFromEnv() got = %v, want = %v", got, want)
}
```

Notice that I use `t.Fatalf` on the first check to prevent test function to execute the next check.
You can refer to this [test](https://github.com/buildkite/test-splitter/blob/6b67efc34acfbca72bbfe69c904aabae4152f6e6/internal/runner/command_test.go#L26) implementation.

wooly: Oh sorry, by how's that, I meant how does this commit look: https://github.com/buildkite/test-splitter/pull/175/commits/b468111e35f766011f598b28c87207519e4359d5

nprizal: ah! 😄

internal/config/read_test.go
```diff
@@ -95,6 +99,57 @@ func TestConfigReadFromEnv_NotInteger(t *testing.T) {
...
+		t.Errorf("config.readFromEnv() error = %v, want InvalidConfigError", err)
+	}
+
+	want := "BUILDKITE_BUILD_ID must not be blank"
+
+	if errors.As(err, &invConfigError) {
+		if got := invConfigError[0].Error(); got != want {
+			t.Errorf("config.readFromEnv() got = %v, want = %v", got, want)
+		}
+	}
```
nprizal, line 129: [nit] we have the error struct conversion and check above, no need to check it again here.
```suggestion
	if got := invConfigError[0].Error(); got != want {
		t.Errorf("config.readFromEnv() got = %v, want = %v", got, want)
	}
```

nprizal (approved):

internal/config/read_test.go
wooly: Ah yep, good catch, thanks!

PR #176, opened 2024-09-02
Update Split By Example implementation

nprizal:
### Description

Previously, Split By Example is implemented against files that take slower than 3 minutes. To extend the functionality of Split By Example for better slow file detection, as well as to support test quarantining in the future, we want to have the server determine which files need to be split by example.

**Before**
1. Client sends request `POST /test_files` endpoint to get duration for all files
2. Client finds files slower than 3 minutes
3. Client executes dry run for those files
4. Client sends examples within those files along with other files to create test plan

**After**
1. Client sends request `POST /test_plan/filter_tests` endpoint to filter files that will be Split by Example
2. Client executes dry run for those files
3. Client sends examples within those files along with other files to create test plan

### Testing

Do integration test with `BUILDKITE_SPLITTER_SPLIT_BY_EXAMPLE=true` against a test suite that has files longer than 3 minutes to run. The files that takes longer than 3 minutes should be split by example.


main.go
```diff
@@ -257,102 +257,91 @@ func fetchOrCreateTestPlan(ctx context.Context, apiClient *api.Client, cfg confi
...
+	testFiles := []plan.TestCase{}
+	for _, file := range files {
+		testFiles = append(testFiles, plan.TestCase{
+			Path: file,
+		})
+	}
+
 	if !cfg.SplitByExample {
 		debug.Println("Splitting by file")
-		testCases := []plan.TestCase{}
```
nprizal, line 274: Moved these lines of code above, so it can be reused when filtering the tests

internal/api/filter_tests.go
```diff
@@ -0,0 +1,42 @@
...
+	"fmt"
+	"net/http"
+
+	"github.com/buildkite/test-splitter/internal/plan"
+)
+
+type FilterTestsParams struct {
+	Files          []plan.TestCase `json:"files"`
+	Parallelism    int             `json:"parallelism"`
+	SplitByExample bool            `json:"split_by_example"`
```
nprizal, line 14:
```suggestion
	Parallelism    int             `json:"parallelism"`
	SplitByExample bool            `json:"split_by_example"`
```
The server won't do anything with this information yet, but these are the least information we need to support dynamic slow file threshold, and quarantine.

wooly: I feel like we had landed on sending the sanitized list of environment variables as part of this payload instead of defining a custom struct.

nprizal: What's the reasoning behind sending the environment variables? Is that so we don't need to update the client if we need more information, or to avoid translating env and struct?

wooly: It's the former yeah.

internal/api/filter_tests.go
```diff
@@ -0,0 +1,41 @@
...
+type FilteredTest struct {
+	Path   string `json:"path"`
+	Reason string `json:"reason"`
+}
+
+type filteredTestResponse struct {
+	Tests []FilteredTest `json:"tests"`
+}
+
+// FilterTests filters tests from the server.
```
wooly, line 25: [nit] Is this comment slightly redundant?

wooly (approved):
Looks good!

PR #178, opened 2024-09-05
Rename TSC to bktec

niceking:
### Description

Have renamed the following: 
- tsc -> bktec
- analytics -> engine
- test splitter API/server -> test plan API/server
- `BUILDKITE_SPLITTER_*` -> `BUILDKITE_TEST_ENGINE_*`

Outstanding tasks that will be done in a separate PR:
- [ ] update README (we should update this before release)
- [ ] update `packaging/Dockerfile`
- [ ] update `.goreleaser.yaml`
- [ ] update `pipeline.yml` and `pipeline.release.yml`

### Testing

This is purely a rename, and shouldn't change any functionality of bktec. It also doesn't change any of the packaging and releasing. 

Will test locally that a build of this still fetches a test plan, given the new env var names


niceking:
I'm a little confused by how to do the go module rename, and the sequencing of all the rename and packaging and releasing actions. 

I found [this guide](https://rodaine.com/til/2023/05/go-module-rename-v2/) and based on that, I think what we want to do is: 
1. * do all code changes in the `bktec` branch 
2. * merge changes to main and tag with v1.0.0 and then do the GitHub Repository rename, which should then update the go mod refs? And it should also preserve history - so I assume previous releases will be available under old namespace 
3. * update all build & release processes to reference the new `bktec` name, I think that's the goreleaser and the release pipeline 

I'm not 100% sure at which point we update the imports at the top of each file? Is that part of step 1? 

Does this sound sane? cc @DrJosh9000 @nprizal / anyone

DrJosh9000:
> I'm a little confused by how to do the go module rename, and the sequencing of all the rename and packaging and releasing actions.
> 
> I found [this guide](https://rodaine.com/til/2023/05/go-module-rename-v2/) and based on that, I think what we want to do is:
> 
> 1. do all code changes in the `bktec` branch
> 2. merge changes to main and tag with v1.0.0 and then do the GitHub Repository rename, which should then update the go mod refs? And it should also preserve history - so I assume previous releases will be available under old namespace
> 3. update all build & release processes to reference the new `bktec` name, I think that's the goreleaser and the release pipeline
> 
> (this plan is a little different from what the guide recommends, cos they keep a v1 branch around when they release a v2 with a new name, but we've decided that we're releasing v1.0.0 with the new name. If previous versions have issues we will ask the customer to upgrade and not patch them)
> 
> I'm not 100% sure at which point we update the imports at the top of each file? Is that part of step 1?
> 
> Does this sound sane? cc @DrJosh9000 @nprizal / anyone

Because the whole thing is a binary, nobody (Buildkite or otherwise) is importing packages from it - so a lot of the worries to do with locating Go modules across versions and ensuring they are published in a findable place aren't really a concern. I'm pretty sure you won't be breaking anybody's build (other than your own). So you could do this:

1. Rename the repo (breaking the build for a while)
2. Fix all the names
3. When happy, tag v1.0.0

nprizal:
> Changing the binary build name - does this just get done as part of the go mod rename?

We use [goreleaser](https://goreleaser.com/customization/builds/) to build the binary, and by default it uses the project directory name. So, until we rename the git repository, we may need to explicitly set the binary name in the goreleaser config file.

> Are there Pact changes needed as part of this?

We changed the consumer and provider name in this PR,  therefore we need to change the Pact configuration (state) in the Buildkite repository.

> aws ssm param namespace maybe should be changed? not super necessary tho

Yes, I've captured this as Linear tickets.

internal/api/post_test_plan_metadata.go
```diff
@@ -12,9 +12,9 @@ type Timeline struct {
 }
 
 type TestPlanMetadataParams struct {
-	Version     string            `json:"version"`
-	SplitterEnv map[string]string `json:"splitter_env"`
-	Timeline    []Timeline        `json:"timeline"`
+	Version  string            `json:"version"`
+	Env      map[string]string `json:"env"`
```
nprizal, line 16: I think we can just call it `env` instead of `bktec_env`?

niceking: yeah i had it as `bktec_env` as the prev name was `splitter_env`.  Happy for just `env`

niceking:
I can't approve this as I'm the original author, but LGTM once the tests are passing

wooly (approved):
LGTM!

PR #180, opened 2024-09-09
Update release process for bktec

amybiyuliu:
### Description

Update Goreleaser: 
- The binary name is called `bktec`
- The binary will be published to new dockerhub and package cloud repositories (with `test-engine-client` in the url) 
- The binary will be published to homebrew as `bktec`

### Testing

Verified the change locally by runing `goreleaser release --clean --skip publish --snapshot`. The binaries are created with name `bktec`

After merging this PR we can release v1.0.0-rc.1 to verify the actual release works

### DIFF
--- a/.buildkite/steps/upload-linux-packages.sh
+++ b/.buildkite/steps/upload-linux-packages.sh
@@ -2,7 +2,7 @@
 
 file=${1}
 extension="${file##*.}"
-registry="test-splitter-${extension}"
+registry="test-engine-client-${extension}"
 audience="https://packages.buildkite.com/buildkite/${registry}"
 
 echo "--- :key: :buildkite: Fetching OIDC token for ${audience}"
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -19,6 +19,7 @@ builds:
     goos: [linux, darwin]
     goarch: [amd64, arm64]
     ldflags: "-X 'main.Version=v{{ .Version }}'"
+    binary: bktec
 
 brews:
   - name: test-splitter
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -2,7 +2,7 @@ version: 2
 project_name: test-engine-client
 
 release:
-  name_template: Buildkite Test Engine Client v{{.Version}}
+  name_template: Test Engine Client v{{.Version}}
   draft: false
   prerelease: auto
   make_latest: false
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,4 +1,5 @@
 version: 2
+project_name: test-engine-client
 
 release:
   name_template: Buildkite Test Engine Client v{{.Version}}
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -19,15 +19,16 @@ builds:
     goos: [linux, darwin]
     goarch: [amd64, arm64]
     ldflags: "-X 'main.Version=v{{ .Version }}'"
+    binary: bktec
 
 brews:
-  - name: test-engine-client
+  - name: bktec
     description: "Buildkite Test Engine Client"
     homepage: "https://github.com/buildkite/test-engine-client"
     skip_upload: auto
     directory: .
     test: |
-      version_output = shell_output("test-engine-client --version")
+      version_output = shell_output("bktec --version")
       assert_match "v#{version}\n", version_output
     repository:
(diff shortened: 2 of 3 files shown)


.goreleaser.yaml
```diff
@@ -22,8 +22,8 @@ builds:
 
 brews:
   - name: test-splitter
-    description: "Buildkite test splitting client"
-    homepage: "https://github.com/buildkite/test-splitter"
+    description: "Buildkite test engine client"
```
nprizal, line 25:
```suggestion
    description: "Buildkite Test Engine Client"
```

.goreleaser.yaml
```diff
@@ -86,9 +86,9 @@ docker_manifests:
 nfpms:
   - vendor: Buildkite
     id: linux-pkg
-    homepage: https://github.com/buildkite/test-splitter
+    homepage: https://github.com/buildkite/test-engine-client
     maintainer: Buildkite <support@buildkite.com>
-    description: Buildkite test splitting client
+    description: Buildkite test engine client
```
nprizal, line 91:
```suggestion
    description: Buildkite Test Engine Client
```

nprizal (approved):
Paired with @amybiyuliu and this looks good to me. @niceking Just one thing to confirm, we want the binary name to `bktec`, is that right? So the usage will be
```
# export other env vars here

bktec
```

.buildkite/steps/upload-linux-packages.sh
```diff
@@ -2,7 +2,7 @@
 
 file=${1}
 extension="${file##*.}"
-registry="test-splitter-${extension}"
+registry="test-engine-client-${extension}"
```
niceking, line 5: For thing that aren't already namespaced under "buildkite", I reckon we should call it "buildkite-test-engine-client"

.goreleaser.yaml
```diff
@@ -1,7 +1,7 @@
 version: 2
 
 release:
-  name_template: Test Splitter v{{.Version}}
+  name_template: Test Engine Client v{{.Version}}
```
niceking, line 4: As above

.buildkite/steps/upload-linux-packages.sh
nprizal: Package cloud registry is stored within `buildkite` namespace. See this [example](https://buildkite.com/organizations/buildkite/packages/registries/test-splitter-deb/packages/019106fc-6d20-781b-816e-74c39e3dee99) from the old registry.

.goreleaser.yaml
```diff
@@ -1,7 +1,8 @@
 version: 2
+project_name: test-engine-client
```
amybiyuliu, line 2: Add `project_name` here as I found the binary being put into `dist/test-splitter_*` on the test engine client pipeline build. When testing locally it puts the binary to `dist/test-engine-client*`


> building binaries
> --
>   | 2024-09-16 16:51:08 ACST | • partial build                                  match=target=darwin_amd64
>   | 2024-09-16 16:51:08 ACST | • building                                       binary=dist/test-splitter_darwin_amd64_v1/bktec

The example build: https://buildkite.com/buildkite/test-engine-client/builds/1003#0191f9b3-4701-4d79-951a-7c8a44704f89


@nprizal is this a sensible change?

amybiyuliu:
I made some minor changes to this PR in the afternoon, would be nice to have another review.
I'm not sure whether this is blocking release for https://github.com/buildkite/test-engine-client/pull/188 
@wooly please feel free to take over this, or merge this PR if it's blocking your release

.goreleaser.yaml
nprizal: Ah, interesting! g
oreleaser get the project name from repository name by default, we have renamed the repository, but the pipeline still referring to the old repo name (see image below). I think explicitly setting the project name is a good idea here.
Screenshot 2024-09-17 at 9 14 26 AM

nprizal (approved):

.goreleaser.yaml
```diff
@@ -1,7 +1,8 @@
...
 
 release:
-  name_template: Test Splitter v{{.Version}}
+  name_template: Buildkite Test Engine Client v{{.Version}}
```
nprizal, line 5:
```suggestion
  name_template: Test Engine Client v{{.Version}}
```
I don't think this is necessary. This is the title for GitHub release. The repo is already under Buildkite namespace.

PR #181, opened 2024-09-09
Filter tests regardless split by example on or off

nprizal:
### Description

To support test quarantining, we need to filter tests on the server side regardless whether split by example is on or off. With this changes, and some changes on the server side, test splitter will be able split files that has quarantined tests, without having to enable the Split by Example. Splitting slow files will only be done when Split by Example feature is on.

### DIFF
--- a/internal/api/filter_tests.go
+++ b/internal/api/filter_tests.go
@@ -14,8 +14,7 @@ type FilterTestsParams struct {
 }
 
 type FilteredTest struct {
-	Path   string `json:"path"`
-	Reason string `json:"reason"`
+	Path string `json:"path"`
 }
 
 type filteredTestResponse struct {
--- a/main.go
+++ b/main.go
@@ -271,6 +271,10 @@ func createRequestParam(ctx context.Context, cfg config.Config, files []string, 
 		})
 	}
 
+	if cfg.SplitByExample {
+		debug.Println("Splitting by example")
+	}
+
 	debug.Printf("Filtering %d files", len(files))
 	filteredFiles, err := client.FilterTests(ctx, cfg.SuiteSlug, api.FilterTestsParams{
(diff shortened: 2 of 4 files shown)


main.go
```diff
@@ -271,20 +271,6 @@ func createRequestParam(ctx context.Context, cfg config.Config, files []string,
...
-			Identifier:  cfg.Identifier,
-			Parallelism: cfg.Parallelism,
-			Branch:      cfg.Branch,
-			Tests: api.TestPlanParamsTest{
-				Files: testFiles,
-			},
-		}, nil
-	}
-
-	debug.Println("Splitting by example")
```
niceking, line 286: Do you think its worth retaining these debug messages?

niceking (approved):
nice ✨

wooly:
Awesome!

internal/api/filter_tests.go
```diff
@@ -14,8 +14,7 @@ type FilterTestsParams struct {
 }
 
 type FilteredTest struct {
-	Path   string `json:"path"`
-	Reason string `json:"reason"`
```
nprizal, line 18: I have removed the `Reason` for now because we don't use it on the client side, so the API response is not restricted to having `reason` with `string` value.

wooly: You can use `omitempty`, to make things line up?

nprizal: I can, but I'm thinking to change the type from string to an array of string, as 1 file can be slow and has quarantined. And because we don't need it on the client side, I don't see a reason why we need to parse it into the struct.

wooly: All good!

niceking (approved):
🙌

PR #182, opened 2024-09-09
Update node index and parallelism validation logic

amybiyuliu:
### Description
Currently when BUILDKITE_PARALLEL_JOB and BUILDKITE_PARALLEL_JOB_COUNT are not set. The error message looks like this. 
```
 ./run-splitter
Buildkite Test Splitter: Invalid configuration
BUILDKITE_PARALLEL_JOB was "", must be a number
BUILDKITE_PARALLEL_JOB was 0, must not be greater than -1
BUILDKITE_PARALLEL_JOB_COUNT was 0, must be greater than or equal to 1
```

We found the error message `BUILDKITE_PARALLEL_JOB was 0, must not be greater than -1` and `BUILDKITE_PARALLEL_JOB_COUNT was 0, must be greater than or equal to 1` confusing when the value of the env variables are not set

This PR updates the validation logic and prints more readable error message. 
The current code structure validates BUILDKITE_PARALLEL_JOB and BUILDKITE_PARALLEL_JOB_COUNT in two steps. 1. validate the type and presence of the env  2. validate the content of the env. 

This PR will skip the content/range validation if the first step failed for `BUILDKITE_PARALLEL_JOB_COUNT` or `BUILDKITE_PARALLEL_JOB`. The result looks like:

```
./test-splitter
Buildkite Test Splitter: Invalid configuration...
BUILDKITE_BUILD_ID must not be blank
BUILDKITE_ORGANIZATION_SLUG must not be blank
BUILDKITE_PARALLEL_JOB was "", must be a number
BUILDKITE_SPLITTER_API_ACCESS_TOKEN must not be blank
BUILDKITE_SPLITTER_SUITE_SLUG must not be blank
BUILDKITE_STEP_ID must not be blank
ResultPath must not be blank
```

### Changes

- Add a map `errs` field to Config struct to keep truck of all errors for each environment variable
- Conditionally skip value check if  `BUILDKITE_PARALLEL_JOB` or `BUILDKITE_PARALLEL_JOB_COUNT` has invalid type.

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
 func setEnv(t *testing.T) {
@@ -43,9 +44,10 @@ func TestNewConfig(t *testing.T) {
 		ResultPath:       "tmp/rspec.json",
 		SuiteSlug:        "my_suite",
 		TestRunner:       "rspec",
+		errs:             InvalidConfigError{},
 	}
 
-	if diff := cmp.Diff(c, want); diff != "" {
+	if diff := cmp.Diff(c, want, cmpopts.IgnoreUnexported(Config{})); diff != "" {
 		t.Errorf("config.New() diff (-got +want):\n%s", diff)
 	}
 }
@@ -85,7 +87,7 @@ func TestNewConfig_MissingConfigWithDefault(t *testing.T) {
 		ResultPath:       "tmp/rspec.json",
 	}
 
-	if diff := cmp.Diff(c, want); diff != "" {
+	if diff := cmp.Diff(c, want, cmpopts.IgnoreUnexported(Config{})); diff != "" {
 		t.Errorf("config.New() diff (-got +want):\n%s", diff)
 	}
 }
--- a/internal/config/error.go
+++ b/internal/config/error.go
@@ -6,7 +6,7 @@ import (
 	"strings"
 )
 
-// InvalidConfigError is an error that contains a list of all invalid fields in the config.
+// InvalidConfigError is an error that contains a map of all validation errors on each field of the configuration.
 type InvalidConfigError map[string][]error
 
 func (i InvalidConfigError) Error() string {
@@ -18,16 +18,6 @@ func (i InvalidConfigError) Error() string {
 	}
 	sort.Strings(errs)
 	return strings.Join(errs, "\n")
-}
-
-func (i InvalidConfigError) Unwrap() []error {
-	errs := make([]error, 0, len(i))
-	for field, value := range i {
-		for _, v := range value {
-			errs = append(errs, fmt.Errorf("%s %s", field, v))
-		}
-	}
-	return errs
 }
 
 func (e InvalidConfigError) appendFieldError(field, format string, v ...any) {
--- a/internal/config/error.go
+++ b/internal/config/error.go
@@ -7,12 +7,14 @@ import (
 )
 
 // InvalidConfigError is an error that contains a list of all invalid fields in the config.
-type InvalidConfigError []invalidFieldError
+type InvalidConfigError map[string][]error
 
 func (i InvalidConfigError) Error() string {
 	var errs []string
-	for _, err := range i {
-		errs = append(errs, err.Error())
+	for field, value := range i {
+		for _, v := range value {
+			errs = append(errs, fmt.Sprintf("%s %s", field, v))
+		}
 	}
 	sort.Strings(errs)
 	return strings.Join(errs, "\n")
@@ -20,31 +22,17 @@ func (i InvalidConfigError) Error() string {
 
 func (i InvalidConfigError) Unwrap() []error {
 	errs := make([]error, 0, len(i))
-	for _, e := range i {
-		errs = append(errs, e)
+	for field, value := range i {
+		for _, v := range value {
+			errs = append(errs, fmt.Errorf("%s %s", field, v))
+		}
 	}
 	return errs
 }
 
-func (e *InvalidConfigError) appendFieldError(field, format string, v ...any) {
-	*e = append(*e, invalidFieldError{
-		name: field,
-		err:  fmt.Errorf(format, v...),
-	})
-}
-
-// invalidFieldError is the detailed error of an invalid rule for a field in the config.
-type invalidFieldError struct {
-	// name is the name of the field.
-	name string
-	// err is the error.
-	err error
-}
-
-func (f invalidFieldError) Error() string {
-	return fmt.Sprintf("%s %s", f.name, f.err)
-}
-
-func (f invalidFieldError) Unwrap() error {
-	return f.err
+func (e InvalidConfigError) appendFieldError(field, format string, v ...any) {
+	if e[field] == nil {
+		e[field] = make([]error, 0)
+	}
+	e[field] = append(e[field], fmt.Errorf(format, v...))
 }
--- a/internal/config/read.go
+++ b/internal/config/read.go
@@ -69,16 +69,14 @@ func (c *Config) readFromEnv() error {
 
 	parallelism := os.Getenv("BUILDKITE_PARALLEL_JOB_COUNT")
 	parallelismInt, err := strconv.Atoi(parallelism)
-	c.Parallelism = parallelismInt
-	if err != nil {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %q, must be a number", parallelism)
+	if err == nil {
+		c.Parallelism = &parallelismInt
 	}
 
 	nodeIndex := os.Getenv("BUILDKITE_PARALLEL_JOB")
 	nodeIndexInt, err := strconv.Atoi(nodeIndex)
-	c.NodeIndex = nodeIndexInt
-	if err != nil {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %q, must be a number", nodeIndex)
+	if err == nil {
+		c.NodeIndex = &nodeIndexInt
 	}
 
 	if len(errs) > 0 {
--- a/internal/config/read.go
+++ b/internal/config/read.go
@@ -30,7 +30,6 @@ import (
 // If we are going to support other CI environment in the future,
 // we will need to change where we read the configuration from.
 func (c *Config) readFromEnv() error {
-	var errs InvalidConfigError
 
 	c.AccessToken = os.Getenv("BUILDKITE_SPLITTER_API_ACCESS_TOKEN")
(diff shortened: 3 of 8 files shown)


amybiyuliu:
@DrJosh9000 can I pls have a sanity check with you if this is the correct use case of pointers 😅 
or perhaps there are better ways to solve the problem of not being able to tell `nil` vs `0`

niceking:
I think your instinct is right that we need to tell apart nil values from 0 values. I don't think you need to use pointers to do this though? Does something like this work? 

```
parallelism := os.Getenv("BUILDKITE_PARALLEL_JOB_COUNT")
	parallelismInt, err := strconv.Atoi(parallelism)
	if err != nil {
		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %q, must be a number", parallelism)
		c.Parrallelism = nil // maybe you don't have to explicitly set this 
	} else {
		c.Parallelism = parallelismInt
	}
```

amybiyuliu:
Hey @niceking I did try your approach first, Go decided to give `0` to those fields as we didn't explicitly set the value. That's why I go down the pointer path 😅 Sorry about not making this clear in the PR description.

internal/config/validate.go
```diff
@@ -12,20 +12,23 @@ func (c *Config) validate() error {
 		errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
 	}
 
-	if got, min := c.Parallelism, 1; got < min {
+	if c.Parallelism == nil {
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "not set in environment")
+	} else if got, min := *c.Parallelism, 1; got < min {
```
nprizal, line 17: Should we keep the max validation?
```suggestion
	if c.Parallelism == nil {
		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "not set in environment")
	} else {
	  if got, min := *c.Parallelism, 1; got < min {
	    ...
	  }
	  if got, max := *c.Parallelism, 1000; got > max {
	    ...
	  }
	}
```

internal/config/read.go
```diff
@@ -69,16 +69,14 @@ func (c *Config) readFromEnv() error {
 
 	parallelism := os.Getenv("BUILDKITE_PARALLEL_JOB_COUNT")
 	parallelismInt, err := strconv.Atoi(parallelism)
-	c.Parallelism = parallelismInt
-	if err != nil {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %q, must be a number", parallelism)
+	if err == nil {
```
nprizal, line 72: Maybe add a comment why we're ignoring the error, and mention that it will be validated in the `validate` function instead.

nprizal:
> I don't think you need to use pointers to do this though?

@niceking @amybiyuliu  The [zero value](https://go.dev/ref/spec#The_zero_value) for `int` type in Go is `0`, and you cannot assign `nil` value to a var with `int` type. If we want to distinguish between empty value (`nil`) and zero value (`0`), we need to use a pointer.

internal/config/validate.go
```diff
@@ -12,20 +12,23 @@ func (c *Config) validate() error {
 		errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
 	}
 
-	if got, min := c.Parallelism, 1; got < min {
+	if c.Parallelism == nil {
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "not set in environment")
```
niceking, line 16: this should say `must not be blank` to bring it in line with the other error messages

internal/config/validate.go
```diff
@@ -12,20 +12,23 @@ func (c *Config) validate() error {
...
-	}
-
-	if got, min := c.NodeIndex, 0; got < 0 {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %d, must be greater than or equal to %d", got, min)
-	}
-
-	if got, max := c.NodeIndex, c.Parallelism-1; got > max {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "was %d, must not be greater than %d", got, max)
+	if c.NodeIndex == nil {
+		errs.appendFieldError("BUILDKITE_PARALLEL_JOB", "not set in environment")
```
niceking, line 22: this should say `must not be blank` to bring it in line with the other error messages

niceking:
@nprizal TIL and yes makes sense that the nil value of int type is zero

It seems like a bit of a code smell to me that the config struct now has some pointers, and some declared types. It seems pretty counter intuitive that you have to dereference some of the config, and can just directly use parts of the others. Would this behaviour also be surprising in Go? If so, I think we should lean towards a refactor that doesn't introduce pointers

internal/config/validate.go
amybiyuliu: ohh good catch! will add the max validation back

nprizal:
I wouldn't worry about having some pointers in the config struct. But yes, I agree that we should find a solution that don't require us to use pointers. We can keep the config as `int`, and keep the existing `read` function as it is, then change the validation logic to only compare the node index and parallelism when parallelism is valid (i.e. greater than 0)

```
if parallelism < 1 {
   // must be greater than or equal to 1
} else {
  if node_index > (parallelism - 1)
    // must not be greater than parallelism
}
```

amybiyuliu:
Just reading your approach @nprizal It would be nice to avoid using pointer
I think we still need to tell whether `node_index` is blank or 0 🤔 (0 is allowed for `node_index`)

internal/config/error.go
```diff
@@ -7,44 +7,32 @@ import (
...
+	for field, value := range i {
+		for _, v := range value {
+			errs = append(errs, fmt.Sprintf("%s %s", field, v))
+		}
 	}
 	sort.Strings(errs)
 	return strings.Join(errs, "\n")
 }
 
 func (i InvalidConfigError) Unwrap() []error {
```
nprizal, line 23: I think we don't need this function anymore

internal/config/config_test.go
```diff
@@ -85,7 +87,7 @@ func TestNewConfig_MissingConfigWithDefault(t *testing.T) {
 		ResultPath:       "tmp/rspec.json",
 	}
 
-	if diff := cmp.Diff(c, want); diff != "" {
+	if diff := cmp.Diff(c, want, cmpopts.IgnoreUnexported(Config{})); diff != "" {
```
amybiyuliu, line 90: `errs` is unexported field of the config struct. IgnoreUnexported option ignores the unexported fields during comparison.

internal/config/validate.go
```diff
@@ -6,52 +6,57 @@ import (
...
 	if c.MaxRetries < 0 {
-		errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
+		c.errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
 	}
 
-	if got, min := c.Parallelism, 1; got < min {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %d, must be greater than or equal to %d", got, min)
-	}
+	if c.errs["BUILDKITE_PARALLEL_JOB"] == nil {
+		if got, min := c.NodeIndex, 0; got < 0 {
```
nprizal, line 15: Can we add a comment explaining why we are doing this for `PARALLEL_JOB` and `PARALLEL_JOB_COUNT`, but not for other envs? Also, I think the order of the validation matters here, maybe write that in the comment as well?

internal/config/read_test.go
```diff
@@ -144,8 +149,60 @@ func TestConfigReadFromEnv_MissingStepId(t *testing.T) {
...
 			t.Errorf("config.readFromEnv() got = %v, want = %v", got, want)
 		}
 	}
 }
+
+func TestConfigReadFromEnv_InvalidParallelJob(t *testing.T) {
+	os.Setenv("BUILDKITE_SPLITTER_BASE_URL", "")
+	os.Setenv("BUILDKITE_SPLITTER_MODE", "")
+	os.Setenv("BUILDKITE_SPLITTER_TEST_CMD", "")
+	os.Setenv("BUILDKITE_SPLITTER_RETRY_COUNT", "")
```
nprizal, line 162:
```suggestion
```
I don't think we need this?

internal/config/read_test.go
```diff
@@ -144,8 +149,60 @@ func TestConfigReadFromEnv_MissingStepId(t *testing.T) {
...
+	if got := invConfigError.Error(); got != want {
+		t.Errorf("config.readFromEnv() got = %v, want = %v", got, want)
+	}
+}
+
+func TestConfigReadFromEnv_InvalidParallelJobCount(t *testing.T) {
+	os.Setenv("BUILDKITE_SPLITTER_BASE_URL", "")
+	os.Setenv("BUILDKITE_SPLITTER_MODE", "")
+	os.Setenv("BUILDKITE_SPLITTER_TEST_CMD", "")
+	os.Setenv("BUILDKITE_SPLITTER_RETRY_COUNT", "")
```
nprizal, line 188:
```suggestion
```

nprizal (commented):
Also, can we change the PR title as it's not only changes the messaging, but changes how the node index and parallelism get validated?

internal/config/validate.go
amybiyuliu: That's good call. I added some comments. Let me know if it's still unclear 😀

internal/config/validate.go
```diff
@@ -6,52 +6,66 @@ import (
...
 	if c.MaxRetries < 0 {
-		errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
+		c.errs.appendFieldError("BUILDKITE_SPLITTER_RETRY_COUNT", "was %d, must be greater than or equal to 0", c.MaxRetries)
 	}
 
-	if got, min := c.Parallelism, 1; got < min {
-		errs.appendFieldError("BUILDKITE_PARALLEL_JOB_COUNT", "was %d, must be greater than or equal to %d", got, min)
-	}
+	// We validate BUILDKITE_PARALLEL_JOB and BUILDKITE_PARALLEL_JOB_COUNT in two steps.
+	// 1. Validate the type and presence of BUILDKITE_PARALLEL_JOB and BUILDKITE_PARALLEL_JOB_COUNT when reading them from the environment.
```
nprizal, line 15:
```suggestion
	// 1. Validate the type and presence of BUILDKITE_PARALLEL_JOB and BUILDKITE_PARALLEL_JOB_COUNT when reading them from the environment. See readFromEnv() in ./read.go.
```

nprizal (approved):
Looks good to me.

PR #188, opened 2024-09-16
[TET-367] Fix path bug when using shared examples.

wooly:
### Description

SBE is broken when shared examples are involved, because of the way we generate the `Path` field for each example.

Prior to this change, the file path and line number of the shared example would be used as the `Path` value, but this wont work because RSpec doesn't know which file to run the shared example for.

This change updates the `Path` field to use the identifier for the test instead.


internal/runner/rspec_test.go
```diff
@@ -365,3 +365,28 @@ func TestRspecGetExamples_WithOtherFormatters(t *testing.T) {
...
+		TestCommand: "rspec",
+	})
+	files := []string{"./fixtures/rspec/spec/specs_with_shared_examples_spec.rb"}
+	got, err := rspec.GetExamples(files)
+
+	want := []plan.TestCase{
+		{
+			Identifier: "./fixtures/rspec/spec/specs_with_shared_examples_spec.rb[1:1:1]",
+			Name:       "behaves like a shared example",
+			Path:       "./fixtures/rspec/spec/specs_with_shared_examples_spec.rb[1:1:1]",
```
nprizal, line 380: 👍

nprizal (approved):
Good catch!

niceking:
Already merged but good catch 👍  I was just verifying this against the server side implementation. Would be good to do a release for this and test on bk^3

PR #191, opened 2024-09-17
Handle billing error

nprizal:
### Description

We have updated the API to require the new billing plan. This PR updates the client to handle the billing error and switch to fallback mode. 

This is the output will be after the changes

```
http://api.buildkite.localhost/v2/analytics/organizations/buildkite/suites/buildkite/test_plan?identifier=fruitszxc/rspec
Billing Error: Test Splitting is not enabled on your plan: please upgrade at https://buildkite.com/organizations/buildkite/billing.
Using fallback mode. Your build may take longer than usual.
+++ Buildkite Test Engine Client: Running tests
rspec ./test/spec/fruits/apple_spec.rb ./test/spec/fruits/tomato_spec.rb --format documentation --format json --out ./tmp/rspec.json
...
```

### DIFF
--- a/main.go
+++ b/main.go
@@ -208,14 +208,14 @@ func fetchOrCreateTestPlan(ctx context.Context, apiClient *api.Client, cfg confi
 
 	handleError := func(err error) (plan.TestPlan, error) {
 		if errors.Is(err, api.ErrRetryTimeout) {
-			fmt.Println("Could not fetch or create plan from server, using fallback mode. Your build may take longer than usual.")
+			fmt.Println("\033[33mCould not fetch or create plan from server, using fallback mode. Your build may take longer than usual.\033[0m")
 			p := plan.CreateFallbackPlan(files, cfg.Parallelism)
 			return p, nil
 		}
 
 		if billingError := new(api.BillingError); errors.As(err, &billingError) {
-			fmt.Println(billingError.Message)
-			fmt.Println("Using fallback mode. Your build may take longer than usual.")
+			fmt.Println("\033[33m" + billingError.Message)
+			fmt.Println("Using fallback mode. Your build may take longer than usual.\033[0m")
 			p := plan.CreateFallbackPlan(files, cfg.Parallelism)
 			return p, nil
 		}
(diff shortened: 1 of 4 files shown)


amybiyuliu (approved):
LGTM 🥳

wooly:
Do we want to be a bit more vocal if we're using fallback mode in this scenario?

Maybe with a few asterisks and some spacing around the error message?

main.go
```diff
@@ -212,6 +212,14 @@ func fetchOrCreateTestPlan(ctx context.Context, apiClient *api.Client, cfg confi
 			p := plan.CreateFallbackPlan(files, cfg.Parallelism)
 			return p, nil
 		}
+
+		if billingError := new(api.BillingError); errors.As(err, &billingError) {
+			fmt.Println(billingError.Message)
+			fmt.Println("Using fallback mode. Your build may take longer than usual.")
```
niceking, line 218: I agree with @wooly's comment about being more shouty that this has happened, also maybe the second error line should say something like `Falling back to non-intelligent split, your build may take longer than usual` just so it matches up with marketing copy?

main.go
```diff
@@ -208,10 +208,18 @@ func fetchOrCreateTestPlan(ctx context.Context, apiClient *api.Client, cfg confi
...
 		if errors.Is(err, api.ErrRetryTimeout) {
-			fmt.Println("Could not fetch or create plan from server, using fallback mode. Your build may take longer than usual.")
+			fmt.Println("\033[33mCould not fetch or create plan from server, falling back to non-intelligent splitting. Your build may take longer than usual.\033[0m")
 			p := plan.CreateFallbackPlan(files, cfg.Parallelism)
 			return p, nil
 		}
+
+		if billingError := new(api.BillingError); errors.As(err, &billingError) {
+			fmt.Println("\033[33m" + billingError.Message)
+			fmt.Println("Falling back to non-intelligent splitting. Your build may take longer than usual.\033[0m")
```
nprizal, line 218: I made the text yellow and updated the message.
What do you think @niceking @wooly ?
I'm planning to refactor the logging using a logger package in the future to tidy up the logging, but I think this is fine for now.

niceking: Yeah it's fine for now, I wonder how the yellow highlight looks in light mode on Buildkite? 
Screenshot 2024-09-17 at 12 13 11 PM
Is it a similar yellow to this WARN here? Perhaps for accessibility/readability reasons, might be better use like the alert or danger emoji or similar?

nprizal: Yes, it will be that yellow color. I have updated the log to use the ⚠️ emoji instead.

niceking (approved):

PR #192, opened 2024-09-17
Update release pipeline with new AWS role and parameter store key

nprizal:
We renamed the release pipeline from `Test Splitter - Release` to `Test Engine Client - Release`. This breaks the AWS authorization because the role that was used in the pipeline is scoped with the pipeline slug. This PR updates the release pipeline configuration with the new AWS role that is scoped with the new pipeline slug.


wooly (approved):

wooly:
Nice catch!

PR #193, opened 2024-09-17
Update release script to login to correct Packagecloud docker registry

nprizal:
We changed the registry to push the docker image to Packagecloud, but we haven't change the release script to login to the correct registry. This PR updates the release script to login to correct Packagecloud docker registry.


amybiyuliu (approved):
Good catch!

PR #194, opened 2024-09-17
Update linux package name

nprizal:
The linux package still built as `test-engine-client` instead of `bktec`. 
Screenshot 2024-09-17 at 5 26 22 PM

By default `goreleaser` uses project name which is `test-engine-client` as package name, however, we want to use `bktec` as binary/package name. This PR updates the `nfpms` configuration to build linux packages with `bktec` name.

I created a snapshot locally, and this is the generated artifacts
Screenshot 2024-09-17 at 5 31 33 PM

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -22,6 +22,9 @@ builds:
     ldflags: "-X 'main.Version=v{{ .Version }}'"
     binary: bktec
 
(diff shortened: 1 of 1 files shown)


amybiyuliu (approved):
Thanks for making the change 🙌

PR #196, opened 2024-09-18
Skip sendMetadata when using fallback plan

nprizal:
We don't want to send metadata when the client falls back to non-intelligent split, because there is very little value of tracking the execution of the test with non-intelligent split. Moreover, when the client falls back to non-intelligent split, there might not be a test plan in the server, which means that sending the metadata back to the server is even useless.

This PR adds `Fallback` attribute to the `TestPlan` struct, updates the `CreateFallbackPlan` method to return test plan with `Fallback: true`, and updates the `main` function to skip sending metadata when the plan has `Fallback: true`.

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,7 +92,9 @@ func main() {
 		}
 
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
-			sendMetadata(ctx, apiClient, cfg, timeline)
+			if !testPlan.Fallback {
+				sendMetadata(ctx, apiClient, cfg, timeline)
+			}
 			logErrorAndExit(exitError.ExitCode(), "%s exited with error: %v", testRunner.Name(), err)
 		}
 
@@ -100,14 +102,19 @@ func main() {
 	}
 
 	if testResult.Status == runner.RunStatusFailed {
-		sendMetadata(ctx, apiClient, cfg, timeline)
+		if !testPlan.Fallback {
+			sendMetadata(ctx, apiClient, cfg, timeline)
+		}
+
 		if failedCount := len(testResult.FailedTests); failedCount > 1 {
 			logErrorAndExit(1, "%s exited with %d failures", testRunner.Name(), failedCount)
 		}
 		logErrorAndExit(1, "%s exited with 1 failure", testRunner.Name())
 	}
 
-	sendMetadata(ctx, apiClient, cfg, timeline)
+	if !testPlan.Fallback {
+		sendMetadata(ctx, apiClient, cfg, timeline)
+	}
 }
 
 func createTimestamp() string {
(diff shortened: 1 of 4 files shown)


main.go
```diff
@@ -92,22 +92,29 @@ func main() {
 		}
 
 		if exitError := new(exec.ExitError); errors.As(err, &exitError) {
-			sendMetadata(ctx, apiClient, cfg, timeline)
+			if !testPlan.Fallback {
```
niceking, line 95: Wow I really miss the `unless` of ruby 😂

main.go
```diff
@@ -92,22 +92,29 @@ func main() {
...
+
 		if failedCount := len(testResult.FailedTests); failedCount > 1 {
 			logErrorAndExit(1, "%s exited with %d failures", testRunner.Name(), failedCount)
 		}
 		logErrorAndExit(1, "%s exited with 1 failure", testRunner.Name())
 	}
 
-	sendMetadata(ctx, apiClient, cfg, timeline)
+	if !testPlan.Fallback {
+		sendMetadata(ctx, apiClient, cfg, timeline)
```
niceking, line 116: 🤔 seeing as you have to introduce this conditional check three times, does it make more sense to put the check inside the `sendMetadata` function

wooly: Do we want to rename the method to something like `sendMetadataUnlessFallback` to make it clearer?

nprizal: I'm thinking of adding `skip bool` to the args, and keep the name as is.

```
func sendMetadata(..., skip bool) {
  if !skip {
    ...
  }
}

// main()
...
sendMetadata(..., plan.Fallback)
...
```

niceking: lol ok now that I see the change, putting the skip inside the `sendMetadata` method doesn't seem great either, maybe the original implementation was better

niceking: I think perhaps it made more sense to put the skip inside the `sendMetadata` function if it was operating on an object rather than a payload. Perhaps having the repeated conditional makes more sense in a procedural flow? I'm gonna approve and let you decide which is better, seems like its mostly a style thing and I am no Go style expert

niceking (approved):
Change looks good, up to you where to check the fallback condition

main.go
nprizal: Agree that having `if !plan.Fallback { sendMetadata(...) }` makes more sense in procedural flow. But adding `skip` to the `sendMetadata` enforces us to explicitly tell whether to send the metadata or not (it will be caught by compiler if we don't specify). Having the condition outside `sendMetada` function will be unchecked, and we don't have unit test for `main()`. So, I prefer the `skip` flag, but I don't have a strong opinion on this.

@DrJosh9000 any hot take?

DrJosh9000: > @DrJosh9000 any hot take?

Yes. I think repeating `if !plan.Fallback { sendMetadata(...) }` three times is not a big deal:

* each instance of the conditional is 3 lines of code
* it makes it clear without having to read `sendMetadata` that it is skipped if `plan.Fallback`
* `sendMetadata` remains focused on sending metadata, and isn't involved in the business logic of deciding _whether_ to send metadata

PR #197, opened 2024-09-18
Making test runner env mandatory

amybiyuliu:


main.go
```diff
@@ -50,7 +50,7 @@ func main() {
 
 	testRunner, err := runner.DetectRunner(cfg)
 	if err != nil {
-		logErrorAndExit(16, "Unsupported value for BUILDKITE_TEST_ENGINE_TEST_RUNNER")
+		logErrorAndExit(16, `Unsupported value for BUILDKITE_TEST_ENGINE_TEST_RUNNER %q: %v`, cfg.TestRunner, err)
```
nprizal, line 53: Is there any reason you use backtick `` ` `` instead of double quote? `"`?

amybiyuliu: ohhh no I was trying to quote the value of the runner env started with `"%s"` then I realized %q will just do the quote for me but forgot to update the backtick! I'll make the change

nprizal (approved):

PR #199, opened 2024-09-24
Configure golangci-lint

nprizal:
Configure [golangci-lint](https://golangci-lint.run) for better linting. It has [gosec](https://github.com/securego/gosec) support that is useful to catch security vulnerabilities. Followings are the issues caught by the configuration set in this PR.

```
internal/api/client.go:182:28: printf: non-constant format string in call to fmt.Errorf (govet)
			return resp, fmt.Errorf(errorResp.Message)
			                        ^
internal/config/config.go:45:15: Error return value of `c.readFromEnv` is not checked (errcheck)
	c.readFromEnv()
	             ^
internal/config/config.go:46:12: Error return value of `c.validate` is not checked (errcheck)
	c.validate()
	          ^
internal/runner/command.go:42:23: Error return value of `cmd.Process.Signal` is not checked (errcheck)
				cmd.Process.Signal(sig)
				                  ^
internal/runner/discover.go:25:20: Error return value of `parsedPattern.Glob` is not checked (errcheck)
	parsedPattern.Glob(func(path string, d fs.DirEntry, err error) error {
	                  ^
internal/runner/jest.go:161:16: unusedresult: result of slices.Replace call not used (govet)
	slices.Replace(words, outputIdx, outputIdx+1, j.ResultPath)
	              ^
internal/runner/jest.go:187:16: unusedresult: result of slices.Replace call not used (govet)
	slices.Replace(words, outputIdx, outputIdx+1, j.ResultPath)
```


internal/config/config.go
```diff
@@ -42,8 +42,10 @@ type Config struct {
 // It returns Config struct and an InvalidConfigError if there is an invalid configuration.
 func New() (Config, error) {
 	c := Config{errs: InvalidConfigError{}}
-	c.readFromEnv()
-	c.validate()
+
+	// TODO: remove error from readFromEnv and validate functions
```
nprizal, line 46: The error is now included in the `Config` struct. We can remove the `error` from `readFromEnv` and `validate` returns.

wooly (approved):

PR #200, opened 2024-10-03
Exclude pending examples from Split by Examples

nprizal:
### Description

When Split by Example is enabled, `bktec` will do the dry run for filtered files to get the examples within those files. The examples within those files could include a skipped example (i.e. using `xit` in RSpec). Currently, `bktec` includes those skipped examples when requesting the test plan, which reduce the accuracy of the test plan.

We need to exclude skipped tests when requesting the test plan to have more accurate test plan.

### DIFF
--- a/internal/runner/fixtures/rspec/spec/with_skipped_examples_spec.rb
+++ b/internal/runner/fixtures/rspec/spec/with_skipped_examples_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe("Spec with skipped examples") do
     fail
   end
 
-  it("skipped using pending option", pending: "skipped") do
+  it("pending example", pending: "not implemented") do
     fail
   end
 end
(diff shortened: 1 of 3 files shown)


internal/runner/fixtures/rspec/spec/with_skipped_examples_spec.rb
```diff
@@ -0,0 +1,17 @@
...
+
+  xit("skiped using xit") do
+    fail
+  end
+
+  it("skipped using skip option", skip: "skipped") do
+    fail
+  end
+
+  it("skipped using pending option", pending: "skipped") do
```
nprizal, line 14: Test example marked as `pending` (as opposed to `skip`) will still be executed by RSpec. So we should find a way to include `pending` examples, but exclude `skip` examples.

PR #201, opened 2024-10-07
Add Playwright support

nprizal:
Adds support for Playwright test runner. The Playwright runner currently supports basic splitting, include/exclude patterns, retries. Split by example is not yet supported.

### Validation
Build the binary and run it from inside `internal/runner/testdata/playwright`
```
export BUILDKITE_BUILD_ID=abc
export BUILDKITE_ORGANIZATION_SLUG=my-org
export BUILDKITE_PARALLEL_JOB=0
export BUILDKITE_PARALLEL_JOB_COUNT=2
export BUILDKITE_STEP_ID=xyz
export BUILDKITE_TEST_ENGINE_DEBUG_ENABLED=true
export BUILDKITE_TEST_ENGINE_RESULT_PATH=test-results/results.json
export BUILDKITE_TEST_ENGINE_TEST_RUNNER=playwright
bktec
```

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -10,10 +10,12 @@ go install gotest.tools/gotestsum@v1.8.0
 go install github.com/pact-foundation/pact-go/v2@latest
 pact-go -l DEBUG install
 
-# Install node dependencies for js runner tests
+# Install dependencies for js runner tests
 cd internal/runner/testdata
 yarn install
-cd ../../..
+cd playwright
+yarn playwright install
+cd ../../../..
 
 echo '+++ Running tests'
 gotestsum --junitfile "junit-${BUILDKITE_JOB_ID}.xml" -- -count=1 -coverprofile=cover.out -failfast "$@" ./...
--- a/internal/runner/playwright.go
+++ b/internal/runner/playwright.go
@@ -48,7 +48,8 @@ func (p Playwright) Run(testCases []string, retry bool) (RunResult, error) {
 
 	cmd := exec.Command(cmdName, cmdArgs...)
 
-	fmt.Printf("%s %s\n", cmdName, strings.Join(cmdArgs, " "))
+	// double new line is required because Playwright output removes the last line
+	fmt.Printf("%s %s\n\n", cmdName, strings.Join(cmdArgs, " "))
 	err = runAndForwardSignal(cmd)
 
 	if err == nil { // note: returning success early
--- a/internal/runner/playwright.go
+++ b/internal/runner/playwright.go
@@ -1,0 +1,91 @@
+package runner
+
+import (
+	"fmt"
+	"os/exec"
+	"slices"
+	"strings"
+
+	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/plan"
+	"github.com/kballard/go-shellquote"
+)
+
+var _ = TestRunner(Playwright{})
+
+type Playwright struct {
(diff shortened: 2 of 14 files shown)


.buildkite/steps/tests.sh
```diff
@@ -10,10 +10,12 @@ go install gotest.tools/gotestsum@v1.8.0
 go install github.com/pact-foundation/pact-go/v2@latest
 pact-go -l DEBUG install
 
-# Install node dependencies for js runner tests
+# Install dependencies for js runner tests
 cd internal/runner/testdata
 yarn install
-cd ../../..
+cd playwright
```
niceking, line 16: not quite following this - what creates the playwright directory here?

internal/runner/playwright.go
```diff
@@ -0,0 +1,155 @@
...
+	"os/exec"
+	"slices"
+	"strings"
+
+	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/plan"
+	"github.com/kballard/go-shellquote"
+)
+
+var _ = TestRunner(Playwright{})
```
niceking, line 17: what does this line do? seems that `_` is an unused variable?

internal/runner/playwright.go
```diff
@@ -0,0 +1,155 @@
...
+	if p.TestCommand == "" {
+		p.TestCommand = "npx playwright test"
+	}
+
+	if p.TestFilePattern == "" {
+		p.TestFilePattern = "**/{*.spec,*.test}.{ts,js}"
+	}
+
+	if p.TestFileExcludePattern == "" {
+		p.TestFileExcludePattern = "**/node_modules"
```
niceking, line 37: 🤔 so by default we exclude node modules? If the customer specifies an exclude pattern, should we still exclude the node modules as well?

niceking (commented):
Looks good, just a few questions. 

Will there be a README update for this as well?

internal/runner/playwright.go
DrJosh9000: This is a way of testing at compile time whether `Playwright` implements the `TestRunner` interface. The value is not needed at run time and thrown away (hence the name `_`).

This kind of thing is often not necessary. In this case a `Playwright` is returned from `DetectRunner`, and it declares the type of its first return to be `TestRunner`. If `Playwright` didn't satisfy the `TestRunner` interface, then you'd get the same compiler error within `DetectRunner` as you would here.

.buildkite/steps/tests.sh
nprizal: playwright directory contains playwright sample project, commited to this repo.

internal/runner/playwright.go
nprizal: Discussed in sync, we will exclude `**/node_modules` as built-in feature, without having to set it through exclude pattern. I will open a separate PR for this.

internal/runner/playwright.go
nprizal: I copied this from the other runners. But yes, I think this is redundant. I'll remove this from all runners in a separate PR.

niceking (approved):

wooly:
Nice job!

wooly:
When this version is released do we need to update the setup docs?
Any readme updates to go in with this too?

nprizal:
> When this version is released do we need to update the setup docs?
> Any readme updates to go in with this too?

README will be updated when we release the version and requires restructuring. It will be done in a separate PR.

PR #205, opened 2024-10-14
Tidy up test data

nprizal:
We have test data across different folders, which makes it a little bit messy. Test data in Golang usually stored within `testdata` folder in each package. This PR moved test data to the `testdata` folder and removes test data that is no longer used.


internal/runner/jest_test.go
```diff
@@ -53,6 +53,8 @@ func TestNewJest(t *testing.T) {
 }
 
 func TestJestRun(t *testing.T) {
+	mockCwd(t, "./testdata/jest")
```
nprizal, line 56: This is required because Jest need to find `jest.config.json` which is now stored inside `testdata/jest` directory.

wooly (approved):
Nice!

PR #206, opened 2024-10-15
Add Cypress Runner

nprizal:
Adds support for the Cypress test runner. The Cypress runner **doesn't support** test retry and split by example due to limitation in Cypress to run individual test. This PR also add yarn workspace configuration for `jest` and `cypress` examples to improve the developer experience.

### Validation
Build the binary and run it from inside `internal/runner/testdata/cypress`
```
export BUILDKITE_BUILD_ID=abc
export BUILDKITE_ORGANIZATION_SLUG=my-org
export BUILDKITE_PARALLEL_JOB=0
export BUILDKITE_PARALLEL_JOB_COUNT=2
export BUILDKITE_STEP_ID=xyz
export BUILDKITE_TEST_ENGINE_TEST_RUNNER=cypress
bktec
```


internal/runner/cypress.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return RunResult{Status: RunStatusError}, fmt.Errorf("failed to build command: %w", err)
+	}
+
+	cmd := exec.Command(cmdName, cmdArgs...)
+
+	fmt.Printf("%s %s\n", cmdName, strings.Join(cmdArgs, " "))
+	err = runAndForwardSignal(cmd)
+
+	if err == nil { // note: returning success early
```
DrJosh9000, line 51: Why return success early here? The other way around (https://google.github.io/styleguide/go/decisions#indent-error-flow) would be more conventional.

internal/runner/cypress_test.go
```diff
@@ -0,0 +1,205 @@
...
+	"os/exec"
+	"syscall"
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/kballard/go-shellquote"
+)
+
+func TestCypressRun(t *testing.T) {
+	mockCwd(t, "./testdata/cypress")
```
DrJosh9000, line 14: [Note for something defined outside this PR] `mockCwd` _actually really for real_ changes the working directory for the duration of the test, so it's not a "mock"!

DrJosh9000 (approved):
Some good looking code! As usual I have some remarks.

internal/runner/cypress.go
nprizal: Ah, good point! In other runners, we handle the error differently and returning early is more intuitive. But not in this case, so yes we should do the other way here.

internal/runner/cypress_test.go
nprizal: Oh thanks! I didn't think deeply about this. Maybe `changeWorkingDirectory`  more accurate?

PR #207, opened 2024-10-23
Exclude `node_modules` all the time

nprizal:
Currently, we set the default of `BUILDKITE_TEST_ENGINE_TEST_EXCLUDE_PATTERN` to be `**/node_modules` to exclude files within `node_modules` directory.  If customers configure the `BUILDKITE_TEST_ENGINE_TEST_EXCLUDE_PATTERN` to their needs, files inside the `node_modules` can accidentally get picked up by the client. We need to exclude files inside `node_modules` all the time.

### DIFF
--- a/internal/runner/cypress.go
+++ b/internal/runner/cypress.go
@@ -26,10 +26,6 @@ func NewCypress(c RunnerConfig) Cypress {
 		c.TestCommand = "npx cypress run --spec {{testExamples}}"
 	}
 
-	if c.TestFileExcludePattern == "" {
-		c.TestFileExcludePattern = "**/node_modules"
-	}
-
 	if c.TestFilePattern == "" {
 		c.TestFilePattern = "**/*.cy.{js,jsx,ts,tsx}"
 	}
--- a/internal/runner/discover.go
+++ b/internal/runner/discover.go
@@ -36,6 +36,12 @@ func discoverTestFiles(pattern string, excludePattern string) ([]string, error) 
 			}
 			return nil
 		}
+
+		// Skip the node_modules directory
+		if d.Name() == "node_modules" {
+			return fs.SkipDir
+		}
+
 		// Skip directories that happen to match the include pattern - we're
 		// only interested in files.
 		if d.IsDir() {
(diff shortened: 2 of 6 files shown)


nprizal:
> I'm not sure I 100% agree with this change. For me, we should either use the default value if an override isn't provided, or do what the user says they want to do.
> 
> Did someone raise this as an issue?

This is to avoid accidentally running a test inside `node_modules` when customers configure the exclude patterns. Without this, the customer will have to write a complicated glob pattern to exclude `node_modules` plus other patterns. Customer might not aware that they have to also exclude `node_modules` explicitly if they want to configure the exclude patterns. 

Do you think there will be a case where a customer wants to run a test inside `node_modules`, I doubt so?

wooly:
> This is to avoid accidentally running a test inside `node_modules` when customers configure the exclude patterns. Without this, the customer will have to write a complicated glob pattern to exclude `node_modules` plus other patterns. Customer might not aware that they have to also exclude `node_modules` explicitly if they want to configure the exclude patterns.
> 
> Do you think there will be a case where a customer wants to run a test inside `node_modules`, I doubt so?

I doubt there will be either, it's just the principle of 'We provide all of the behaviour through the default or you provide all of the behaviour by overriding that default' feels a bit nicer than 'We provide all of the behaviour through the default or you provide some of the behaviour with us adding to it if you override it'. It feels like this is a downside of using the environment variables as a configuration pattern. File-based formats like YAML or JSON or TOML would allow you to specify an array of regex values to exclude rather than having to craft a single regex to handle it all. We don't have that though, so it's a bit moot.

Happy for this to be non-blocking and just something to keep in mind.

niceking (approved):
Hmmm yeah I feel like its pretty hard to say what the best default is and I don't feel like I know enough about js to properly comment. Good thoughts from everyone to consider, best course of action is probably just to release _something_ and then wait for customers to give us feedback 😁

nprizal:
Thanks for putting your thoughts, everyone. I'll go forward with these changes. We can update this if customers find this harder to use, or when we switch to using a configuration file that will allow us to have more flexibility in configuring the pattern.

PR #208, opened 2024-10-23
Remove interface check for all runners.

nprizal:
In each runner, we have the following code to check if the runner implements the `TestRunner` interface. 
```
var _ = TestRunner(Runner{})
```
This is redundant, because the compiler already checks this inside `DetectRunner` function.

### DIFF
--- a/internal/runner/cypress.go
+++ b/internal/runner/cypress.go
@@ -11,8 +11,6 @@ import (
 	"github.com/kballard/go-shellquote"
 )
 
-var _ = TestRunner(Cypress{})
-
 type Cypress struct {
 	RunnerConfig
 }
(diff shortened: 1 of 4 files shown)


wooly:
This was a pattern suggested by @wolfeidau, so might be worth confirming if it's still useful or not, as I don't know enough Go!

nprizal:
> This was a pattern suggested by @wolfeidau, so might be worth confirming if it's still useful or not, as I don't know enough Go!

This was brought up by @DrJosh9000 in this [PR](https://github.com/buildkite/test-engine-client/pull/201#discussion_r1811523691).

DrJosh9000 (approved):
I don't feel particularly strongly about whether or not to do this. Ultimately it's up to you. LGTM.

PR #209, opened 2024-10-23
Update default test command for jest to be less opinionated

nprizal:
Currently, the default command for jest is `yarn test ...`. Because not all customer will use `yarn`, some are using `npm`, and the script name can be anything than `test`, we need to update to be something less opinionated. I opt to using `npx` which came by default in node.

### DIFF
--- a/internal/runner/cypress.go
+++ b/internal/runner/cypress.go
@@ -23,7 +23,7 @@ func (c Cypress) Name() string {
 
 func NewCypress(c RunnerConfig) Cypress {
 	if c.TestCommand == "" {
-		c.TestCommand = "cypress run --spec {{testExamples}}"
+		c.TestCommand = "npx cypress run --spec {{testExamples}}"
 	}
 
 	if c.TestFileExcludePattern == "" {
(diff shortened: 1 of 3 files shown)


wooly:
This is fine but I would have expected a README update as part of this?

nprizal:
> This is fine but I would have expected a README update as part of this?

We need to start updating the README during the release process, not when merging the PR. If we update the README before the release, customer can find something in the README that is not yet released.

wooly:
I'm conflicted about this, because in that flow the README would be out of date with the repo, so if I wanted to build it from source the instructions would be wrong.

I wonder if there's a way to state that this README is for the currently released version, and have a parallel README.next.md or something for the un-released version, and then as part of the release process, we make that one the actual one.

wooly:
This is non-blocking feedback so will give this a tick, but something to think about!

wooly (approved):

nprizal:
> I'm conflicted about this, because in that flow the README would be out of date with the repo, so if I wanted to build it from source the instructions would be wrong.

> I wonder if there's a way to state that this README is for the currently released version, and have a parallel README.next.md or something for the un-released version, and then as part of the release process, we make that one the actual one.

That's a good point. I don't expect people to build from source unless they are contributing the tools. This whole README/docs is a bit of a mess at this moment. I have brought up this issue with the team internally.

PR #213, opened 2024-10-30
Escape special characters in testCases in Jest retry command

nprizal:
Jest use regex to execute an individual test. When there is a failed test and retry is enabled, `bktec` passes the failed test name as regex to Jest to execute. However, test name may contain special characters that need to be escaped. 

This pull request fixes the issue and ensures that special characters are escaped using the `regexp.QuoteMeta` function before constructing the test name pattern. This prevents any potential issues or errors when retrying tests with special characters in their names.

### DIFF
--- a/internal/runner/jest_test.go
+++ b/internal/runner/jest_test.go
@@ -299,6 +299,33 @@ func TestJestRetryCommandNameAndArgs_HappyPath(t *testing.T) {
 	}
 }
 
+func TestJestRetryCommandNameAndArgs_WithSpecialCharacters(t *testing.T) {
+	testCases := []string{"this will fail", "test with special characters .+*?()|[]{}^$"}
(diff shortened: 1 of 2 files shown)


internal/runner/jest_test.go
```diff
@@ -299,6 +299,33 @@ func TestJestRetryCommandNameAndArgs_HappyPath(t *testing.T) {
 	}
 }
 
+func TestJestRetryCommandNameAndArgs_WithSpecialCharacters(t *testing.T) {
+	testCases := []string{"this will fail", "test with special characters .+*?()|[]{}^$"}
```
niceking, line 303: How come this first string is called`"this will fail"`? It made me think that string was going to fail for some reason? But actually I think the intent of your test is to check that special characters are escaped correctly?

nprizal: I wanted to test if the `testNamePattern` is passed correctly if there is multiple test. I just used the same test name from other test. I'll change the test name to avoid confusion.

wooly (approved):

PR #214, opened 2024-10-30
Print actual command when running Jest.

nprizal:
`bktec` doesn't print the actual command when running Jest. This PR fixes the issue by moving the printing from each test runner's `Run` function to the underlying `runAndForwardSignal` function used within the `Run` function. This avoids having to print the command in each test runner.

### DIFF
--- a/internal/runner/command.go
+++ b/internal/runner/command.go
@@ -1,9 +1,11 @@
 package runner
 
 import (
+	"fmt"
 	"os"
 	"os/exec"
 	"os/signal"
+	"strings"
 	"syscall"
 )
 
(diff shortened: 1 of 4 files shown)


niceking (commented):
Have you verified that it doesn't affect the formatting of the run commands for the other test runners? I see from the PR that they all have custom formatting atm

nprizal:
> Have you verified that it doesn't affect the formatting of the run commands for the other test runners? I see from the PR that they all have custom formatting atm

Yes, I have verified all test runners locally.

wooly (approved):

PR #215, opened 2024-11-04
Send test runner to create test plan endpoint

nprizal:
### Description

The test runner is missing from the request params for the create test plan endpoint that triggers the server to falls back to the default test runner `rspec`, which caused an issue when splitting test for other test runners, such as `jest`.

I have removed the `Runner` enum from `api` package, as I feel like it is a bit redundant where we already validate the test runner in the `runner.DetectRunner` function. Removing this enum keeps the source simple and clean.


DrJosh9000 (approved):
LGTM

wooly (approved):
Nice :+1:

PR #216, opened 2024-11-06
Add script to record support ended for Pact.

wooly:
### Description

Use it like: `bin/pact-record-support-ended v0.6.0`


nprizal (approved):
Nice! ✨

PR #224, opened 2024-11-11
Fix Cypress command

nprizal:
### Description

bktec replaces `{{testExamples}}` with the list of test files, separated by a space, which is not recommended by cypress. 

```bash
⚠ Warning: It looks like you're passing --spec a space-separated list of arguments:
"./cypress/e2e/click.cy.js ./cypress/e2e/visit.cy.js"
This will work, but it's not recommended.
If you are trying to pass multiple arguments, separate them with commas instead:
  cypress run --spec arg1,arg2,arg3
  
The most common cause of this warning is using an unescaped glob pattern. If you are
trying to pass a glob pattern, escape it using quotes:
  cypress run --spec "**/*.spec.js"
```

This PR update the Cypress command to join the list of tests with `,`.

### DIFF
--- a/internal/runner/cypress.go
+++ b/internal/runner/cypress.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os/exec"
 	"slices"
+	"strings"
 
 	"github.com/buildkite/test-engine-client/internal/debug"
 	"github.com/buildkite/test-engine-client/internal/plan"
@@ -73,11 +74,11 @@ func (c Cypress) commandNameAndArgs(cmd string, testCases []string) (string, []s
 		return "", []string{}, err
 	}
 	idx := slices.Index(words, "{{testExamples}}")
+	specs := strings.Join(testCases, ",")
 	if idx < 0 {
-		words = append(words, "--spec")
-		words = append(words, testCases...)
+		words = append(words, "--spec", specs)
 	} else {
-		words = slices.Replace(words, idx, idx+1, testCases...)
+		words[idx] = specs
 	}
 
 	return words[0], words[1:], nil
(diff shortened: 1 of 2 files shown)


internal/runner/cypress.go
```diff
@@ -73,11 +74,12 @@ func (c Cypress) commandNameAndArgs(cmd string, testCases []string) (string, []s
...
 	}
 	idx := slices.Index(words, "{{testExamples}}")
+	specs := strings.Join(testCases, ",")
 	if idx < 0 {
 		words = append(words, "--spec")
-		words = append(words, testCases...)
+		words = append(words, specs)
 	} else {
-		words = slices.Replace(words, idx, idx+1, testCases...)
+		words = slices.Replace(words, idx, idx+1, specs)
```
DrJosh9000, line 82: That's equivalent to:

```suggestion
		words[idx] = specs
```

internal/runner/cypress.go
```diff
@@ -73,11 +74,12 @@ func (c Cypress) commandNameAndArgs(cmd string, testCases []string) (string, []s
 		return "", []string{}, err
 	}
 	idx := slices.Index(words, "{{testExamples}}")
+	specs := strings.Join(testCases, ",")
 	if idx < 0 {
 		words = append(words, "--spec")
-		words = append(words, testCases...)
+		words = append(words, specs)
```
DrJosh9000, line 80: Also, this could be combined:

```suggestion
		words = append(words, "--spec", specs)
```

wooly (approved):
Giving this a tick pending changes suggested by @DrJosh9000

PR #226, opened 2024-11-19
Display report after run

nprizal:
### Description

Print report at the end of run that includes run statistics (number of tests that are passed/muted/failed), list of muted tests, and list of failed tests.

```
========== Buildkite Test Engine Report  ==========
+--------+-------------+---+
| Passed | first run   | 2 |
+        +-------------+---+
|        | on retry    | 0 |
+--------+-------------+---+
| Muted  | passed      | 0 |
+        +-------------+---+
|        | failed      | 1 |
+--------+-------------+---+
| Failed |             | 1 |
+--------+-------------+---+
|                TOTAL | 4 |
+--------+-------------+---+

+++ Muted Tests:
- Tomato is vegetable (failed)

+++ Failed Tests:
- Banana is red
===================================================
```

### Changes

- `RunResult` struct is repurposed to keep track of run result, including all individual test results. The struct now contains the logic to calculate the run status, muted test, and failed tests. It also has a function to return the statistics of the run.  (previously it only contains failed test and status as a static attribute)
- Each runner is initialized with the `runResult` struct, and records all test results to that struct after each run.
- Muted tests filter logic in main package's `runTestsWithRetry` has been removed.
- Report will be printed at the end of run, and run statistics is displayed as ASCII table using [tablewriter](https://github.com/olekukonko/tablewriter) package.


internal/runner/run_result.go
```diff
@@ -0,0 +1,149 @@
...
+	RunStatusError  RunStatus = "error"
+)
+
+// RunResult is a struct to keep track the results of a test run.
+// It contains the logics to record test results, calculate the status of the run.
+type RunResult struct {
+	// tests is a map containing individual test results.
+	tests map[string]*TestResult
+	// mutedTestLookup is a map containing the test identifiers of muted tests.
+	// This list might contain tests that are not part of the current run (i.e. belong to a different node).
```
niceking, line 21: Why?

niceking (commented):
Will defer to @DrJosh9000 on the code, but seems about right in terms of what is in the report 😄 We can always change/update this later. 

One niggle I had is that it now doesn't easily show the total passed and total failed, maybe we could append that to the end? 

It'll also be interesting to see how this looks with different numbers on the bk/bk test suite, and the possible double up of the 🔥 Test Analytics Failures 🔥 summary. Another reason why the collector/bktec double up is annoying 😅 

Might be worth a once-over by @jameshill as well!

internal/runner/run_result.go
nprizal: The test plan currently returns muted tests for all nodes in a single field, it doesn't distribute the muted tests to the specific node. We don't want to display a muted test that belongs to a test file assigned for another node.

niceking:
Ooooo yeah I like that summary line!

> It can be a bit noisy. Also, the 🔥 Test Analytics Failures 🔥 only available in ruby collector at the moment.

Yeah but that's what we use on bk/bk which is what we'll be recording for the demo so it'll be rather obvious 😛

wooly:
I'm a bit confused by this:
```
+--------+-------------+---+
| Muted  | passed      | 0 |
+        +-------------+---+
|        | failed      | 1 |
+--------+-------------+---+
```

The muted tests will never be assigned to the node in the first place, so how does the node know how many muted tests to display?

The other information like
```
+--------+-------------+---+
| Passed | first run   | 2 |
+        +-------------+---+
|        | after retry | 0 |
+--------+-------------+---+
```

Is per-node so makes sense, but the muted tests are not per-node, so it might not make sense to group them as a row in the table like this?

niceking:
@wooly muted tests are assigned to the node? Because they still need to be run

wooly:
Sorry!! Confusing muted and skipped, grrr!

main.go
```diff
@@ -146,11 +147,12 @@ func sendMetadata(ctx context.Context, apiClient *api.Client, cfg config.Config,
...
 }
 
-func runTestsWithRetry(testRunner TestRunner, testsCases *[]plan.TestCase, maxRetries int, mutedTest []plan.TestCase, timeline *[]api.Timeline) (runner.RunResult, error) {
+func runTestsWithRetry(testRunner TestRunner, testsCases *[]plan.TestCase, maxRetries int, mutedTests []plan.TestCase, timeline *[]api.Timeline) (runner.RunResult, error) {
 	attemptCount := 0
 
-	var testResult runner.RunResult
-	var err error
+	// load mutedTests from test plan into the runResult
+	runResult := testRunner.Result()
```
DrJosh9000, line 196: I'm confused by the data flow.

This seems to be the only call to `Result()`, a method added to the test runner interface so that each test runner can provide ...a result? _before_ running the tests?? merely so that you can load muted tests into it???

And as a consequence of the new interface method, each test runner must not only implement it but pre-initialise a result and save it in a field, and also keep it updated?

I'd reconsider this approach - at least the `Result` method. Maybe the initial `*RunResult` should be an argument passed to `testRunner.Run`, and the runner updates that `RunResult` that it points to.

main.go
```diff
@@ -113,22 +115,63 @@ func main() {
...
 }
 
+func printReport(runResult runner.RunResult) {
+	fmt.Println("========== Buildkite Test Engine Report  ==========")
+
+	// Print statistics
+	statistics := runResult.Statistics()
+	data := [][]string{
+		{"Passed", "first run", strconv.Itoa(statistics.PassedOnFirstRun)},
+		{"Passed", "after retry", strconv.Itoa(statistics.PassedAfterRetry)},
```
jameshill, line 140: can this be passed **on retry** instead of _after_

main.go
```diff
@@ -113,22 +115,63 @@ func main() {
...
+		os.Exit(1)
 	}
 
 	if !testPlan.Fallback {
 		sendMetadata(ctx, apiClient, cfg, timeline)
 	}
 }
 
+func printReport(runResult runner.RunResult) {
+	fmt.Println("========== Buildkite Test Engine Report  ==========")
```
niceking, line 134: Can we please make this always expanded?

main.go
nprizal: Yes, you are right. The muted tests need to be loaded before each run so it can correctly update the result. I was trying to avoid having the main function initialize the `RunResult` struct, so I initialize it inside each test runner. But looking back at it now, I can see how it can be confusing and unnecessarily complicated. 

I have updated the implementation by passing the `*RunResult` to the `testRunner.Run` function. Let me know what you think.

internal/runner/cypress.go
```diff
@@ -28,28 +28,32 @@ func NewCypress(c RunnerConfig) Cypress {
...
 	err = runAndForwardSignal(cmd)
 
 	if err != nil {
-		return RunResult{Status: RunStatusError}, err
+		result.err = err
+		return err
 	}
 
-	return RunResult{Status: RunStatusPassed}, nil
+	return nil
```
DrJosh9000, line 56: `result.err` can be set to whatever `err` is? i.e. this trailing section can probably be simplified to:

```suggestion
result.err = err
return err
```

internal/runner/run_result.go
```diff
@@ -0,0 +1,155 @@
...
+			}
+		} else if testResult.Status == TestStatusPassed {
+			if testResult.ExecutionCount > 1 {
+				passedAfterRetry++
+			} else {
+				passedOnFirstRun++
+			}
+		} else if testResult.Status == TestStatusFailed {
+			failed++
+		}
```
DrJosh9000, line 144: I would lean more on `switch` here, even when the conditions are comparing different things; `switch` checks cases sequentially:

```suggestion
		switch {
		case testResult.Muted:
			switch testResult.Status {
			case TestStatusPassed:
				mutedPassed++
			case TestStatusFailed:
				mutedFailed++
			}
			
		case testResult.Status == TestStatusPassed:
			if testResult.ExecutionCount > 1 {
				passedAfterRetry++
			} else {
				passedOnFirstRun++
			}
			
		case testResult.Status == TestStatusFailed:
			failed++
		}
```

DrJosh9000 (approved):
Very nice work!

PR #230, opened 2024-11-28
Handle errors outside of example

nprizal:
### Description

When there is a failure outside of test (e.g. syntax error, runtime error), bktec might exit with `0` exit code when there is no other test failure. These changes handle the failure outside of tests to avoid bktec reporting a false positive (false "pass").

### Changes
1. To handle test error (not test failure), we need to distinguish between error in the command (runner exits prematurely) and error in the test (test finished with error). The `Run` function in each runner will return the followings:
   - test finished without failure/error: `RunResult.Status: 'passed', err: nil`
   - test finished with failure(s): `RunResult.Status: 'failed', err: nil`
   - test finished with error(s): `RunResult.Status: 'error', err: nil`
   - test didn't finish or finished without readable result: `RunResult.Status: 'unknown', err: any error from the runner process`
1. I updated and tidied up the "exit" logic in main function. The main function will now exit with code `1` when the `RunResult.Status` is `error` (e.g. there are errors outside of tests). It will also print the run status in the report.

Screenshot 2024-12-09 at 5 14 58 PM

### DIFF
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -69,7 +69,6 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 		}
 		commandName, commandArgs, err := j.commandNameAndArgs(j.TestCommand, testPaths)
 		if err != nil {
-			result.err = err
 			return fmt.Errorf("failed to build command: %w", err)
 		}
 
@@ -81,7 +80,6 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 		}
 		commandName, commandArgs, err := j.retryCommandNameAndArgs(j.RetryTestCommand, testNames)
 		if err != nil {
-			result.err = err
 			return fmt.Errorf("failed to build command: %w", err)
 		}
 
@@ -91,14 +89,12 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 	err = runAndForwardSignal(cmd)
 
 	if ProcessSignaledError := new(ProcessSignaledError); errors.As(err, &ProcessSignaledError) {
-		result.err = err
 		return err
 	}
 
 	report, parseErr := j.ParseReport(j.ResultPath)
 	if parseErr != nil {
 		fmt.Println("Buildkite Test Engine Client: Failed to read Jest output, tests will not be retried.")
-		result.err = err
 		return err
 	}
 
@@ -123,6 +119,10 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 		}
 	}
 
+	for i := 0; i < report.NumRuntimeErrorTestSuites; i++ {
+		result.errors = append(result.errors, ErrOutsideOfTest)
+	}
+
 	return nil
 }
 
@@ -134,8 +134,9 @@ type JestExample struct {
 }
 
 type JestReport struct {
-	NumFailedTests int
-	TestResults    []struct {
+	NumFailedTests            int
+	NumRuntimeErrorTestSuites int
+	TestResults               []struct {
 		AssertionResults []JestExample
 	}
 }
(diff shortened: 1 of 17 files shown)


internal/runner/run_result.go
```diff
@@ -149,11 +156,12 @@ func (r *RunResult) Statistics() RunStatistics {
 	}
 
 	return RunStatistics{
-		Total:            len(r.tests),
+		Total:            len(r.tests) + len(r.errors),
```
nprizal, line 159: I'm not sure if we should add the error count to the total because error doesn't necessarily have 1to1 relationship with a test. But, it will be confusing if we exclude it from the total because the error is displayed in the same table. Any thoughts? @wooly @niceking

wooly: I feel like if the test throws an exception then the test runner should pass this straight up to the orchestrator and it should fail CI immediately. Is this what happens if you just run straight rspec with a test that throws an exception, or does it keep going until it gets to the end?

nprizal: > I feel like if the test throws an exception then the test runner should pass this straight up to the orchestrator and it should fail CI immediately. Is this what happens if you just run straight rspec with a test that throws an exception, or does it keep going until it gets to the end?

It depends on the runner. RSpec and Jest will keep running the other tests, but Playwright will immediately exit with error.

wooly: Ah all good then. I think it's reasonable to add an entry to the table for the count of errors detected.

> because error doesn't necessarily have 1to1 relationship with a test

Do you have an example of this?

nprizal: To clarify, any error that prevents the runner from “completing” the tests is immediately terminated by bktec. By “completing” the tests, I mean that the runner has finished all the tests, regardless of their status or any errors outside of the expected ones (e.g., runtime errors).

nprizal: > > because error doesn't necessarily have 1to1 relationship with a test
> 
> Do you have an example of this?
> 

The following test file will be reported as having 1 error, and 0 examples.
```
RSpec.describe "bad syntax" do

  before do
    if "this block is missing an end"
  end

  it "passes" do
    expect(true).to be_truthy
  end

  it "fails" do
    expect(true).to be_falsey
  end
end
```

wooly: Ah I see. In that case then I wouldn't allow the error count to contribute towards the 'total', since it's not directly related to the count of tests.

nprizal: That's what I thought initially. What about we show the total, then the error, something like this
```
+++ ========== Buildkite Test Engine Report  ==========
┌─────────────┬─────────────┬───┐
│ Passed      │ first run   │ 1 │
│             ├─────────────┼───┤
│             │ on retry    │ 0 │
├─────────────┼─────────────┼───┤
│ Muted       │ passed      │ 0 │
│             ├─────────────┼───┤
│             │ failed      │ 0 │
├─────────────┼─────────────┼───┤
│ Failed      │             │ 1 │
├─────────────┴─────────────┼───┤
│                Total test │ 2 │
├───────────────────────────┼───┤
│               Total error │ 1 │
└───────────────────────────┴───┘
```

nprizal: We decided to remove the error count from the table.

DrJosh9000 (commented):
Overall pretty good

internal/runner/jest.go
```diff
@@ -123,6 +119,10 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 		}
 	}
 
+	for i := 0; i < report.NumRuntimeErrorTestSuites; i++ {
```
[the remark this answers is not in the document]
nprizal: That’s a good point. The original implementation printed the number of errors in the report. However, we’ve decided to remove that entirely. It’s no longer necessary and I’ll update the PR.

wooly (approved):
I've tested this with RSpec and it's working as expected. bktec exits with a status code of `1` for errors, and the runner is logged at the top of the report.

DrJosh9000 (approved):
LGTM

PR #237, opened 2024-12-10
Write a script to run e2e test

nprizal:
### Description

We now have four runners supported in `bktec`. Since we have sample projects for each test runner that we use for unit and integration tests, I thought it would be great to use those sample projects to run end-to-end tests. This PR adds a script that runs `bktec` (using `go run …`) against a sample project located in `internal/runner/testdata`.

To run the e2e test:
```shell
#.bin/e2e <runner> <parallel-job>
.bin/e2e rspec 0
```

### DIFF
--- a/internal/runner/testdata/rspec/Gemfile
+++ b/internal/runner/testdata/rspec/Gemfile
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "rspec"
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,7 @@ internal/api/pacts/
 
 dist/
 node_modules/
+
+internal/runner/testdata/jest/jest-result.json
+
+internal/runner/testdata/rspec/rspec-result.json
--- a/README.md
+++ b/README.md
@@ -95,3 +95,18 @@ bktec may exit with a variety of exit statuses, outlined below:
   SIGABRT, the exit status returned will be equal to 128 plus the signal number.
   For example, if the runner raises a SIGSEGV, the exit status will be (128 +
   11) = 139.
+
+## Development
+
+Make sure you have Go, Ruby, and Node.js installed in your environment. You can follow the installation guides for each of these tools:
+
+- [Go Installation Guide](https://golang.org/doc/install)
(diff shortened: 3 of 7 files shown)


internal/runner/testdata/rspec/Gemfile
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
```
nprizal, line 1: I added thi so we can run `bktec` from this project using the default command `bundle exec rspec ...`

wooly (commented):
I've been testing this this morning and think it could maybe use some automation/instruction around bootstrapping.

I'm getting an error relating to `bundle install` not being run for the rspec test suite, and a module not found error when trying to run it for playwright.

niceking:
I like this! This just enables running the tests for each language in the bktec repo yeah? 

One comment on the usage interface:

```
#.bin/e2e <runner> <parallel-job>
.bin/e2e rspec 0
```

I understand why we'd wanna specify the framework as a top level argument into the script here, but I'm less convinced that the script should mandate passing in a parallel job? What possible variance do you see in this argument? IMO if parallel job count has a default then its fine that parallel job has one too

nprizal:
> I like this! This just enables running the tests for each language in the bktec repo yeah?

Yes, that's right


> One comment on the usage interface:
> 
> ```
> #.bin/e2e <runner> <parallel-job>
> .bin/e2e rspec 0
> ```
> 
> I understand why we'd wanna specify the framework as a top level argument into the script here, but I'm less convinced that the script should mandate passing in a parallel job? What possible variance do you see in this argument? IMO if parallel job count has a default then its fine that parallel job has one too

That's a good point. I've removed parallel job from the arg. You can still set that variable manually as
```
BUILDKITE_PARALLEL_JOB=1 bin/e2e
```

nprizal:
> I've been testing this this morning and think it could maybe use some automation/instruction around bootstrapping.
> 
> I'm getting an error relating to `bundle install` not being run for the rspec test suite, and a module not found error when trying to run it for playwright.

@wooly I created a `bin/setup` to install all dependencies for sample projects. This will be useful for engineers contributing to this repo as well.

niceking (approved):

PR #241, opened 2024-12-16
Add retry count

niceking:
### Description

This PR reads `BUILDKITE_RETRY_COUNT` from the env and sends it through as part of the request to fetch a test plan. Defaults to a value of `0` if the env var isn't found. 

### Testing

Updated Pact

### DIFF
--- a/internal/api/fetch_test_plan.go
+++ b/internal/api/fetch_test_plan.go
@@ -10,8 +10,8 @@ import (
 
 // FetchTestPlan fetchs a test plan from the server.
 // ErrRetryTimeout is returned if the client failed to communicate with the server after exceeding the retry limit.
-func (c Client) FetchTestPlan(ctx context.Context, suiteSlug string, identifier string) (*plan.TestPlan, error) {
-	url := fmt.Sprintf("%s/v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s", c.ServerBaseUrl, c.OrganizationSlug, suiteSlug, identifier)
+func (c Client) FetchTestPlan(ctx context.Context, suiteSlug string, identifier string, jobRetryCount int) (*plan.TestPlan, error) {
+	url := fmt.Sprintf("%s/v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s&job_retry_count=%d", c.ServerBaseUrl, c.OrganizationSlug, suiteSlug, identifier, jobRetryCount)
 
 	var testPlan plan.TestPlan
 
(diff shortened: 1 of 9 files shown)


internal/api/create_test_plan.go
```diff
@@ -19,6 +19,7 @@ type TestPlanParams struct {
 	Identifier  string             `json:"identifier"`
 	Parallelism int                `json:"parallelism"`
 	Branch      string             `json:"branch"`
+	RetryCount  int                `json:"retryCount"`
```
nprizal, line 22: Actually, this value is expected in the fetch test plan endpoint, not to create test plan endpoint. The retried job will not hit the create test plan endpoint because the test plan is already been created when the job is retried. Also, the server expected the value as `job_retry_count`.

internal/api/fetch_test_plan.go
```diff
@@ -10,8 +10,8 @@ import (
 
 // FetchTestPlan fetchs a test plan from the server.
 // ErrRetryTimeout is returned if the client failed to communicate with the server after exceeding the retry limit.
-func (c Client) FetchTestPlan(ctx context.Context, suiteSlug string, identifier string) (*plan.TestPlan, error) {
-	url := fmt.Sprintf("%s/v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s", c.ServerBaseUrl, c.OrganizationSlug, suiteSlug, identifier)
+func (c Client) FetchTestPlan(ctx context.Context, suiteSlug string, identifier string, jobRetryCount int) (*plan.TestPlan, error) {
+	url := fmt.Sprintf("%s/v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s&job_retry_count=%i", c.ServerBaseUrl, c.OrganizationSlug, suiteSlug, identifier, jobRetryCount)
```
niceking, line 14: @nprizal i think you're expecting a string? Not an int? If I change this to a `%s` will it just work?

internal/api/fetch_test_plan.go
```diff
@@ -10,8 +10,8 @@ import (
 
 // FetchTestPlan fetchs a test plan from the server.
 // ErrRetryTimeout is returned if the client failed to communicate with the server after exceeding the retry limit.
-func (c Client) FetchTestPlan(ctx context.Context, suiteSlug string, identifier string) (*plan.TestPlan, error) {
-	url := fmt.Sprintf("%s/v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s", c.ServerBaseUrl, c.OrganizationSlug, suiteSlug, identifier)
+func (c Client) FetchTestPlan(ctx context.Context, suiteSlug string, identifier string, jobRetryCount int) (*plan.TestPlan, error) {
+	url := fmt.Sprintf("%s/v2/analytics/organizations/%s/suites/%s/test_plan?identifier=%s&job_retry_count=%v", c.ServerBaseUrl, c.OrganizationSlug, suiteSlug, identifier, jobRetryCount)
```
niceking, line 14: @nprizal your PR expects this as a string, not 100% sure how `%v` will present this value here? It's an int

nprizal: You can use `%d` to print an `int` in a string template. Although, `%v` will default to `%d` when passed an `int`. But I'd use `%d` to be explicit.

nprizal (approved):
Nice one!

PR #248, opened 2025-01-07
Skip filtering and splitting files for non rspec runners

nprizal:
### Description

As part of the implementation of test skipping, `bktec` will filter files that contain skipped test, and split those files into individual examples. However, splitting files into examples only supported in RSpec currently. If there are files that contain 
skipped test in non RSpec runners, `bktec` will be terminated because it cannot split those files.
Screenshot 2025-01-08 at 11 23 34 AM

This PR fix the issue by skipping the filtering and splitting process for non RSpec runners.

### Changes

- Extract filtering & splitting files into separate function
- Skip filtering & splitting files for non RSpec runners

### Testing

I did manual test using a locally running server. I skipped a test in a Jest suite and run `bktec` for that suite.

### DIFF
--- a/main.go
+++ b/main.go
@@ -362,12 +362,9 @@ func fetchOrCreateTestPlan(ctx context.Context, apiClient *api.Client, cfg confi
 	return testPlan, nil
 }
 
-// createRequestParam creates the request parameters for the test plan with the given configuration and files.
-// The files should have been filtered by include/exclude patterns before passing to this function.
-// If SplitByExample is disabled (default), it will return the default params that contain all the files.
-// If SplitByExample is enabled, it will split the slow files into examples and return it along with the rest of the files.
-//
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -376,68 +373,90 @@ func createRequestParam(ctx context.Context, cfg config.Config, files []string,
 		})
 	}
 
+	// Splitting files by example is only supported for rspec runner.
+	if runner.Name() != "RSpec" {
```
niceking, line 377: 👍

niceking (approved):
@nprizal and I had a sync about this approach (only filtering files for rspec) and its pros and cons. While this means that the behaviour for rspec is different from other languages, which makes the flow of bktec harder to understand, it's advantage is that we're not doing extra API requests we don't need. There isn't any reason for non-rspec languages to hit the filter tests endpoint right now and if there are we'll have to update the client anyways, so from a future maintenance angle its also a good solution

Issue #236, opened 2024-12-10
[Question] "On Retry" is > 0. But where are we retrying those tests?


jonathanyeong:
Hey team, we migrated to using the new bktec (from v0.8.0!) and enjoying the extra observability so far. I had a question around "On retry". I see on retry greater then 0, but I don't see bktec actually retrying the test. Am I understanding this value correctly?

<img width="409" alt="SCR-20241210-jxkb" src="https://github.com/user-attachments/assets/11ed8a8f-2d7d-42fd-aac0-2be21fd61056">

If we compare to an actual failing test, I can see that bktec shows the retries
<img width="583" alt="SCR-20241210-jwza" src="https://github.com/user-attachments/assets/e68e704e-e9b4-4328-9484-fd0742ca0633">

For context, migrating from 0.8.0 has resulted in test times that are overall a minute or two slower. So I wonder if that's the retry mechanism doing its thing.

nprizal:
Hey @jonathanyeong, thanks for reaching out. We’re facing a similar issue in Buildkite. It turns out that there are a few tests that share the same description. `bktec` and Test Engine use the combination of `scope` (example group description in Rspec) and `name` (example description in Rspec) to identify tests. If there are tests with the same scope and name, `bktec` reports them as passed on retry because it mistakenly believes that the “same” test has already been run.

We’re investigating this issue and will see if we can find a solution. In the meantime, could you please check if all test descriptions are unique? If you’re using RSpec (and Rubocop), you can use [RSpec/RepeatedDescription](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecrepeateddescription) and [RSpec/RepeatedExampleGroupDescription](https://docs.rubocop.org/rubocop-rspec/cops_rspec.html#rspecrepeatedexamplegroupdescription) to check for duplicates.

As for the test getting slower, there could be several factors contributing to this. Retry is one of them. I wouldn’t expect upgrading from `0.8.0` to `1.2.0` to slow down your test suites. To confirm if the test suite is indeed getting slower, you can check your test runs from the Test Splitting page in your Buildkite Test Suite. You can also look into individual runs to see the impact of job retry.

PR #252, opened 2025-01-15
Fix issue where a test incorrectly reported as "Passed on Retry"

nprizal:
### Description

We found a bug where a test marked as "Passed on retry". This is happening when there are different tests with the same `scope` and `name`. Currently, `bktec` will use `scope` and `name` to identify a test. If there are 2 tests with the same `scope` and `name` it will treat those tests as the same test with 2 execution counts, leading to incorrect reporting. The initial implementation was made to follow the server-side implementation so that `bktec` can match a local test with the list of muted tests received from the server.

This PR resolves the issues by introducing 2 different ways of identifying a test. When reading the result from test runner, `bktec` will identify a test using `scope`, `name` and `path` to make sure tests with the same `name` and `scope` counted as separate test. However, when checking if a test is muted, it identifies the test using `scope` and `name` only to follow the server-side implementation.

This PR will resolve https://github.com/buildkite/test-engine-client/issues/236


internal/runner/test_result.go
```diff
@@ -18,6 +18,16 @@ type TestResult struct {
 	Muted          bool
 }
 
+// testIdentifier returns a unique identifier for a test case based on its scope, name and path.
+// Different tests can have the same name and scope, therefore the path is included in the identifier
+// to make it unique.
 func testIdentifier(testCase plan.TestCase) string {
+	return testCase.Scope + "/" + testCase.Name + "/" + testCase.Path
```
niceking, line 25: Is path information available for all supported frameworks?

niceking (approved):

internal/runner/test_result.go
nprizal: Yes, except for Cypress, as we don't currently parse/read Cypress result because Cypress can't output result into file out of the box. We already read test path for RSpec and Playwright, so no changes are required. As for Jest, I updated the runner to read the test path in this PR.

niceking: Ok, mute and retry aren't even supported in Cypress so that is fine :)

PR #253, opened 2025-01-19
Update changelog for 1.3.2

nprizal:
### Description

Update CHANGELOG for v1.3.1

https://github.com/buildkite/test-engine-client/compare/v1.3.1...v1.3.2-rc.1

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.3.2 - 2025-01-20
+- Fix issue where a test incorrectly reported as "Passed on Retry".
+
 ## 1.3.1 - 2025-01-10
 - Fix issue where non-RSpec runners would terminate when attempting to split by example, as splitting by example is only supported in RSpec.
 


CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 1.3.2 - 2025-01-20
+- Fix issue where a test incorrectly reported as "Passed on Retry".
```
niceking, line 4:
```suggestion
- Fix issue for rspec where a test incorrectly reported as "Passed on Retry".
```

niceking (approved):
One minor comment but looks good otherwise

CHANGELOG.md
nprizal: I'd avoid specifying RSpec there because the issue can happen in other runner (Jest & Playwright) if the test have same `scope` and `name` with other test. Maybe
```suggestion
- Fix issue where a test with the same name and scope incorrectly reported as "Passed on Retry".
```

niceking: Yeah my intention was to say that the issue is fixed *for* rspec - as the previous wording sounded like the issue was fixed in general.

Obviously the wording still isn't clear... I don't think we should say "name and scope" as those are terms we've invented internally ourselves. I also think we shouldn't think about this too hard 🤷‍♀️

nprizal: Yes, but this also fixed the issue for jest and playwright, not only rspec.

I'll just merge as it currently is.

niceking: Ahhh my bad - I thought you said previously it was only fixed for rspec 🙇‍♀️

PR #258, opened 2025-01-21
Handle 409 status codes the same as we do 500s

wooly:
### Description

We currently return an error in the 5XX range when the client loses the race to create the plan. This is not ideal as it's skewing our error budgets, so we're going to return 409 instead.


nprizal (approved):
Looks good to me

PR #265, opened 2025-02-10
Update CHANGELOG for v1.3.3

nprizal:
### Description

Update CHANGELOG for v1.3.1

https://github.com/buildkite/test-engine-client/compare/v1.3.2...dc82b223d5b27ac2a09887cf35b1ddf3d889f709


wooly (approved):
[nit] non-blocking, but perhaps it should be 'server-side'

PR #275, opened 2025-03-11
Bazel BEP/BES

pda:
Bazel [Build Event Service](https://bazel.build/remote/bep#build-event-service) gRPC listener for the [Build Event Protocol](https://bazel.build/remote/bep).

More news as it happens…


pda:
Still keen, but not this specific PR.

PR #68, opened 2024-05-21
Add --version flag to aid in debugging.

wooly:
### Description

Self-explanatory from title.

### Context

https://linear.app/buildkite/issue/TAT-154/add-a-version-flag-to-the-splitter-to-aid-in-debugging

### Changes

NA

### Testing

`test-splitter -v or test-splitter --version`

PR #276, opened 2025-03-12
refactor: move `main.Version` to new internal package `version`

pda:
I found that with the `Version` global being in package `main`, it couldn't be reached by other packages.

I see that existing usage is either directly in `main` (arguably ready to be refactored into other packages), or threaded through via `api.ClientConfig{…, Version: Version}` and `authTransport{…,  version: …}`. I think this is a (rare?) example of a static build-time value that can just be compiled in and referenced as a global constant from anywhere in the codebase.

So, this PR moves the global `Version` string from package `main` to a new internal package `version`. This allows it to be reached from other packages, e.g. an upcoming `bktec upload` command, without having to thread it through from `main`.

It also initialises the string as `"dev"` instead of `""` so that non-release builds don't just have an empty string:

And, the `-version` flag now prints out `bktec {version}` rather than just `{version}`.

```
pda@pda-m3 ~/bk/bktec $ go build -o bktec && ./bktec --version
bktec dev
```

Related:
- #68

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -19,7 +19,7 @@ builds:
       - CGO_ENABLED=0
     goos: [linux, darwin]
     goarch: [amd64, arm64]
-    ldflags: "-X 'main.Version=v{{ .Version }}'"
+    ldflags: "-X 'github.com/buildkite/test-engine-client/internal/version.Version=v{{ .Version }}'"
     binary: bktec
 
 checksum:
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -15,6 +15,7 @@ import (
 
 	"github.com/buildkite/roko"
 	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/version"
 )
 
 // client is a client for the test plan API.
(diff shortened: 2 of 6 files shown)


.goreleaser.yaml
```diff
@@ -19,7 +19,7 @@ builds:
       - CGO_ENABLED=0
     goos: [linux, darwin]
     goarch: [amd64, arm64]
-    ldflags: "-X 'main.Version=v{{ .Version }}'"
+    ldflags: "-X 'version.Version=v{{ .Version }}'"
```
wooly, line 22: Oh I didn't think this would be possible if the version package was internal!

wooly (approved):

wooly:
Awesome, looking forward to seeing where this goes :+1:

.goreleaser.yaml
pda: I _think_ it works 😅

pda: Oh, it doesn't work the way I had it, but it _can_ work:

```
$ go build -o bktec -ldflags "-X version.Version=hello" && ./bktec --version
bktec dev
```

```
$ go build -o bktec -ldflags "-X github.com/buildkite/test-engine-client/internal/version.Version=hello" && ./bktec --version
bktec hello
```

nprizal (approved):
Nice!

PR #277, opened 2025-03-14
`env.Env` interface with `OS` (real) and `Map` (test) implementations

pda:
Introduce a new internal package `env`, with an `Env` interface abstracting `os.Getenv`, `os.LookupEnv` etc.

There's an `env.OS` implementation for real use, and an `env.Map` implementation for testing. This makes it easier for tests to inject fake environment, without affecting or being affected by the actual environment of the test process.

An upcoming `bktec upload` feature will be loading yet more environment variables from environment (detecting CI build environment etc); this PR is intended to make it easier to write tests in that PR.

~I chose to thread these through from `main()` to anything that depends on environment, which ended up a bit awkward purely because of `Config.DumpEnv()` which I think maybe could use a bit of thought anyway; I don't think it should need to read from environment at call time, it should probably read from the config values? Anyway, that's out of scope, but it explains the threading of `env` through several layers of method calls.~

There's a bit of stutter in the `env.Env` interface name, but I think it's acceptable; similar to `template.Template`, `context.Context` etc in stdlib.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -38,6 +38,9 @@ type Config struct {
 	Branch string
 	// JobRetryCount is the count of the number of times the job has been retried.
 	JobRetryCount int
+	// Env provides access to environment variables.
+	// It's public because many tests in other packages reference it (perhaps they should not).
+	Env env.Env
 	// errs is a map of environment variables name and the validation errors associated with them.
 	errs InvalidConfigError
 }
@@ -46,6 +49,7 @@ type Config struct {
 // It returns Config struct and an InvalidConfigError if there is an invalid configuration.
 func New(env env.Env) (Config, error) {
 	c := Config{
+		Env:  env,
 		errs: InvalidConfigError{},
 	}
 
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -1,5 +1,7 @@
 package config
 
+import "github.com/buildkite/test-engine-client/internal/env"
+
 // Config is the internal representation of the complete test engine client configuration.
 type Config struct {
 	// AccessToken is the access token for the API.
@@ -42,11 +44,13 @@ type Config struct {
 
 // New wraps the readFromEnv and validate functions to create a new Config struct.
 // It returns Config struct and an InvalidConfigError if there is an invalid configuration.
-func New() (Config, error) {
-	c := Config{errs: InvalidConfigError{}}
+func New(env env.Env) (Config, error) {
+	c := Config{
+		errs: InvalidConfigError{},
+	}
 
 	// TODO: remove error from readFromEnv and validate functions
-	_ = c.readFromEnv()
+	_ = c.readFromEnv(env)
 	_ = c.validate()
 
 	if len(c.errs) > 0 {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -27,7 +27,9 @@ func getExampleEnv() env.Env {
 }
 
 func TestNewConfig(t *testing.T) {
-	c, err := New(getExampleEnv())
+	env := getExampleEnv()
+
+	c, err := New(env)
 	if err != nil {
 		t.Errorf("config.New() error = %v", err)
 	}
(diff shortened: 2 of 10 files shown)


pda:
I considered storing `env` inside the `Config` struct, since `NewConfig(env)` now has `env`, but there's many existing tests that manually create `Config{}` structs without using `NewConfig`. I guess I could get them all to pass an empty env in to the struct 🤔 I decided against that _prior_ to discovering how much threading of env parameters was required to get it to the `DumpEnv(…)` call sites. Maybe I'll make this change and see how it looks.

pda:
I've done that now ☝🏼 and it _substantially_ reduces the delta to `main.go` code, which I think is good.

wooly (approved):
Thanks for this!

internal/config/config.go
```diff
@@ -36,17 +38,23 @@ type Config struct {
 	Branch string
 	// JobRetryCount is the count of the number of times the job has been retried.
 	JobRetryCount int
+	// Env provides access to environment variables.
+	// It's public because many tests in other packages reference it (perhaps they should not).
+	Env env.Env
```
zhming0, line 43: Hmm, this seems to be an abstraction leak? Does tests have to depends on this variable?

internal/config/config.go
```diff
@@ -36,17 +38,23 @@ type Config struct {
...
 
 // New wraps the readFromEnv and validate functions to create a new Config struct.
 // It returns Config struct and an InvalidConfigError if there is an invalid configuration.
-func New() (Config, error) {
-	c := Config{errs: InvalidConfigError{}}
+func New(env env.Env) (Config, error) {
+	c := Config{
+		Env:  env,
+		errs: InvalidConfigError{},
+	}
```
zhming0, line 54: It isn't quite clear to me whether this level of abstraction is needed for this project. 

This project currently use `Config` as the main injection point to set env during test, many test cases rely on explicitly setting `Config`, which doesn't seem to be troublesome. 

Using the new `env` parameter as the new injection point, we will probably have to refactor many test cases to make it obvious, otherwise I think most devs will continue to assume `Config` being the boundary. 

I can see benefits such as allowing parallel execution but I doubt if this is the goal? I probably missed some more contexts?

zhming0 (commented):
I left 2 questions, I think I might've missed some contexts. Keen to learn more 👀

internal/config/config.go
nprizal: The only package that references it during the test is the `main` package. Can we use `config.New` to create the config during the main package tests?

Instead of
```
cfg := config.Config{
    Env: env.Map{}
}
```
can we do
```
cfg := config.New(environment.Map{})
```

This way, we can make `config.Env` private?

internal/config/config.go
pda: I think it should _mostly_ be useful when testing the existing `main` implementation, and when testing the upcoming subcommands (`upload`, `bazel listen`, …) which also turn `(argv, envp)` into configuration.

Aside from that, a key thing with `bktec upload` is that a whole lot of common CI environment variables are checked for and, if present, mapped into `run_env`. For example `BUILDKITE_BUILD_ID` etc. These are (arguably) not really `bktec` _configuration_… `bktec` is probing the environment to find out the context that it's being executed within.

So for that reason, I think it's worthwhile to be able to swap in a fake env for testing.

internal/config/config.go
pda: We could, but I wanted to defer that to a future PR.

I'm not convinced `Config` needs `Env` at all, so I'd ideally like to eliminate it completely rather than make it private.

The only thing that's using it is `Config.DumpEnv()`, and I think we could make some changes, since I don't think it's doing a config-related job (e.g. reading `BUILDKITE_PARALLEL_JOB_COUNT`… that's not config, it's environment information)

https://github.com/buildkite/test-engine-client/blob/a1ce56d4532e2b4f1d86b050300d48e03770a712/internal/config/env.go#L38-L67

PR #281, opened 2025-03-19
Filter jest command with file paths on retries

bryankane:
### Description
The initial test run command for jest specifies which tests to run by specifying file paths, ensuring jest only needs to load and parse those specific files. However, when retrying failed tests, the filtering is done via a regex (passed into `--testNamePattern`). This forces jest to load and parse all test files to then filter down to the matching tests. On large repos this can be a very slow and inefficient process.

### Changes
This commit updates the retry command to accept an optional `{{testExamples}}` placeholder to also filter the retry command to only the files that contain the tests to be retried.

This also ensures that all file paths passed as arguments to jest have line number / columns removed, since jest doesn't support running individual tests by line number. This should only affect the retry case using `{{testExamples}}`, since line number locations aren't available until after the initial run

### Testing
Updated jest unit tests to ensure that `{{testExamples}}` placeholder gets updated if available, and continues working as expected when not in the command.

### DIFF
--- a/docs/jest.md
+++ b/docs/jest.md
@@ -61,5 +61,11 @@ export BUILDKITE_TEST_ENGINE_RETRY_CMD="yarn test --testNamePattern '{{testNameP
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=2
 ```
 
+To limit the number of files that Jest parses when retrying tests, you can also include a `{{testExamples}}` placeholder to specify the filenames of retried tests:
+```sh
+export BUILDKITE_TEST_ENGINE_RETRY_CMD="yarn test {{testExamples}} --testNamePattern '{{testNamePattern}}' --json --testLocationInResults --outputFile {{resultPath}}"
+export BUILDKITE_TEST_ENGINE_RETRY_COUNT=2
+```
+
 > [!IMPORTANT]
 > Make sure to append `--testNamePattern '{{testNamePattern}}' --json --testLocationInResults --outputFile {{resultPath}}` in your custom retry command.
--- a/internal/runner/jest_test.go
+++ b/internal/runner/jest_test.go
@@ -126,8 +126,8 @@ func TestJestRun_Retry_WithTestFilePattern(t *testing.T) {
 	})
 
 	testCases := []plan.TestCase{
-		{Scope: "expelliarmus", Name: "disarms the opponent", Path: "./testdata/jest/spells/expelliarmus.spec.js:1:1"},
-		{Scope: "lumos", Name: "illuminates the room", Path: "./testdata/jest/spells/lumos.spec.js:1:1"},
+		{Scope: "expelliarmus", Name: "disarms the opponent", Path: "./testdata/jest/spells/expelliarmus.spec.js"},
+		{Scope: "this will fail", Name: "for sure", Path: "./testdata/jest/failure.spec.js"},
 	}
 	result := NewRunResult([]plan.TestCase{})
 	err := jest.Run(result, testCases, true)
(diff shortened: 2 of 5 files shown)


niceking:
Hey @bryankane this change looks good! Could I please request these additional things:
- add a test case for test failures across multiple files 
- add documentation for this 

We've also changed some settings on our side to enable third party builds, and make the pipeline public for this repository, so you should be able to see your build results on future commits :)

bryankane:
@niceking Thanks for the review! I just pushed another commit that adds another test case and documentation.

Also, as mentioned in Slack, I updated the testPaths that are passed into the `jest` command to remove any line / column numbers, since jest doesn't support addressing individual tests that way. I think this would only affect the retry case, since we wouldn't have specific line numbers for tests when initially running them, but I figured it would be safest to include for both initial test runs and retries.

Let me know if this approach looks okay, and if there's any additional testing that you'd recommend!

internal/runner/jest_test.go
```diff
@@ -112,6 +112,35 @@ func TestJestRun_Retry(t *testing.T) {
...
+	result := NewRunResult([]plan.TestCase{})
+	err := jest.Run(result, testCases, true)
+
+	if err != nil {
+		t.Errorf("Jest.Run(%q) error = %v", testCases, err)
+	}
+
+	if result.Status() != RunStatusPassed {
+		t.Errorf("Jest.Run(%q) RunResult.Status = %v, want %v", testCases, result.Status(), RunStatusPassed)
+	}
```
zhming0, line 141: [blocking] We probably want to add an assertion to check test result count to equal `2` to prove that the file name filter works. 

(I presume this is the reason why you added a new test fixture?)

internal/runner/testdata/jest/spells/lumos.spec.js
zhming0, line 1: [nitpick] although I don't see harm, it doesn't seem to be necessary to have this extra spec fixture file? 

I am curious to see your thoughts 🙏🏿

zhming0 (commented):
🤩 Looks good to me overall, with only one potentially blocking comment.

internal/runner/testdata/jest/spells/lumos.spec.js
bryankane: Sorry for the delay here, and thanks for flagging this. I think I originally got confused when I tried to use the `failure.spec.js` fixture, and mixed up the jest test failure and the go test failure. I just swapped it back to the failure.spec.js reference in the test, and it works, so I'll delete this file!

bryankane:
Apologies for the delay here! I just pushed an update with some changes from the review (thanks for reviewing, @zhming0 🙌 )

I think it might make sense for me to wait for https://github.com/buildkite/test-engine-client/pull/283 to merge, and then I can rebase on that, but if this looks okay and you prefer to get this one in first, totally up to you!

Thanks again 😄

docs/jest.md
```diff
@@ -61,5 +61,11 @@ export BUILDKITE_TEST_ENGINE_RETRY_CMD="yarn test --testNamePattern '{{testNameP
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=2
 ```
 
+To limit the number of files that Jest parses when retrying tests, you can also include a `{{testExamples}}` placeholder to specify the filenames of retried tests:
+```sh
+export BUILDKITE_TEST_ENGINE_RETRY_CMD="yarn test {{testExamples}} --testNamePattern '{{testNamePattern}}' --json --testLocationInResults --outputFile {{resultPath}}"
+export BUILDKITE_TEST_ENGINE_RETRY_COUNT=2
+```
+
```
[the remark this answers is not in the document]
nprizal: We have an internal discussion about it, and agree that `testExamples` is a bit ambiguous. Changing this will require a _breaking change_, so we will stick with that placeholder for now.

nprizal (approved):
This looks good after removing the logic to remove the line code from `Path`.

zhming0 (approved):
@bryankane Thank you for contributing! Looks great.  🙇🏿

nprizal:
> I think after rebasing from your PR the line number logic should be removed -- let me know if i missed anything though!

Yes, it makes more sense to remove it. I don't think it's missing anything. I'll merge the PR and will keep you updated when this fix is released. Thank you for your contribution 😊

PR #282, opened 2025-03-19
Bazel Build Event Protocol (BEP) → `bktec upload`

pda:
_Draft, based on #278 which is based on #277_

Listen to Build Event Protocol (BEP) / Build Event Service (BES) and upload the XML test result files to Test Engine.

Related:
- https://bazel.build/remote/bep
- #278


.gitignore
```diff
@@ -7,6 +7,7 @@
 *.dll
 *.so
 *.dylib
+/bktec
```
nprizal, line 10: 👍

main.go
```diff
@@ -69,6 +71,11 @@ func main() {
 			logErrorAndExit(16, "upload: %v", err)
 		}
 		os.Exit(0)
+	} else if flag.Arg(0) == "bazel" && flag.Arg(1) == "listen" {
```
nprizal, line 74: [nit] I reckon `switch` will be more readable now we are most likely going to have many sub-commands.

pda: Would that be a nested switch?

```go
switch flag.Arg(0) {
case "upload":
  …
case "bazel":
  switch {
  case "listen":
    …
  default:
    …
  }
default:
  …
}
```

I figured it's simpler to use a flatter `if` _for the moment_, and think about whether we want to adopt a bells-and-whistles subcommand/flags package or use some nested `switch` etc.

PR #283, opened 2025-03-21
Remove line and column from Jest test result's path

nprizal:
### Description

@bryankane discovered a bug when a Jest test failed due to a timeout. After retrying, the test eventually passed, but the run result was still marked as failed. This issue happens because we use test name, file path, and location (line and column number) to identify a test.  When a Jest test times out, the location is `null`. However, when the test is retried and passes, the location is set correctly by Jest. Since the test location is `null` during the initial run but set correctly during the retry, these executions are treated as separate tests, causing the final result to be marked as failed.

This PR removes line and column number from Jest's test path, so we can correctly associate the tests that time out (i.e. missing location). This aligns with the approach on https://github.com/buildkite/test-engine-client/pull/281 that removes the line and column number when passing the path to retry failed tests. 

> [!IMPORTANT]
> The implication of this fix is that distinct tests with the same name in the same file will be treated as the same test. This practice is not recommended by Test Engine, and should be handled differently (e.g., by displaying warning when there are multiple distinct tests with same name). However, this is beyond the scope of this PR.

### DIFF
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -128,7 +128,7 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 			testCase := plan.TestCase{
 				Name:  example.Title,
 				Scope: strings.Join(example.AncestorTitles, " "),
-				Path:  fmt.Sprintf("%s:%d:%d", testPath, example.Location.Line, example.Location.Column),
+				Path:  testPath,
 			}
 
 			result.RecordTestResult(testCase, status)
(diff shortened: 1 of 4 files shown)


internal/runner/jest.go
```diff
@@ -128,7 +128,7 @@ func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) erro
 			testCase := plan.TestCase{
 				Name:  example.Title,
 				Scope: strings.Join(example.AncestorTitles, " "),
-				Path:  fmt.Sprintf("%s:%d:%d", testPath, example.Location.Line, example.Location.Column),
+				Path:  testPath,
```
zhming0, line 131: 👀 I have a bit concern over this causing mismatch between JS Test Collector and bktec, I might need to pick your brain on this topic 🙏🏿

zhming0: All good, I understand now `Path` is only used internally by bktec.

zhming0 (approved):
Looks good.

PR #285, opened 2025-03-24
Implement basic `pytest` splitting support

nprizal:
### Description

Add basic splitting support for `pytest`. It doesn't currently read/parse the test output, so retry and other features is not supported.

Missing features:
- Retry
- Skip test
- Mute test


zhming0 (approved):
Looks good to me.

PR #286, opened 2025-03-25
Read test collector output after running pytest

nprizal:
### Description

`bktec` doesn't currently interpret the output/result of a `pytest` run. To support test retry for `pytest`, `bktec` needs to interpret the result. `pytest` can generate native JUnit XML output, but it lacks some information that `bktec` needs. Because `bktec` relies on [Python Test Collector](https://github.com/buildkite/test-collector-python) (to use timing data for intelligent partitioning), we can read the output of Python Test Collector.

The changes allow bktec to correctly report the test result, retry failed tests, and handle muted tests.
Screenshot 2025-03-26 at 12 11 14 PM
Screenshot 2025-03-26 at 12 13 00 PM

### Changes

- Change the default `TestCommand` for pytest to `pytest --json={{resultPath}}`
- Add `TestEngineTest` struct that represents a Test Engine [test result object](https://buildkite.com/docs/test-engine/importing-json#json-test-results-data-reference-test-result-objects), and a function to parse test collector output into `[]TestEngineTest`
- Update `Pytest.Run` function to read test collector output at the end of the run


internal/runner/test_result.go
```diff
@@ -8,6 +8,7 @@ const (
 	TestStatusPassed  TestStatus = "passed"
 	TestStatusFailed  TestStatus = "failed"
 	TestStatusSkipped TestStatus = "skipped"
+	TestStatusUnknown TestStatus = "unknown"
```
zhming0, line 11: what is the context of this change? 🤔

internal/runner/pytest.go
```diff
@@ -46,7 +47,33 @@ func (p Pytest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
...
 	err = runAndForwardSignal(cmd)
 
-	return err
+	// Only rescue exit code 1 because it indicates a test failures.
+	// Ref: https://docs.pytest.org/en/7.1.x/reference/exit-codes.html
+	if exitError := new(exec.ExitError); errors.As(err, &exitError) && exitError.ExitCode() != 1 {
+		return err
+	}
+
+	tests, parseErr := ParseTestEngineJson(p.ResultPath)
```
zhming0, line 56: I wonder if it make sense forcing user to supply a result path, because result path is only used by bktec. We could just create tmp file by ourselves on bktec side.

But if this practice is already used by other runner, then I see no blocker

zhming0 (approved):
I have a few questions, looks great!

internal/runner/test_engine.go
```diff
@@ -0,0 +1,33 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to read json: %v", err)
+	}
+
+	if err := json.Unmarshal(data, &results); err != nil {
+		return nil, fmt.Errorf("failed to parse json: %v", err)
+	}
+
+	return results, nil
+}
```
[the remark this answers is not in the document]
nprizal: Yes, it's only used by `pytest` at the moment. I put this in the `test_engine.go` because the JSON structure is not unique to Python collector (i.e. its used by almost all collectors, and manual JSON upload). I can move this to `pytest.go` and we can extract it if it's used by other runner. I don't have strong opinion.

internal/runner/test_result.go
nprizal: Ah yes, this change is not necessary, actually. TE json structure allows `unknown` as test result, but we don't handle this in `bktec`. I will remove this change.

internal/runner/pytest.go
nprizal: Discussed internally, we will make result path optional for pytest on separate PR

zhming0 (approved):

PR #288, opened 2025-03-26
Allow customization of retry test command for pytest

nprizal:
### Description

https://github.com/buildkite/test-engine-client/pull/286 introduced changes to support retry test for pytest. This PR allows customers to customize the command used when retrying failed tests.

### DIFF
--- a/internal/runner/pytest.go
+++ b/internal/runner/pytest.go
@@ -30,6 +30,10 @@ func NewPytest(c RunnerConfig) Pytest {
 		c.TestFilePattern = "**/{*_test,test_*}.py"
 	}
 
+	if c.RetryTestCommand == "" {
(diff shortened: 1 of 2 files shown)


internal/runner/pytest.go
zhming0, line 1: The change looks good me, but can I pick your brain on the use case of customizable retry test command? 

What use cases are this designed for? For pytest, it seems the vanilla TestCommand would suffice?

nprizal: I'm not that familiar with pytest testing practices, so I can't think of a use case for customising the retry test command at the moment. However, bktec supports customising the retry command for RSpec, Jest, and Playwright, so I think it’s good to keep the behaviour consistent across runners.

zhming0 (approved):

PR #289, opened 2025-03-27
PIE-3521: streamline bktec dev env, make ResultPath optional for pytest

zhming0:
### Description

* Fix test setup for pytest to test the integration with buildkite-test-collector-python.
* Fix bin/setup so it automatically install pact. 
* Change pytest so ResultPath is optional. 

### Context

PIE-3521


bin/setup
```diff
@@ -1,5 +1,26 @@
 #!/usr/bin/env bash
 
+# install pact-go as a dev dependency
+go get github.com/pact-foundation/pact-go/v2
```
nprizal, line 4: yass!! Thanks so much for this.

internal/runner/detector.go
```diff
@@ -13,7 +13,9 @@ type RunnerConfig struct {
 	TestFilePattern        string
 	TestFileExcludePattern string
 	RetryTestCommand       string
-	ResultPath             string
+	// ResultPath is used internally so bktec can read result from Test Runner.
+	// User typically don't need to worry about setting this except in Rspec.
+	ResultPath string
```
nprizal, line 18: 👍

internal/runner/detector.go
```diff
@@ -13,7 +13,9 @@ type RunnerConfig struct {
 	TestFilePattern        string
 	TestFileExcludePattern string
 	RetryTestCommand       string
-	ResultPath             string
+	// ResultPath is used internally so bktec can read result from Test Runner.
+	// User typically don't need to worry about setting this except in Rspec.
```
nprizal, line 17:
```suggestion
	// User typically don't need to worry about setting this except in Rspec, and Playwright.
```
Playwright result needs to be configured in `config.js`, and bktec doesn't have a way to know that, so its' mandatory for Playwright.

zhming0: Thanks! fixed 👍🏿 . I wonder if we can just automatically mutate `config.js` on runtime 🤔

internal/runner/pytest.go
```diff
@@ -156,3 +163,19 @@ func ParsePytestCollectorResult(path string) ([]TestEngineTest, error) {
...
 }
+
+func getRandomTempFilename() string {
+	max := int64(1e10)
+	n, err := rand.Int(rand.Reader, big.NewInt(max))
+	if err != nil {
+		panic(err)
+	}
+	return filepath.Join(os.TempDir(), fmt.Sprintf("pytest-results-%d.json", n))
+}
```
nprizal, line 174: not blocking, but alternatively we can use [os.MkdirTemp](https://pkg.go.dev/os#MkdirTemp) with `*` pattern to create unique directory so we don't have to create unique identifier ourselves.
```suggestion
func getRandomTempFilename() string {
	tempDir, err := os.MkdirTemp("", "pytest-collector-*")
	if err != nil {
		panic(err)
	}
	return filepath.Join(tempDir, "result.json")
}
```

nprizal (commented):
We also need to change the validation for `ResultPath` in the config.

internal/runner/pytest.go
zhming0: Good idea 👍🏿

nprizal (approved):
Nice! 🚀
Not related to this PR, but because we now check if collector is installed, we probably need to change the default test command. Currently, the default test command has `--json={{resultPath}}` which will fails if the test collector is not installed.

zhming0:
@nprizal yeah, I agree. [PIE-3537](https://linear.app/buildkite/issue/PIE-3537/bktec-should-check-if-test-collector-python-installed) should handle that.

PR #291, opened 2025-03-27
PIE-3537: exit bktec for pytest without buildkite-test-collector

zhming0:
### Description

* Replace the simple `pip` checker with a more reliable `python` checker. 
* Exit bktec when pytest is used without our collector. 

### Context

PIE-3537

### DIFF
--- a/internal/runner/pytest.go
+++ b/internal/runner/pytest.go
@@ -23,6 +23,12 @@ func (p Pytest) Name() string {
 }
 
 func NewPytest(c RunnerConfig) Pytest {
+	if !checkPythonPackageInstalled("buildkite_test_collector") { // python import only use underscore
+		fmt.Fprintln(os.Stderr, "Error: Required Python package 'buildkite-test-collector' is not installed.")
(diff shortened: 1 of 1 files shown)


internal/runner/pytest.go
```diff
@@ -23,6 +23,12 @@ func (p Pytest) Name() string {
 }
 
 func NewPytest(c RunnerConfig) Pytest {
+	if !checkPythonPackageInstalled("buildkite_test_collector") { // python import only use underscore
+		fmt.Fprintf(os.Stderr, "Error: Required Python package 'buildkite-test-collector' is not installed.\n")
+		fmt.Fprintf(os.Stderr, "Please install it with: pip install buildkite-test-collector\n")
```
nprizal, line 28: [nit]
```suggestion
		fmt.Fprintln(os.Stderr, "Error: Required Python package 'buildkite-test-collector' is not installed.")
		fmt.Fprintln(os.Stderr, "Please install it with: 'pip install buildkite-test-collector'.")
```

nprizal (approved):

PR #301, opened 2025-04-23
TE-3690: go test x bktec

zhming0:
### Description

Introduce an experimental Go test support. 

Go test, different from other test runners in that it does not really have a concept of "file", so I tried I might to use "package" as a replacement. In another word, the "Path" for a test case in golang, is "package" instead of file path. 

This means: 

- Smart test splitting will never work (because server will never have file stats for "package"). I expect server-side to return non-intelligent simple test splitting test plan.
- Auto test retry happen on package level. 

### Context

part of TE-3690

subsequent PRs incoming for Readme change.

### Changes

* Added `gotest` runner.
* Added many codedoc clarifying behavior of bktec internal.

### DIFF
--- a/internal/plan/type.go
+++ b/internal/plan/type.go
@@ -7,7 +7,8 @@ const (
 	TestCaseFormatExample TestCaseFormat = "example"
 )
 
-// TestCase represents a single test case.
+// TestCase currently can represent a single test case or a single test file (when used as output of test plan API).
+// TODO: it's best if we split this into two types.
 type TestCase struct {
 	EstimatedDuration int            `json:"estimated_duration,omitempty"`
 	Format            TestCaseFormat `json:"format,omitempty"`
@@ -18,14 +19,18 @@ type TestCase struct {
 	// In RSpec, the path can be a test file like `user_spec.rb` or an individual test id like `user_spec.rb[1,2]`.
 	// In Jest, the path is a test file like `src/components/Button.spec.tsx`.
 	// In pytest, the path can be a test file like `test_hello.py` or a node id like `test_hello.py::TestHello::test_greet`
+	// In go test, the path can only be package name like "example.com/foo/bar".
 	Path  string `json:"path"`
 	Scope string `json:"scope,omitempty"`
 }
 
 // Task represents the task for the given node.
 type Task struct {
-	NodeNumber int        `json:"node_number"`
-	Tests      []TestCase `json:"tests"`
+	NodeNumber int `json:"node_number"`
+	// When splitting by file, this tests array is essentially an array of test files.
+	// When splitting by example, this array is an array of proper test cases.
+	// See comment above, we plan to split TestCase into two types or clarify its usage.
+	Tests []TestCase `json:"tests"`
 }
 
 // TestPlan represents the entire test plan.
--- a/internal/runner/gotest.go
+++ b/internal/runner/gotest.go
@@ -1,0 +1,153 @@
+package runner
+
+import (
+	"errors"
+	"fmt"
+	"os/exec"
+	"strings"
+
+	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/plan"
+	"github.com/kballard/go-shellquote"
+)
+
+type GoTest struct {
+	RunnerConfig
+}
+
+// Compile-time check that GoTest implements TestRunner
+var _ TestRunner = (*GoTest)(nil)
+
+func NewGoTest(c RunnerConfig) GoTest {
+	if c.TestCommand == "" {
+		c.TestCommand = "gotestsum --junitfile={{resultPath}} {{packages}}"
+	}
+
+	if c.RetryTestCommand == "" {
+		c.RetryTestCommand = c.TestCommand
+	}
+
+	return GoTest{
+		RunnerConfig: c,
+	}
+}
+
+func (g GoTest) Name() string {
+	return "gotest"
+}
+
+func (g GoTest) GetExamples(files []string) ([]plan.TestCase, error) {
+	return nil, fmt.Errorf("not supported in go test")
+}
+
+// Run executes the configured command for the specified packages.
+func (g GoTest) Run(result *RunResult, testCases []plan.TestCase, retry bool) error {
+	cmdName, cmdArgs, err := g.commandNameAndArgs(g.TestCommand, testCases)
+	if err != nil {
+		return fmt.Errorf("failed to build command: %w", err)
+	}
+
+	cmd := exec.Command(cmdName, cmdArgs...)
+	err = runAndForwardSignal(cmd)
+	// go test output does not differentiate build fail or test fail. They both return 1
+	// What is even more bizarre is that even when go test failed on compliation, it will still generate an output xml
+	// file that says "TestMain" failed..
+	if exitError := new(exec.ExitError); errors.As(err, &exitError) && exitError.ExitCode() != 1 {
+		return err
+	}
+
+	testResults, err := loadAndParseGotestJUnitXmlResult(g.ResultPath)
+
+	if err != nil {
+		return fmt.Errorf("failed to load and parse test result: %w", err)
+	}
+
+	for _, test := range testResults {
+		result.RecordTestResult(plan.TestCase{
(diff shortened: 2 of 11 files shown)


wooly:
Thanks for working on this @zhming0!

> Smart test splitting will never work (because server will never have file stats for "package"). I expect server-side to return dumb test plan.

Does this mean that the timing data aggregated to the package level would never be used to do a pseudo-file-level test split? I'm unsure of what you mean when you say 'dump test plan'. For other frameworks, we have fallbacks in place at the moment to run a deterministic split if Buildkite is unavailable. Are these the sorts of plans that we would expect to see from the Gotest implementation?

zhming0:
@wooly those a re good questions. We don't really have timing data aggregated on package level, as you know "package" is not a field on our data store. 

In this case though, technically we can do a group by "scope" to simulate package timing analysis for Golang only, but I think that will be further down. 

re: dumb test plan

I rephrased that statement, "dumb" is the wrong word 😅 . What will happen is that: bktec will send a bunch of "package" names to test plan API, test plan API would have no timing data, so AFAIK it would split using simple deterministic heuristic. If it doesn't work, like other test runner, bktec fallback to client side deterministic splitting.

nprizal:
> Go test, different from other test runners in that it does not really have a concept of "file", so I tried I might to use "package" as a replacement. In another word, the "Path" for a test case in golang, is "package" instead of file path.
> 
> This means:
> - Smart test splitting will never work (because server will never have file stats for "package"). I expect server-side to return non-intelligent simple test splitting test plan.

💯. This is the approach that we discussed a while ago. Having a non-intelligent splitting is a good start. The next step will be to implement the feature in the server to support _package_ or _scope_ level splitting.

> Auto test retry happen on package level.

While this is fine for now, I believe we can support individual test retry for go test. Something like this will work I suppose,
```
go test ^TestFailedTest$ packageName
```

internal/plan/type.go
```diff
@@ -7,7 +7,8 @@ const (
 	TestCaseFormatExample TestCaseFormat = "example"
 )
 
-// TestCase represents a single test case.
+// Currently TestCase can represent a single test case or a single test file (when used as output of test plan API).
+// TODO: it's best if we split this into two types.
```
nprizal, line 11: 👍 I like the idea of having a separate type for file, individual test, and package.

main.go
```diff
@@ -40,8 +40,13 @@ func printStartUpMessage() {
 }
 
 type TestRunner interface {
+	// Take testCases as input, mutate runner.RunResult -> adding test results to it.
```
nprizal, line 43: [nit] Most go doc starts with the name of the function/attributes that it explains.
```suggestion
	// Run takes testCases as input, executes the test against the test cases, and updates the runner.RunResult with the test results.
```

main.go
```diff
@@ -40,8 +40,13 @@ func printStartUpMessage() {
...
 	Run(result *runner.RunResult, testCases []plan.TestCase, retry bool) error
+	// This function is only used for split by example use case. Currently only supported by RSpec.
```
nprizal, line 45:
```suggestion
	// GetExamples discovers all tests within a given files. This function is only used for split by example use case. Currently only supported by RSpec.
```

main.go
```diff
@@ -40,8 +40,13 @@ func printStartUpMessage() {
...
 	GetExamples(files []string) ([]plan.TestCase, error)
+	// Discover all test files that the runner should execute.
+	// This is sent to server-side when creating test plan.
+	// This is also used to obtain a fallback non-intelligent test splitting mechanism.
```
nprizal, line 49: [nit]
```suggestion
	// GetFiles discovers all test files that the runner should execute.
	// This is sent to server-side when creating test plan.
	// This is also used to obtain a fallback non-intelligent test splitting mechanism.
```

internal/plan/type.go
```diff
@@ -7,7 +7,8 @@ const (
 	TestCaseFormatExample TestCaseFormat = "example"
 )
 
-// TestCase represents a single test case.
+// Currently TestCase can represent a single test case or a single test file (when used as output of test plan API).
```
nprizal, line 10: [nit]
```suggestion
// TestCase currently represents a single test case or a single test file (when used as output of test plan API).
```

wooly:
> Having a non-intelligent splitting is a good start.

I'm still not clear on what the split will actually be. I'd argue that producing a test plan that yields completely cooked splits is potentially worse than not supporting Go at all.

internal/runner/gotest.go
```diff
@@ -0,0 +1,153 @@
...
+
+	return nil // Success
+}
+
+// GetFiles discovers Go packages using `go list ./...`.
+// Note that "file" does not exist as a first level concept in Golang projects
+// So this func is returning a list of packages instead of files.
+// The implication is that the Server-side smart test splitting will never work.
+// It almost will always fallback to simple splitting.
+func (g GoTest) GetFiles() ([]string, error) {
```
nprizal, line 83: Thanks for adding the description for this function. I wonder if we can implement a `GetPackages` that only work for `gotest` runner (or perhaps other runners too in the future). This isn't blocking the PR tho.

internal/runner/gotest.go
```diff
@@ -0,0 +1,153 @@
...
+	cmdName, cmdArgs, err := g.commandNameAndArgs(g.TestCommand, testCases)
+	if err != nil {
+		return fmt.Errorf("failed to build command: %w", err)
+	}
+
+	cmd := exec.Command(cmdName, cmdArgs...)
+	err = runAndForwardSignal(cmd)
+	// go test output does not differentiate build fail or test fail. They both return 1
+	// What is even more bizarre is that even when go test failed on compliation, it will still generate an output xml
+	// file that says "TestMain" failed..
```
nprizal, line 54: 🤯 So what would happen if there is a compilation error / error outside test failure?

nprizal (commented):
Thanks for adding the doc for clarity. The PR looks great! I have a couple of nitpicks for the docs, which you can take or leave, and also a question about the error handling.

nprizal:
> I'm still not clear on what the split will actually be. I'd argue that producing a test plan that yields completely cooked splits is potentially worse than not supporting Go at all.

@wooly The test planner will distribute the packages into multiple jobs without taking the duration for that package into consideration. I'd argue supporting the non-intelligent splitting adds value. You can of course achieve this same result by writing a bash script to distribute the packages into **n** number of jobs, but it can be a bit messy. Also, supporting a go test in bktec will open possibilities to mute a test using test state management.

internal/runner/gotest.go
```diff
@@ -0,0 +1,153 @@
...
+	return args[0], args[1:], nil
+}
+
+// Pluck unique packages from test cases
+func (g GoTest) getPackages(testCases []plan.TestCase) ([]string, error) {
+	packages := make([]string, 0, len(testCases))
+
+	packagesSeen := map[string]bool{}
+	for _, tc := range testCases {
+		packageName := tc.Path
```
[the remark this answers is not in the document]
zhming0: That was my first intuition. But the problem is that the type `plan.TestCase` has been conflated. In some cases, bktec will call `GetFiles` first and then construct TestCase with only `Path` but no Scope and Name. As an end result, prior to test result parsing, the `Path` field is the only reliable field.

internal/runner/gotest.go
zhming0: Yeah, I agree in principle. We will need to carefully examine if we can generalize golang's package-oriented design to other test runners. If we decide to go down this route, this will be big undertaking, currently "package" is not a field on any of our backend store.

internal/runner/gotest.go
zhming0: It will collect a single failed execution: TestMain. Run status will be failed.  I have a [test case for that](https://github.com/buildkite/test-engine-client/blob/7f0a52f5ffe066eaf62b6ebf3d3479f763f69549/internal/runner/gotest_test.go#L57). 

We might have other failing scenarios. It will probably take some real world use case to find out.

zhming0:
> yields completely cooked splits

I _think_ we use the test plan API per its contract. In the local test, test plan api seems to be acting per its spec?

nprizal (approved):
Looks good to me

PR #303, opened 2025-04-28
TE-3715: fix missing token mention in bktec gotest doc

zhming0:
### Description

We forgot to mention the need of `BUILDKITE_ANALYTIC_TOKEN`

### Context

TE-3715


jameshill (approved):
One day we'll have one API token, today isn't that day, and that's ok for now.

PR #305, opened 2025-04-29
TE-3707 follow up: stop collecting test fixture test cases

zhming0:
### Description

one funny thing with using bktec on bktec is that the BUILDKITE_ANALYTICS_TOKEN will be used by the test code too.
So we end up having our test cases + those test fixture test execution collected in test engine.

This PR stops collecting test fixture test cases.


pda (approved):

nprizal (commented):
Nice!

PR #307, opened 2025-04-30
TE-3708-2: add windows build to CI and fix one more Windows issue

zhming0:
### Description

self-explanatory.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -61,6 +61,7 @@ steps:
           os:
             - darwin
             - linux
+            - windows
           arch:
             - amd64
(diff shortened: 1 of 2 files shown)


nprizal (commented):
I don't think printing the signal name (i.e. using `unix.SignalName`)  in the log is a hard requirement. Maybe we can remove it, so we don't have to handle the function separately for unix and for windows. What do you think?

zhming0:
@nprizal yeah, fair point. I dislike over-using the build flag thing too. 

Another look please?

nprizal (approved):
Much cleaner ^^

PR #312, opened 2025-05-14
TE-3818: retry on failed muted tests by default

zhming0:
### Description

As we make more progress on statistic based flaky monitor, we discovered a problem: if a test is muted, it won't be retried. It means a sudden signal cutoff for some of our flaky detection models. 

We want rich signal. 

This PR: 
* make retry for failed muted test by default. 
* offer a new env var `BUILDKITE_TEST_ENGINE_DISABLE_RETRY_FOR_MUTED_TEST` as opt-out mechanism. 

### Context

TE-3818

I recommend ignoring whitespace when reviewing this change.

### DIFF
--- a/internal/config/read.go
+++ b/internal/config/read.go
@@ -53,6 +53,7 @@ func (c *Config) readFromEnv(env env.Env) error {
 	c.TestFilePattern = env.Get("BUILDKITE_TEST_ENGINE_TEST_FILE_PATTERN")
 	c.TestFileExcludePattern = env.Get("BUILDKITE_TEST_ENGINE_TEST_FILE_EXCLUDE_PATTERN")
 	c.TestRunner = env.Get("BUILDKITE_TEST_ENGINE_TEST_RUNNER")
+	c.RetryForMutedTest = strings.ToLower(env.Get("BUILDKITE_TEST_ENGINE_DISABLE_RETRY_FOR_MUTED_TEST")) != "true"
 	c.ResultPath = env.Get("BUILDKITE_TEST_ENGINE_RESULT_PATH")
 
 	c.SplitByExample = strings.ToLower(env.Get("BUILDKITE_TEST_ENGINE_SPLIT_BY_EXAMPLE")) == "true"
--- a/internal/runner/run_result_test.go
+++ b/internal/runner/run_result_test.go
@@ -84,7 +84,7 @@ func TestFailedTests(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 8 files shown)


internal/config/read.go
```diff
@@ -53,6 +53,7 @@ func (c *Config) readFromEnv(env env.Env) error {
 	c.TestFilePattern = env.Get("BUILDKITE_TEST_ENGINE_TEST_FILE_PATTERN")
 	c.TestFileExcludePattern = env.Get("BUILDKITE_TEST_ENGINE_TEST_FILE_EXCLUDE_PATTERN")
 	c.TestRunner = env.Get("BUILDKITE_TEST_ENGINE_TEST_RUNNER")
+	c.RetryForMutedTest = strings.ToLower(env.Get("BUILDKITE_TEST_ENGINE_DISABLE_RETRY_FOR_MUTED_TEST")) != "true"
```
KatieWright26, line 56: How come you need the `!=` logic here instead of `= "false"`?

zhming0: If I use `== "false"` check, I will essentially force user to set: `BUILDKITE_TEST_ENGINE_DISABLE_RETRY_FOR_MUTED_TEST`  to be `false` in order to use this feature.

Then it turns a opt-out feature flag, into a opt-in feature flag 😅

nprizal (commented):
The changes looks good to me. One feedback maybe to add a test for `run_result` to make sure retrying the muted test doesn't affect the run result.

Also not related to this PR, maybe we need to start thinking about yaml config :D

KatieWright26 (approved):
I don't know Go but nothing stands out to me. Happy to approve to unblock you

internal/runner/run_result_test.go
```diff
@@ -121,16 +121,25 @@ func TestFailedTests_ExcludeMutedTests(t *testing.T) {
...
 	})
-	r.RecordTestResult(apple1, TestStatusFailed)
+	r.RecordTestResult(apple1, TestStatusFailed) // Muted so no impact
+	r.RecordTestResult(apple1, TestStatusFailed) // Retry fails, but muted so no impact
 	r.RecordTestResult(apple2, TestStatusPassed)
-	r.RecordTestResult(banana, TestStatusFailed)
 
-	failedTests := r.FailedTests()
+	// At this point the run status is "passsed"
+	if r.Status() != RunStatusPassed {
```
nprizal, line 129: 👍 nice!

PR #333, opened 2025-06-19
Fix missing scope

wooly:
### Description

This should have been included in the original Cucumber PR but my push failed.


tommeier (approved):
aha - perfect 👏

PR #359, opened 2025-10-07
[TE-4765] Add `-files` flag to override default test discovery

nprizal:


nprizal:
@gchan, the previous implementation (https://github.com/buildkite/test-engine-client/pull/33) accepted test files as comma-separated string command line arguments. However, since the list of test files could be large, and to anticipate support for individual test discovery which may need a structured format, I decided to pass it as a file.

gchan:
> @gchan, the previous implementation (#33) accepted test files as comma-separated string command line arguments. However, since the list of test files could be large, and to anticipate support for individual test discovery which may need a structured format, I decided to pass it as a file.

Accepting a file as input seems like a good approach. Space separated file names seems better than comma-separated. I assume if file names have spaces, they would be escaped? Perhaps we should add a test for this?

PR #364, opened 2025-10-14
Add .tool-versions file for asdf / mise

malclocke:


.tool-versions
```diff
@@ -0,0 +1,3 @@
+golang 1.22.4
+nodejs 24.9.0
+python 3.13.2
```
malclocke: The selection of versions here is totally unscientific, just what I have installed for other projects. Happy to update with different versions.

gchan (approved):

PR #367, opened 2025-10-16
Local test fixes

malclocke:
This PR includes some changes to make running the full test suite locally possible from a fresh install.

- Fix a typo in `internal/runner/pytest_test.go`
- Add a `bin/test` script which mimics the setup on CI when run locally. Without the python virtualenv activated and the `CI` env var set the python runner tests do not work.

### DIFF
--- a/bin/test
+++ b/bin/test
@@ -1,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# Required for internal/runner python tests
+source .venv/bin/activate
+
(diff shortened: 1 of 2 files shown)


bin/test
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# Required for internal/runner python tests
+source .venv/bin/activate
```
nprizal: 👍

PR #369, opened 2025-10-16
Add `run` command.

malclocke:
Allows `bktec` to be run as `bktec run`. This is the same behaviour as when no command argument is given, so the following are equiavalent:

```
bktec
bktec run
```

```
$ bktec help
NAME:
   bktec - Buildkite Test Engine Client

USAGE:
   bktec [global options] [command [command options]]

COMMANDS:
   run      Run tests (default)
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --files string  override the default test file discovery by providing a path to a file containing a list of test files (one per line)
   --version       print version information and exit (default: false)
   --help, -h      show help
```


main.go
```diff
@@ -54,19 +54,45 @@ type TestRunner interface {
...
 
-	versionFlag := flag.Bool("version", false, "print version information")
-	filesFlag := flag.String("files", "", "override the default test file discovery by providing a path to a file containing a list of test files (one per line)")
+func printVersion(ctx context.Context, cmd *cli.Command) error {
+	fmt.Printf("bktec %s\n", version.Version)
+	return nil
+}
 
-	flag.Parse()
+func run(ctx context.Context, cmd *cli.Command) error {
```
nprizal: [nit] maybe its time we need to extract this and its related functions to its own file? one file per action?

main.go
```diff
@@ -96,7 +122,6 @@ func main() {
 	}
 
 	// get plan
-	ctx := context.Background()
```
gchan: What's the reason behind removing this background context?

main.go
```diff
@@ -54,19 +54,45 @@ type TestRunner interface {
...
+		},
+	}
 
-	debug.SetDebug(env.Get("BUILDKITE_TEST_ENGINE_DEBUG_ENABLED") == "true")
+	if err := cmd.Run(context.Background(), os.Args); err != nil {
+		logErrorAndExit(16, "Error: %v", err)
+	}
+}
 
-	versionFlag := flag.Bool("version", false, "print version information")
```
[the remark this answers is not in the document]
malclocke: @nprizal I've reworked things so `--version` is retained.

nprizal (approved):

PR #371, opened 2025-10-17
Te 4789 move run command to package

malclocke:
This change is best reviewed by commit.

The first commit moves most of the content of `main.go` to `command/run.go`. With this change `command` needs to export 2 funcs that are still required by `main()`, `Run()` and `LogErrorAndExit()`.

The second commit modifies the code that was moved from `main` to `command` so that it no longer needs to reference `LogErrorAndExit`. All the previous calls to `LogErrorAndExit` in the `command` package are modified to `return err` and `command.LogErrorAndExit` is moved back to being an unexported func in `main`.

### DIFF
--- a/internal/runner/detector.go
+++ b/internal/runner/detector.go
@@ -1,7 +1,7 @@
 package runner
 
 import (
-	"errors"
+	"fmt"
 
 	"github.com/buildkite/test-engine-client/internal/config"
 	"github.com/buildkite/test-engine-client/internal/plan"
@@ -36,7 +36,7 @@ func DetectRunner(cfg config.Config) (TestRunner, error) {
 		ResultPath:             cfg.ResultPath,
 	}
 
-	switch cfg.TestRunner {
+	switch testRunner := cfg.TestRunner; testRunner {
 	case "rspec":
 		return NewRspec(runnerConfig), nil
 	case "jest":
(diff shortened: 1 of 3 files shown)


internal/runner/detector.go
```diff
@@ -55,6 +55,6 @@ func DetectRunner(cfg config.Config) (TestRunner, error) {
 		return NewCucumber(runnerConfig), nil
 	default:
 		// Update the error message to include the new runner
-		return nil, errors.New("runner value is invalid, possible values are 'rspec', 'jest', 'cypress', 'playwright', 'pytest', 'pytest-pants', 'gotest', or 'cucumber'")
+		return nil, fmt.Errorf("runner value %q is invalid, possible values are 'rspec', 'jest', 'cypress', 'playwright', 'pytest', 'pytest-pants', 'gotest', or 'cucumber'", testRunner)
```
malclocke: Previously the caller was prepending the name of the invalid runner to the error message https://github.com/buildkite/test-engine-client/blob/45db707f44cb53ef92da660edcc41bf3fb155f0c/main.go#L113

With this change the caller error wrapping is simplified.
```go
return fmt.Errorf("Unsupported value for BUILDKITE_TEST_ENGINE_TEST_RUNNER: %v", err)
```

main.go
```diff
@@ -95,467 +56,15 @@ func printVersion(ctx context.Context, cmd *cli.Command, versionFlag bool) error
...
-	}
-
-	// The SplitByExample flag indicates whether to filter slow files for splitting by example.
-	// Regardless of the flag's state, the API will still filter other files that need to be split by example, such as those containing skipped tests.
-	// Therefore, we must filter and split files even when SplitByExample is disabled.
-	testParams, err := filterAndSplitFiles(ctx, cfg, client, testFiles, runner)
-	if err != nil {
-		return api.TestPlanParams{}, err
+		// otherwise exit code 16
+		os.Exit(16)
```
malclocke: Almost all of the places where `logErrorAndExit` were being called previously were using exit code `16` except for the code that handled the exit status of the actual command that ran the tests https://github.com/buildkite/test-engine-client/blob/45db707f44cb53ef92da660edcc41bf3fb155f0c/main.go#L159-L162

This change lifts that logic here so this method only needs an `err` argument.

PR #374, opened 2025-10-21
Te 4789 add plan command

malclocke:
This changes adds a new command, `bktec plan [...]`, which generates a test plan without running any tests.

The command prints select metadata about the plan to `STDOUT`:

```
$ bktec plan
{"BKTEC_PLAN_IDENTIFIER":"facecafe","BKTEC_PARALLELISM":"42"}
```

This output format is intended to be used as input for `buildkite-agent env set` e.g.:

```
$ bktec plan | buildkite-agent env set --input-format json
Added:
+ BKTEC_PARALLELISM
+ BKTEC_PLAN_IDENTIFIER
```

### DIFF
--- a/internal/command/plan.go
+++ b/internal/command/plan.go
@@ -20,7 +20,12 @@ var planWriter io.Writer = os.Stdout
 
 // Structure of the JSON that is output when running `bktec plan`.
 type TestPlanSummary struct {
-	Identifier  string `json:"BUILDKITE_TEST_ENGINE_PLAN_IDENTIFIER"`
+	Identifier string `json:"BUILDKITE_TEST_ENGINE_PLAN_IDENTIFIER"`
+
+	// Parallelism is strictly an int not a string. It's represented as a string
+	// here because when this struct is Marshaled to JSON it's intended to be
+	// piped into buildkite-agent env set --input-format=json -, which requires
+	// string keys and string values.
 	Parallelism string `json:"BUILDKITE_TEST_ENGINE_PARALLELISM"`
 }
 


internal/command/files.go
```diff
@@ -0,0 +1,43 @@
...
+
+func getTestFiles(fileList string, testRunner TestRunner) ([]string, error) {
+	if fileList != "" {
+		return getTestFilesFromFile(fileList)
+	} else {
+		return testRunner.GetFiles()
+	}
+}
+
+func getTestFilesFromFile(path string) ([]string, error) {
```
malclocke: This method is moved unchanged from `command.Run` as it's now used in both `command.Run` and `command.Plan`.

internal/command/testdata/rspec/Gemfile
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "rspec"
```
malclocke: The new test in `internal/command/run_test.go` calls `rspec --dry-run` during the plan generation, so needs a `Gemfile` to be present.

internal/plan/type.go
```diff
@@ -35,6 +35,8 @@ type Task struct {
 
 // TestPlan represents the entire test plan.
 type TestPlan struct {
+	Identifier   string           `json:"identifier"`
+	Parallelism  int              `json:"parallelism"`
```
malclocke: These fields are already present in the JSON response from the test plan API, but were not declared in this struct.

internal/command/plan.go
```diff
@@ -0,0 +1,83 @@
...
+	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/env"
+	"github.com/buildkite/test-engine-client/internal/runner"
+	"github.com/urfave/cli/v3"
+)
+
+var planWriter io.Writer = os.Stdout
+
+// Structure of the JSON that is output when running `bktec plan`.
+type TestPlanSummary struct {
```
nprizal: Is this a placeholder or the final structure we want for triggering the dynamic pipeline? We’re using `BUILDKITE_TEST_ENGINE_XX` for other environment variables, so I’m wondering if using `BKTEC_XX` would be inconsistent.

nprizal (commented):
Tested locally and it works. But I have a question around the test plan summary structure and env vars key.

The other thing I noticed that the default error message for invalid option value is a bit cryptic. I wonder if we can customize the error message
> Incorrect Usage: invalid value "foobar" for flag -max-parallelism: strconv.ParseInt: parsing "foobar": invalid syntax
> invalid value "foobar" for flag -max-parallelism: strconv.ParseInt: parsing "foobar": invalid syntax

internal/command/plan.go
gchan: Comparing this file to run.go was useful for my review.

https://github.com/buildkite/test-engine-client/blob/d17feb2337462aa6fba054595fb56d39515df9a6/internal/command/run.go#L1

Will we add debugging statements later?
Is it worth adding TODOs about handling server errors and connectivity issues gracefully?

internal/command/plan.go
gchan: Consistency seems desirable.

My understanding these are temporary env vars that exist only for the purposes for generating the dynamic pipeline steps so perhaps we want to differentiate these from the other `BUILDKITE_TEST_ENGINE_XX` env vars?

internal/command/plan.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/env"
+	"github.com/buildkite/test-engine-client/internal/runner"
+	"github.com/urfave/cli/v3"
+)
+
+// Structure of the JSON that is output when running `bktec plan`.
+type TestPlanSummary struct {
+	Identifier  string `json:"BKTEC_PLAN_IDENTIFIER"`
+	Parallelism string `json:"BKTEC_PARALLELISM"`
```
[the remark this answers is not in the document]
gchan: I wonder if a shorter version of this comment be made into a code comment?

internal/command/plan.go
```diff
@@ -0,0 +1,83 @@
...
+}
+
+// This command creates a test plan via the API and returns the plan identifier
+// and parallelism of the plan in JSON format to STDOUT.
+func Plan(ctx context.Context, cmd *cli.Command) error {
+	env := env.OS{}
+
+	debug.SetDebug(env.Get("BUILDKITE_TEST_ENGINE_DEBUG_ENABLED") == "true")
+	debug.SetOutput(os.Stderr)
```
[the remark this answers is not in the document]
malclocke: Yes it's intentional, because the output on `STDOUT` for this command is intended to be parsed. If we "cross the streams" the output won't be valid JSON:

```json
DEBUG: some message
{"BKTEC_PARALLELISM":"5","BKTEC_PLAN_IDENTIFIER":"abc123/zxy987"}
```

I'll probably pitch for all debug to go to `STDERR` at some point, although I see looking at the git history this has flip-flopped over time between `STDOUT / STDERR` so I imagine there is some background there.

internal/command/plan.go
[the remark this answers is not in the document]
malclocke: Yeh, sounds good. Will update.

internal/command/plan.go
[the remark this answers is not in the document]
nprizal: I can't remember why we flip-flopped from streaming the debug to stdout/stderr, but it will be nice if they are being streamed to the same place.

nprizal (approved):
Looks good! I’ll leave the decision around making the `SetPlanWriter` public or private to you.

PR #378, opened 2025-10-23
[TE-4791] Allow bktec run command to specify 'plan-identifier' flag

gchan:
### Description

Allow the `run` command to support specifying a `plan_identifier`
flag. This would override the default behaviour deriving the
`identifier` by concatenating BUILDKITE_BUILD_ID and BUILDKITE_STEP_ID

This will be useful when a different step (or build) generates
the plan with the `plan` command. And we have a separate step
to run/execute the plan.

### Context

TE-4791

### Changes

Modified config.New to accept cli.Command which will derive
the config from the commands in addition to env vars.

### Testing

We currently don't have a comprehensive process to test the CLI tool 
and the flags.

### DIFF
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -55,6 +55,10 @@ func Run(ctx context.Context, cmd *cli.Command) error {
 		return fmt.Errorf("invalid configuration...\n%w", err)
 	}
 
+	if cmd.String("plan-identifier") != "" {
+		cfg.Identifier = cmd.String("plan-identifier")
+	}
+
 	testRunner, err := runner.DetectRunner(cfg)
 	if err != nil {
 		return fmt.Errorf("unsupported value for BUILDKITE_TEST_ENGINE_TEST_RUNNER: %w", err)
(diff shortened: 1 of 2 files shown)


internal/command/run.go
```diff
@@ -50,7 +50,7 @@ func Run(ctx context.Context, cmd *cli.Command) error {
 	printStartUpMessage()
 
 	// get config
-	cfg, err := config.New(env)
+	cfg, err := config.New(env, cmd)
```
malclocke: Changing the method signature of `config.New` is going to break a log of things.

This is likely to get reworked very soon, see #379 , so I'd recommend something like this which will keep the change minimal:

```go
	cfg, err := config.New(env)

    // check err ...

    if cmd.String("plan-identifier") != "" {
      cfg.Identifier = cmd.String("plan-identifier")
    }
```

basically set this directly here rather than in `Config.New`.

gchan: I originally had code you suggested but thought it would be better to consolidate the creation of the config by passing through the CLI flags. But with the extra context that's all going to change, I'll revert back to the minimal change.

PR #379, opened 2025-10-23
Convert env to cli

malclocke:
This change moves config parsing into the third party `cli` package.

Selected flags and related env vars are show in the command help:

```
$ ./test-engine-client --help
NAME:
   bktec - Buildkite Test Engine Client

USAGE:
   bktec [global options] [command [command options]]

COMMANDS:
   run      Run tests (default)
   plan     Generate test plan without running tests
   help, h  Shows a list of commands or help for one command

GLOBAL OPTIONS:
   --debug     Enable debug output (default: false) [$BUILDKITE_TEST_ENGINE_DEBUG_ENABLED]
   --help, -h  show help
   --version   print version information and exit (default: false)

   BUILD ENVIRONMENT

   --parallelism int  Run the specified number of bktec processes in parallel (default: 0) [$BUILDKITE_PARALLEL_JOB_COUNT]

   TEST ENGINE

   --access-token string  Buildkite API access token [$BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN]
   --suite-slug string    Buildkite suite slug [$BUILDKITE_TEST_ENGINE_SUITE_SLUG]

   TEST RUNNER

   --disable-retry-muted               Disable retry for muted tests (default: false) [$BUILDKITE_TEST_ENGINE_DISABLE_RETRY_FOR_MUTED_TEST]
   --files string                      Override the default test file discovery by providing a path to a file containing a list of test files (one per line)
   --result-path string                Path to the output file for the test runner [$BUILDKITE_TEST_ENGINE_RESULT_PATH]
   --retry-command string              Command to run when retrying failed tests. [$BUILDKITE_TEST_ENGINE_RETRY_CMD]
   --split-by-example                  Enable split by example (not supported by all test runners) (default: false) [$BUILDKITE_TEST_ENGINE_SPLIT_BY_EXAMPLE]
   --test-command string               Test command [$BUILDKITE_TEST_ENGINE_TEST_CMD]
   --test-engine-retry-count int       Number of times to retry failing tests (default: 0) [$BUILDKITE_TEST_ENGINE_RETRY_COUNT]
   --test-file-exclude-pattern string  Test file exclude pattern [$BUILDKITE_TEST_ENGINE_TEST_FILE_EXCLUDE_PATTERN]
   --test-file-pattern string          Test file pattern [$BUILDKITE_TEST_ENGINE_TEST_FILE_PATTERN]
   --test-runner string                Test runner [$BUILDKITE_TEST_ENGINE_TEST_RUNNER]
```

## Changes

Configuration parsing is moved into `main.go` using the `cli` package. This library can handle loading options either from command line flags or environment variables.

Currently some configuration validation is performed while loading the configuration in `config.readFromEnv` (e.g. `BUILDKITE_BUILD_ID` must not be blank) and some validation is performed in the `config.validate()` method. This change centralises the validation in `config.Validate()`.

Currently a string map of all of the env vars that are relevant to `bktec` are stored in `config.Env` and are sent with some API request payloads. Each member of this map duplicates values that are already in the `config` struct, e.g. `config.Env["BUILDKITE_BUILD_ID"] == config.BuildId`. This change uses JSON marshalling to convert the `Config` struct directly to JSON and removes the need for the `Config.Env` field.

The changes above allow the removal of the `env` package and the `config/read.go` and `config/env.go`.

## Downside of changes

The main downside of the changes is a difference of behaviour during the type casting of the configuration variables. Currently all failures casting env vars are aggregated and reported in a single error message:

```
$ BUILDKITE_PARRALEL_JOB_COUNT="foo" BUILDKITE_PARALLEL_JOB="bar" ./test-engine-client
--8<--
invalid configuration...
BUILDKITE_PARALLEL_JOB was "bar", must be a number
```

With these changes to use the `cli` package for config parsing only the first conversion failure is reported.

```
BUILDKITE_PARRALEL_JOB_COUNT="foo" BUILDKITE_PARALLEL_JOB="bar" ./test-engine-client 
could not parse "bar" as int value from environment variable "BUILDKITE_PARALLEL_JOB" for flag parallel-job: strconv.ParseInt: parsing "bar": invalid syntax
```

Whether the latter error message is more or less palatable depends on the taste of the reader, but only one of the errors is reported.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -6,7 +6,7 @@ type Config struct {
 	JobId   string `json:"BUILDKITE_JOB_ID"`
 	StepId  string `json:"BUILDKITE_STEP_ID"`
 	// AccessToken is the access token for the API.
-	AccessToken string
+	AccessToken string `json:"-"`
 	// Identifier is the identifier of the build.
 	Identifier string `json:"BUILDKITE_TEST_ENGINE_IDENTIFIER"`
 	// MaxRetries is the maximum number of retries for a failed test.
@@ -20,14 +20,14 @@ type Config struct {
 	// Parallelism is the number of parallel tasks to run.
 	Parallelism int `json:"BUILDKITE_PARALLEL_JOB_COUNT"`
 	// Maximum parallelism when calculating parallelism dynamically.
-	MaxParallelism int
+	MaxParallelism int `json:"-"`
 	// The path to the result file.
-	ResultPath string
+	ResultPath string `json:"-"`
 	// Whether a failed muted test should be retried.
 	// This is default to true because we want more signal for our flaky detection system.
-	RetryForMutedTest bool
+	RetryForMutedTest bool `json:"-"`
 	// ServerBaseUrl is the base URL of the test plan server.
-	ServerBaseUrl string
+	ServerBaseUrl string `json:"-"`
 	// SplitByExample is the flag to enable split the test by example.
 	SplitByExample bool `json:"BUILDKITE_TEST_ENGINE_SPLIT_BY_EXAMPLE"`
 	// SuiteSlug is the slug of the suite.


internal/command/run_test.go
```diff
@@ -1030,7 +1002,7 @@ func TestSendMetadata(t *testing.T) {
 				"BUILDKITE_TEST_ENGINE_TEST_CMD":    "bundle exec rspec",
 				"BUILDKITE_STEP_ID":                 "pqr",
 				// ensure that empty env vars is included in the request
-				"BUILDKITE_TEST_ENGINE_SPLIT_BY_EXAMPLE":          "",
+				"BUILDKITE_TEST_ENGINE_SPLIT_BY_EXAMPLE":          "false",
```
malclocke: A small change in behaviour, when parsing env vars from the `cli` the absence of this var - and any other `bool` vars - will be converted to `false` instead of an empty string.

internal/config/config.go
```diff
@@ -1,9 +1,16 @@
...
+	"fmt"
+
+	"github.com/urfave/cli/v3"
+)
 
 // Config is the internal representation of the complete test engine client configuration.
 type Config struct {
+	BuildId string
+	JobId   string
+	StepId  string
```
malclocke: These additional fields and `DebugEnabled` have been added to `Config` so we don't need to store the additional `Config.Env` element - see the additional removal comment.

malclocke:
> Draft LGTM. I couldn't find the code path where env vars are read into Config from main.go. Perhaps that will be clear soon.

@gchan each `Flag` definition in `main.go` has a pointer to the `Destination` variable, e.g.:

```go
		&cli.BoolFlag{
			Name:        "debug",
			Usage:       "Enable debug output",
			Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_DEBUG_ENABLED"),
			Destination: &cfg.DebugEnabled,

```

nprizal (commented):
Looks good! The benefit of setting the config using the `cli` is that you can display all the configs and its usage right in the terminal. Not to mention this will allow us to support file base config easily in the future. 

The `usage` description for each config might need to be improved, but its good for now.

internal/config/config.go
```diff
@@ -41,18 +48,22 @@ type Config struct {
 	Branch string
 	// JobRetryCount is the count of the number of times the job has been retried.
 	JobRetryCount int
-	// Env provides access to environment variables.
-	// It's public because many tests in other packages reference it (perhaps they should not).
-	Env env.Env
```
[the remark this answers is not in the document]
malclocke: > I noticed the AccessToken doesn't have json key name mapping, is that the only thing we need to do? Is there any other preventative measures we can take?

@nprizal very good catch! Showing my lack of golang knowledge here :sweat_smile: 

No `json` tag means the field is serialized with the default field name:

https://go.dev/play/p/9TyL7wQgure

The correct incantation is to add `json:"-"` to always omit the field, I will update.

https://go.dev/play/p/_pWoCh23rvw

internal/api/filter_tests.go
```diff
@@ -5,12 +5,13 @@ import (
...
 
+	"github.com/buildkite/test-engine-client/internal/config"
 	"github.com/buildkite/test-engine-client/internal/plan"
 )
 
 type FilterTestsParams struct {
-	Files []plan.TestCase   `json:"files"`
-	Env   map[string]string `json:"env"`
+	Files []plan.TestCase `json:"files"`
+	Env   config.Config   `json:"env"`
```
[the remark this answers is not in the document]
malclocke: @nprizal I might tackle this in a separate PR if that's OK with you? This one is already getting huge.

internal/config/config.go
```diff
@@ -1,70 +1,55 @@
...
+	OrganizationSlug string `json:"BUILDKITE_ORGANIZATION_SLUG"`
 	// Parallelism is the number of parallel tasks to run.
-	Parallelism int
+	Parallelism int `json:"BUILDKITE_PARALLEL_JOB_COUNT"`
 	// Maximum parallelism when calculating parallelism dynamically.
-	MaxParallelism int
+	MaxParallelism int `json:"-"`
 	// The path to the result file.
-	ResultPath string
+	ResultPath string `json:"-"`
```
malclocke: I've maintained the list of config fields that are excluded from the environment metadata from before this change, but I expect some of these are not intentionally excluded. I may follow up to ensure all fields we might be interested in are sent.

nprizal (approved):
Awesome! Tested it locally, and it works (happy path)

PR #380, opened 2025-10-23
[TE-4860] Bump golangci-lint from 1.64.8 to 2.50.0 and update config

gchan:
### Description

test-engine-client is currently using golangci-lint 1.64.8 which was
missing a few modern linters and formatters which are useful and
we can enable.

Notably staticcheck is a popular linter which wasn't enabled.

### Context

TE-4860

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,7 +1,10 @@
 version: "2"
 linters:
+  # https://golangci-lint.run/docs/linters/#enabled-by-default
+  default: standard
   enable:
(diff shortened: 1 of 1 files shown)


gchan:
The following looks useful and are enabled in other Buildkite projects. We can stage these in later

gocritic https://github.com/go-critic/go-critic
gofumpt https://github.com/mvdan/gofumpt

I'm hesitant to bring them in just yet given there is some major code changes in flight.

PR #387, opened 2025-10-31
[TE-4869] Add target-time flag for dynamic parallelism calculation

gchan:
## Description

This PR adds support for a `--target-time` flag that allows users to specify a desired target completion time for their test suite when using dynamic parallelism. This flag works in conjunction with `--max-parallelism` to help the test plan calculation determine optimal parallelism based on both historical test timing data and target completion time.

The implementation adds:
- A new `--target-time` CLI flag that accepts duration values (e.g., `4m30s`)
- Validation to ensure `--target-time` is only used with `--max-parallelism`
- Validation to ensure the target time is within reasonable bounds (0-86400 seconds)
- Plumbing of the `target_time` parameter through the config, request params, and API client

When `--target-time` is set to 0 (default), the flag is ignored and dynamic parallelism calculations proceed as before.

## Context

**Ticket:** TE-4869

This is the first step to supporting target-time-based dynamic parallelism. The Test Engine API will use this parameter to calculate appropriate parallelism levels that aim to complete the test suite within the specified target time, while respecting the `max_parallelism` constraint.

## Verification

1. Build the client: `go build`
2. Test flag validation:
   - `./test-engine-client --target-time 5m` should error (requires --max-parallelism)
   - `./test-engine-client --max-parallelism 10 --target-time 5m` should accept the flags
   - `./test-engine-client --max-parallelism 10 --target-time -1s` should error (invalid range)
   - `./test-engine-client --max-parallelism 10 --target-time 100000s` should error (exceeds max)

## Deployment

Low risk. The `target_time` parameter is optional and backwards compatible - existing usage without the flag continues to work as before. The API will handle the new parameter appropriately.

## Rollback

Yes


main.go
```diff
@@ -247,6 +248,23 @@ var cliCommand = &cli.Command{
...
+						if v != 0 && cmd.Int("max-parallelism") == 0 {
+							return fmt.Errorf("target-time must be used in conjunction with --max-parallelism")
+						}
+
+						if v < 0 || v > 86400*time.Second {
+							return fmt.Errorf("target-time must be between 0 and 86400 seconds")
+						}
+
+						return nil
+					},
```
malclocke: If we end up making this an env var annoyingly we'll have to move this validation to `config.Validate()`, the flag parser doesn't call these actions when it reads from environment variables.

gchan: It seems like it was a unexpected change between v2 and v3 of the urfave/cli library.

https://github.com/urfave/cli/issues/2132
https://github.com/urfave/cli/issues/2041

There is a potential workaround suggested in the thread I haven't yet verified.

Should we go ahead and support allow this and the `max-parallelism` flag to source from environment variables?

malclocke: @gchan yeh, I looked at those tickets previously. I'd suggest moving it to `config.Validate()`. I think it's cleaner to consolidate the validation in one place.

Feel free to add an env var for `max-parallelism` at the same time. I was going to sweep back and add one anyway, it was an oversight rather than an intentional omission.

internal/command/request_param.go
```diff
@@ -27,6 +27,7 @@ func createRequestParam(ctx context.Context, cfg config.Config, files []string,
 			Identifier:     cfg.Identifier,
 			Parallelism:    cfg.Parallelism,
 			MaxParallelism: cfg.MaxParallelism,
+			TargetTime:     float64(cfg.TargetTime.Milliseconds() / 1000),
```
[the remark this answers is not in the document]
malclocke: @gchan are you OK to change this?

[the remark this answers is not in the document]
gchan: Done!

PR #390, opened 2025-11-05
Create separate validation for run and plan

malclocke:
Refactor configuration into 3 methods.
- A method for validation common to all commands
- A method for validation specific to `bktec run`
- A method for validation specific to `bktec plan`

### DIFF
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -6,32 +6,8 @@ import (
 	"time"
 )
 
-type ValidationOpts struct {
-	SkipParallelism bool
-}
-
-// Validate checks if the Config struct is valid and returns InvalidConfigError if it's invalid.
-func (c *Config) Validate(opts ValidationOpts) error {
-	if c.TargetTime != 0 {
-		if c.TargetTime <= 0 {
-			c.errs.appendFieldError("target-time", "was %s, must be greater than 0", c.TargetTime.String())
-		}
-
-		if c.TargetTime > time.Hour*24 {
-			c.errs.appendFieldError("target-time", "was %s, must be less than or equal to 24 hours", c.TargetTime.String())
-		}
-
-		if c.MaxParallelism == 0 {
-			c.errs.appendFieldError("max-parallelism", "must be set when target-time is set")
-		}
-	}
-
-	if c.MaxParallelism != 0 {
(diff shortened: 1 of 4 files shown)


internal/config/validate_test.go
```diff
@@ -241,33 +193,77 @@ func TestConfigValidate_ResultPathOptionalWithPytest(t *testing.T) {
...
 	c.TestRunner = "pytest"
 
-	err := c.Validate(opts)
+	err := c.validate()
 	if err != nil {
 		t.Errorf("config.validate() error = %v", err)
 	}
 }
 
-func TestTargetTimeParsing(t *testing.T) {
```
malclocke: @gchan I've maintained all the tests in this PR except for this test, which I've removed.

The test seems to be testing the behvaiour of stdlib code (`time.ParseDuration`) so I think it's unnecessary. LMK if that's incorrect and I'm missing something.

internal/config/validate.go
```diff
@@ -110,3 +59,70 @@ func (c *Config) Validate(opts ValidationOpts) error {
 
 	return nil
 }
+
+// Validation for the `bktec run` command
+func (c *Config) ValidateForRun() error {
+	_ = c.validate()
```
malclocke: We can ignore the return from `validate()` here because the errors will continue to be accumulated on `c.errs` by the checks in `ValidateForRun` and then all returned at the end of the method.

The same applies to `ValidateForPlan`.

The return value of `validate()` could technically be removed but it would require a rework of all of the existing tests.

internal/config/validate_test.go
```diff
@@ -241,33 +193,77 @@ func TestConfigValidate_ResultPathOptionalWithPytest(t *testing.T) {
...
+// Validation specific to `bktec run`
+func TestConfigValidateForRun_NodeIndexLessThanOne(t *testing.T) {
 	c := createConfig()
-	c.TargetTime, _ = time.ParseDuration("1.5s")
-	c.MaxParallelism = 10
+	c.NodeIndex = -1
```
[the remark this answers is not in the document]
malclocke: > Perhaps that means the test name should say 'less than 0'?

Oh damn, now I understand the confusion :facepalm: 

Sorry about that, updated.

PR #401, opened 2025-11-12
[TE-4927] Remove ability to run bktec run as bktec

gchan:
## Description

This PR removes the ability to run `bktec` without a command. Previously, running `bktec` was equivalent to `bktec run`, but this behavior has been removed as a breaking change for the 2.0.0 release.

Going forward, `bktec` will need to be explicitly called with the `run` command: `bktec run`.

The previous implementation had a fallback action that would call `run()` when no command was given. This also meant that typos like `bktec plaan` would incorrectly attempt to execute the run command, which was confusing for users.

Additionally, all inline flag definitions have been extracted into variables for better code organization and maintainability.

Note some of these flags referenced in the `plan` command are not strictly required for planning and we will remove these in future iterations.

## Context

Related to TE-4927

## Verification

- Run `bktec` without arguments - should show help/error instead of executing run
- Run `bktec run` - should execute normally

## Deployment

Low risk - this is a planned breaking change for the 2.0.0 release.

## Rollback

Yes

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -201,21 +201,62 @@ var retryCommandFlag = &cli.StringFlag{
 	Destination: &cfg.RetryCommand,
 }
 
+// Global Flags
+var versionFlag = &cli.BoolFlag{
+	Name:   "version",
+	Usage:  "print version information and exit",
+	Action: printVersion,
+}
+
+var debugFlag = &cli.BoolFlag{
+	Name:        "debug",
+	Usage:       "Enable debug output",
+	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_DEBUG_ENABLED"),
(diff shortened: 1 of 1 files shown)


cli.go
```diff
@@ -0,0 +1,344 @@
...
+				testFilePatternFlag,
+				testFileExcludePatternFlag,
+				testRunnerFlag,
+				resultPathFlag,
+				splitByExampleFlag,
+				failOnNoTestsFlag,
+				// Runner Retry Flags
+				disableRetryMutedFlag,
+				retryCommandFlag,
+				testEngineRetryCountFlag,
```
malclocke: :heart_eyes:

malclocke (approved):
@gchan love it!

PR #406, opened 2025-11-30
Publish `bktec` to AWS ECR

nprizal:
### Description

Update the release pipeline to publish `bktec` docker image to [Buildkite Public ECR](https://gallery.ecr.aws/buildkite/test-engine-client).

I've tested the ECR login in this [build](https://buildkite.com/buildkite/test-engine-client-release/builds/414/steps/canvas) and was successful. Note that no actual release was created as I added `--snapshot` option.

### DIFF
--- a/.buildkite/steps/release.sh
+++ b/.buildkite/steps/release.sh
@@ -40,4 +40,4 @@ echo "--- :key: :aws: Login to AWS ECR Public"
 )
 
 echo "+++ :rocket: Creating Release"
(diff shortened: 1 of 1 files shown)


.buildkite/steps/release.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env sh
+set -euo pipefail
```
malclocke: :+1:

PR #417, opened 2025-12-11
fix(jest): use --runTestsByPath for literal file path matching

### LINKED ISSUE #416: Jest runner fails to match test files with parentheses or brackets in paths (Next.js App Router)
## Problem

When using bktec with Jest, tests in directories containing parentheses (e.g., `(main)`) or brackets (e.g., `[catalogId]`) are silently excluded from test runs. This commonly affects Next.js App Router projects which use these characters as part of their routing conventions:

- `(folderName)` - Route groups (organizational folders excluded from URL path)
- `[param]` - Dynamic route segments

## Reproduction

Given a Next.js project structure like:
```
src/app/(main)/page.test.tsx
src/app/(main)/[catalogId]/product.test.tsx
src/app/(main)/layout.test.tsx
```

Running bktec with Jest will not execute any tests in these directories. The Jest output shows:
```
Pattern: src/app/(main)/page.test.tsx - 0 matches
```

## Root Cause

Jest interprets CLI arguments as regex patterns. When paths containing `(main)` are passed to Jest, the parentheses are interpreted as regex capture groups instead of literal characters. This causes the pattern matching to fail.

For example:
- `src/app/(main)/test.tsx` is interpreted as a regex where `(main)` is a capture group
- This doesn't match the literal directory path `src/app/(main)/test.tsx`

## Expected Behavior

Test files in directories with parentheses and brackets should be discovered and executed correctly.

## Environment

- Next.js App Router projects
- Any Jest version
- bktec wrapping Jest

## Workaround

Currently, there is no workaround other than renaming directories to avoid special characters, which is not feasible for Next.js App Router projects where these naming conventions are required.

jasonwbarnett:
## Summary

Add `--runTestsByPath` to the default Jest `TestCommand` so that file path arguments are treated as literal paths instead of regex patterns.

Fixes #416

## Problem

Jest interprets positional CLI arguments as regex patterns by default. When test file paths contain characters like `(main)` or `[catalogId]`, Jest treats them as regex metacharacters and fails to match any files.

This commonly affects Next.js App Router projects which use:
- `(folderName)` for route groups
- `[param]` for dynamic route segments

## Solution

The default `TestCommand` now includes `--runTestsByPath`:

```
npx jest --runTestsByPath {{testExamples}} --json --testLocationInResults --outputFile {{resultPath}}
```

This tells Jest to treat positional arguments as literal file paths rather than regex patterns. Per [Jest docs](https://jestjs.io/docs/cli#--runtestsbypath), this is the recommended approach for CI pipelines and is more performant when provided with multiple patterns.

The `RetryTestCommand` is unchanged because:
- Its default does not pass file paths as positional arguments
- Test names use `--testNamePattern` which intentionally accepts regex (already properly escaped with `regexp.QuoteMeta`)

## Changes

- Updated default `TestCommand` in `jest.go` to include `--runTestsByPath`
- Updated default command in `docs/jest.md`
- Added test case `TestJestCommandNameAndArgs_WithSpecialCharactersInPath` verifying Next.js style paths are passed through as-is

### DIFF
--- a/docs/jest.md
+++ b/docs/jest.md
@@ -10,7 +10,7 @@ export BUILDKITE_TEST_ENGINE_RESULT_PATH=tmp/jest-result.json
 By default, bktec runs Jest with the following command:
 
 ```sh
-npx jest {{testExamples}} --json --testLocationInResults --outputFile {{resultPath}}
+npx jest --runTestsByPath {{testExamples}} --json --testLocationInResults --outputFile {{resultPath}}
 ```
 
 In this command, `{{testExamples}}` is replaced by bktec with the list of test files or tests to run, and `{{resultPath}}` is replaced with the value set in `BUILDKITE_TEST_ENGINE_RESULT_PATH`. You can customize this command using the `BUILDKITE_TEST_ENGINE_TEST_CMD` environment variable.
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -22,7 +22,7 @@ type Jest struct {
 
 func NewJest(j RunnerConfig) Jest {
 	if j.TestCommand == "" {
-		j.TestCommand = "npx jest {{testExamples}} --json --testLocationInResults --outputFile {{resultPath}}"
+		j.TestCommand = "npx jest --runTestsByPath {{testExamples}} --json --testLocationInResults --outputFile {{resultPath}}"
 	}
 
 	if j.TestFilePattern == "" {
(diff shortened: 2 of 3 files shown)


niceking:
Hey @jasonwbarnett! Yeah the team's still mostly on holidays and ramping back up. We'll try look at this this month, but for now you could compile your own bktec with these changes and run with that until we merge it to main?

PR #418, opened 2025-12-20
Add support for retrieving test examples in Pytest runner 

suo:
- Implemented GetExamples method to collect test cases using `pytest --collect-only -q`.
- Added parsePytestCollectOutput function to parse the output of the pytest command.
- Updated request_param.go to include pytest in the supported runners for file splitting.
- Added unit tests for GetExamples and parsePytestCollectOutput to ensure correct functionality.


nprizal (changes requested):
Hi @suo, thanks for opening the PR. I’ve noticed a discrepancy between the test collector regarding the generation of `scope` and `name`. Can you please update the PR so that it matches?

nprizal (approved):
Looks good! I've tested it the test plan includes individual examples with correct timing that means they match the data in the server.

PR #441, opened 2026-02-10
Codify runner feature support

malclocke:
This is a proof of concept of embedding the features supported by each runner in the runner implementation.

Aim here is to solve a few problems:

- Avoid "type switch" code. One example is fixed in this WIP, switching on split by example support base on runner name in `request_param.go`. There are others but I haven't implemented all of them in this WIP.
- Generate to feature support matrix in `README.md` from code. This has accumulated some drift over time in the past and required manual fixups. To prove the point there is fix for drift included in this PR :)
- This would give us a way to error when customers try to use a feature that is not supported in the runner, for example setting env vars or cli flags that are not supported by the runner. We've had confusion about this from users in the past.

Fixes TE-5965

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,15 +4,17 @@ Buildkite Test Engine Client (bktec) is an open source tool to orchestrate your 
 
 bktec supports multiple test runners and offers various features to enhance your testing workflow. Below is a comparison of the features supported by each test runner:
 
-| Feature                                            | RSpec | Jest | Playwright | Cypress | pytest | pants (pytest) | Go test | Cucumber | NUnit | Custom |
-| -------------------------------------------------- | :---: | :--: | :--------: | :-----: | :----: | :------------: | :-----: | :------: | :---: | :-----: |
-| Split tests by file[^1]                                |   ✅  |   ✅  |    ✅      |    ✅   |   ✅   |       ❌       |   ❌    |    ✅    |  ✅   | ✅      |
-| [Split slow files by individual test example](https://github.com/buildkite/test-engine-client/blob/main/docs/rspec.md#split-slow-files-by-individual-test-example)        |   ✅  |   ❌  |    ❌      |    ❌   |   ✅   |       ❌       |   ❌    |    ✅    |  ❌   |    ❌   |
-| Filter test files                                  |   ✅  |   ✅  |    ✅      |    ✅   |   ✅   |       ❌       |   ❌    |    ✅    |  ✅   |    ✅   |
-| Filter test by tags                               |   ❌  |   ❌  |    ❌      |    ❌   |   ✅   |       ❌       |   ❌    |    ❌    |  ❌   |    ❌  |
-| Automatically retry failed test                    |   ✅  |   ✅  |    ✅      |    ❌   |   ✅   |       ✅       |   ✅    |    ✅    |  ✅   |    ❌   |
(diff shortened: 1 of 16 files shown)


internal/command/request_param.go
```diff
@@ -18,16 +19,16 @@ import (
...
 	for _, file := range files {
 		testFiles = append(testFiles, plan.TestCase{
 			Path: file,
 		})
 	}
 
-	// Splitting files by example is only supported for rspec, cucumber, and pytest runners
-	if runner.Name() != "RSpec" && runner.Name() != "Cucumber" && runner.Name() != "pytest" {
+	// Short circuit here if the runner doesn't support split by example
+	if !runner.SupportedFeatures().SplitByExample {
```
nprizal: nice!

README.md
```diff
@@ -4,15 +4,17 @@ Buildkite Test Engine Client (bktec) is an open source tool to orchestrate your
...
-| Automatically retry failed test                    |   ✅  |   ✅  |    ✅      |    ❌   |   ✅   |       ✅       |   ✅    |    ✅    |  ✅   |    ❌   |
-| Mute tests (ignore test failures)                  |   ✅  |   ✅  |    ✅      |    ❌   |   ✅   |       ✅       |   ✅    |    ✅    |  ✅   |    ✅   |
-| Skip tests                                         |   ✅  |   ❌  |    ❌      |    ❌   |   ❌   |       ❌       |   ❌    |    ✅    |  ❌   |    ❌   |
+<!-- DO NOT MANUALLY EDIT THE TABLE BELOW. The contents can be generate with `go run util/supported_features/main.go` -->
+
+| Feature | RSpec | Jest | Playwright | Cypress | pytest | pytest-pants | gotest | Cucumber | NUnit | Custom test runner |
+| --- | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: | :---: |
+| Split tests by file[^1] | ✅ | ✅ | ✅ | ✅ | ✅ | ❌ | ❌ | ✅ | ✅ | ✅ |
+| [Split slow files by individual test example](https://github.com/buildkite/test-engine-client/blob/main/docs/rspec.md#split-slow-files-by-individual-test-example) | ✅ | ❌ | ✅ | ❌ | ✅ | ❌ | ❌ | ✅ | ❌ | ❌ |
```
malclocke, line 12: The table has been updated with the output from `go run util/supported_features/main.go`

The diff of the 2 tables can be inspected visually with the "Display the rich diff" button in the file header.

The only differences in content are slight changes in the runner names in the header, and "Split by example" is corrected from `false` to `true` for Playwright.

internal/command/request_param.go
nprizal, line 1: https://github.com/buildkite/test-engine-client/blob/a824a7638499dd95d3a2dfc6b3722a1a57108af3/internal/command/request_param.go#L71

That can be replaced with `runner.SupportedFeatures().FilterTestsByTag`

malclocke: I've tried to keep the touch and time to merge minimal with this PR by just including one cleanup example, and intend on following up over time with others.

I can try and clean everything up in this single PR if you like, LMK.

nprizal: Sounds good

nprizal (approved):
This is nice!

PR #463, opened 2026-03-18
Add {{testExamples}} support to pytest-pants runner

jasonwbarnett:
## Summary

Enable `{{testExamples}}` substitution in the `pytest-pants` runner, allowing bktec to inject intelligently-sharded test file paths into `pants test` commands.

## Motivation

When using pants as a build system, users often want to:
1. Use an external tool (e.g., `pants filter` with `--changed-since` and `--changed-dependents=transitive`) to determine which tests need to run based on changed files and transitive dependencies
2. Pass that list to bktec via `--files` to create a balanced test plan using historical timing data
3. Have each parallel node execute its assigned shard through `pants test`, preserving pants' PEX building, caching, and environment management

Previously, the `pytest-pants` runner explicitly rejected `{{testExamples}}`, making it impossible to use bktec's intelligent test splitting with pants as the test executor. The only options were either using bktec's splitting _without_ pants execution, or using pants execution _without_ intelligent splitting.

Pants' built-in `--test-shard` distributes tests without considering execution time, which leads to significantly unbalanced shards — for example, some agents finishing in 10 minutes while others run for 25 minutes. By replacing `--test-shard` with bktec's timing-aware distribution, users can keep pants handling everything it's good at (resolving dependencies, building PEX files, caching) while bktec ensures each node gets a balanced share of the work.

## How it works

The workflow separates target resolution from test execution:

1. `pants filter` resolves affected test targets into an explicit list
2. `bktec plan --files affected_tests.txt` creates a balanced plan using historical timing data from Buildkite Test Engine
3. Each parallel node runs `bktec run --files affected_tests.txt`, which fetches the plan and substitutes `{{testExamples}}` with its assigned targets before invoking `pants test`

When `{{testExamples}}` is absent from the test command, the existing behavior is preserved — pants handles test selection and bktec only adds retries and muting.

## Documentation

The `docs/pytest-pants.md` has been rewritten with a complete end-to-end guide covering prerequisites, the two-step plan + run workflow, a full Buildkite pipeline example, required environment variables, and a comparison table against `--test-shard`.

## Test plan

- [x] 6 new unit tests covering `{{testExamples}}` substitution (multiple files, paths with spaces, single file, extra pants flags, empty test cases, missing separator)
- [x] All existing tests pass (no regressions)
- [x] `go vet ./...` clean
- [x] Validated end-to-end in a real Buildkite pipeline with 58 pants test targets split across 4 shards — 3 consecutive green builds

### DIFF
--- a/docs/pytest-pants.md
+++ b/docs/pytest-pants.md
@@ -96,7 +96,7 @@ tests/models/test_user.py
 
 ### Example: Replacing `--test-shard` with intelligent splitting
 
-A common pants CI pattern is to use `pants filter` with `--changed-since` and `--changed-dependents=transitive` to resolve the affected test targets for a PR, then shard them across parallel agents using pants' built-in `--test-shard`. While this works, the native sharding distributes tests without considering execution time, which often leads to unbalanced shards — some agents finish in seconds while others run for minutes.
+A common pants CI pattern is to use `pants filter` with `--changed-since` and `--changed-dependents=transitive` to resolve the affected test targets for a PR, then shard them across parallel agents using pants' built-in `--test-shard`. While this works, the native sharding distributes tests without considering execution time, which often leads to significantly unbalanced shards — for example, some agents finishing in 10 minutes while others run for 25 minutes.
 
 By replacing `--test-shard` with bktec's test splitting, you get shards balanced by historical timing data. Here's how the workflow changes:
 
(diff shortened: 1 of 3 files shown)


jasonwbarnett:
Pretty excited about this one. It makes it possible to leverage Buildkite Test Engine's test splitting within the pants ecosystem. We've had significant imbalance across our 25–50 shards (depending on test volume) and can't wait to use this to balance them out and reclaim some much needed CI time.

jameshill:
Had a chat with @jasonwbarnett this morning and there more work to on targets that need to be done before we can support this. So we're closing this and will look for alternative options.

PR #472, opened 2026-04-02
Add Windows CI test support

pda:
### Description

Add a native Windows AMD64 test step to the Buildkite pipeline so the Go test suite runs on Windows in CI. This follows the same pattern as the agent repo: run on a Windows queue without Docker, invoke `bash .buildkite\steps\tests.sh` directly on the native agent.

Pact-based tests are excluded on Windows (pact-go native code does not compile there), and Unix-only runner tests are skipped at runtime. The Windows test command omits coverage/race flags to start with a minimal passing baseline.

### Context

- [TE-5508](https://linear.app/buildkite/issue/TE-5508/bktec-windows-testing-in-ci)
- Modeled after `buildkite/agent` Windows CI pattern

### Changes

- **Pipeline**: New `:windows: Windows AMD64 Tests` step on `$BKTEC_RUNNERS_WINDOWS_QUEUE` (defaults to `agent-runners-windows-amd64`), parallelism 2, no Docker, no AWS role/SSM plugins
- **Test script**: `tests.sh` conditionally skips pact-go install, yarn/Playwright setup, coverage/race flags, and coverage report on Windows (`go env GOOS` check)
- **Pact test files**: Added `//go:build !windows` build tag to all 5 pact-based test files in `internal/api/` — these do not compile on Windows
- **Runner tests**: Added `runtime.GOOS == "windows"` skip guards to 3 signal/Unix tests in `command_test.go` and 4 bash-fixture tests in `custom_test.go`

### Testing

- Verified `GOOS=windows GOARCH=amd64 go test -exec=true ./...` passes (all packages compile)
- Verified `go vet ./...` passes
- Existing Linux tests unaffected (build tags and skips are Windows-only)

### Related

- Resolves TE-5508

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -1,5 +1,6 @@
 env:
   AWS_DEFAULT_REGION: us-east-1
+  BKTEC_RUNNERS_WINDOWS_QUEUE: "${BKTEC_RUNNERS_WINDOWS_QUEUE:-agent-runners-windows-amd64}"
 
 steps:
   - name: ":golangci-lint: Lint"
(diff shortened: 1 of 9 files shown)


pda:
We don't have Windows CI runners/agents available, and don't have a sane path to get/maintain them.

The ones used by the buildkite-agent Windows tests should not be approached under any circumstances. Trust me.

We could trigger a GHA workflow from a Buildkite step to run the tests on their Windows runners. But I don't really want to manage the GHA FG-PAT or App credentials for that.

So I'm calling this unfeasible until Buildkite has better Windows support e.g. hosted agents for Windows.

PR #480, opened 2026-04-14
[TE-5578] Auto-collect git metadata for plan command

gchan:
## Description

When the plan command runs with test selection enabled (`--selection-strategy`), it now automatically collects git commit metadata and sends it with the API request. This removes the need for customers to manually script metadata collection with `--metadata` flags.

The auto-collection gathers up to 17 fields in three phases:
- **Commit metadata** (single `git log` call): SHA, parents, author/committer name/email/date, message
- **Diff fields** (explicit `git merge-base` then 4 `git diff` commands): files changed, numstat, full diff, raw diff
- **Context fields**: branch name, base branch, pipeline slug, build UUID

Base branch resolution uses a 3-level fallback: explicit `--metadata base_branch=...`, `$BUILDKITE_PULL_REQUEST_BASE_BRANCH`, then auto-detection via remote/HEAD → main → master. The `--remote` flag (default `"origin"`, configurable via `BUILDKITE_TEST_ENGINE_REMOTE`) controls which git remote is used.

Each phase degrades gracefully — git failures log debug warnings and skip affected keys without blocking CI. User-provided `--metadata` values always take precedence over auto-collected values. Empty auto-collected values are filtered at the merge site to avoid sending meaningless keys in the API request.

This also DRYs up shared logic between the new plan path and the existing backfill command (TE-5476): record parsing (`parseRecord`), diff command execution (`runDiffCommands`), and struct-to-map conversion (`ToMap`) are now shared. Both paths use explicit two-arg diff (`base commit`) rather than triple-dot syntax.

## Context

Resolves [TE-5578](https://linear.app/buildkite/issue/TE-5578)

## Changes

- Add `ResolveBaseBranch` and `CollectPlanMetadata` for plan auto-collection
- Integrate git metadata auto-collection into `plan()`, gated on `--selection-strategy`
- Add `--remote` flag to plan command under "PREVIEW: TEST SELECTION" category
- Refactor `collectCommitMetadata` to parse into `CommitMetadata` struct via `ToMap`
- Extract shared `runDiffCommands` with two-arg diff form and `CommitDiffs.ToMap`
- Extract shared `parseRecord` from duplicated parsing logic in `metadata.go`
- Use explicit `git merge-base` instead of triple-dot syntax for diff collection
- Filter empty metadata values at the merge site via `mergeNonEmpty`
- Extract `MergeMetadata` with user-precedence semantics and tests
- Guard against empty remote in `ResolveBaseBranch`
- Route success log messages through `debug.Printf`, keep warnings on stderr
- Always include `parent_shas` in `CommitMetadata.ToMap` (empty string for root commits)
- Fix `TrimRight` → `TrimSuffix` for record separator stripping
- Use `buildRecord` test helper with package constants instead of hex escapes
- Use `cmp.Diff` with full expected maps across all tests
- Document auto-collection in README and update backfill docs with shared env var

## Verification

Backed by specs. AI assisted.

```bash
go test ./internal/git/ ./internal/command/ . -count=1
```

21 new tests covering `ResolveBaseBranch`, `CollectPlanMetadata`, `CommitMetadata.ToMap`, `CommitDiffs.ToMap`, and `MergeMetadata`. All tests use `cmp.Diff` against complete expected maps.

## Deployment

Low risk. Auto-collection is gated behind `BKTEC_PREVIEW_SELECTION` and only runs when `--selection-strategy` is set. No change to existing plan or backfill behaviour when selection is not enabled.

## Rollback

Yes

### DIFF
--- a/internal/command/plan.go
+++ b/internal/command/plan.go
@@ -13,6 +13,7 @@ import (
 	"github.com/buildkite/test-engine-client/internal/api"
 	"github.com/buildkite/test-engine-client/internal/config"
 	"github.com/buildkite/test-engine-client/internal/debug"
+	"github.com/buildkite/test-engine-client/internal/git"
 	"github.com/buildkite/test-engine-client/internal/plan"
 	"github.com/buildkite/test-engine-client/internal/runner"
 	"github.com/buildkite/test-engine-client/internal/version"
@@ -35,6 +36,11 @@ var (
 // This command creates a test plan via the API
 func Plan(ctx context.Context, cfg *config.Config, testFileList string, outputFormat PlanOutput, template string) error {
 	fmt.Fprintln(os.Stderr, "+++ Buildkite Test Engine Client: bktec "+version.Version+"\n")
+
+	// Auto-collect git metadata when selection is active
+	if cfg.SelectionStrategy != "" {
+		autoCollectGitMetadata(ctx, cfg, &git.ExecGitRunner{})
+	}
 
 	testRunner, err := runner.DetectRunner(cfg)
 	if err != nil {
@@ -145,6 +151,31 @@ func createTestPlan(ctx context.Context, cfg *config.Config, files []string, api
 	}
 
 	return testPlan, nil
+}
+
+// autoCollectGitMetadata collects git commit metadata and merges it into
+// cfg.Metadata. User-provided metadata values (from --metadata) take
+// precedence over auto-collected values.
+func autoCollectGitMetadata(ctx context.Context, cfg *config.Config, runner git.GitRunner) {
+	// Check if we're in a git repo
+	if _, err := runner.Output(ctx, "rev-parse", "--git-dir"); err != nil {
+		fmt.Fprintln(os.Stderr, "Warning: not a git repository, skipping metadata auto-collection")
+		return
+	}
+
+	// Use user-provided base_branch from --metadata if present
+	explicit := cfg.Metadata["base_branch"]
+	remote := cfg.Remote
+	baseBranch, err := git.ResolveBaseBranch(ctx, runner, explicit, remote)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: could not resolve base branch for diff metadata. "+
+			"Set --metadata base_branch=<branch> if your repo uses a non-standard default branch.\n")
+	} else {
+		debug.Printf("auto-detected base branch: %s", baseBranch)
+	}
+
+	autoMetadata := git.CollectPlanMetadata(ctx, runner, baseBranch)
+	cfg.Metadata = git.MergeMetadata(cfg.Metadata, autoMetadata)
 }
 
 func handleError(err error) error {
--- a/internal/command/plan.go
+++ b/internal/command/plan.go
@@ -168,8 +168,8 @@ func autoCollectGitMetadata(ctx context.Context, cfg *config.Config, runner git.
 	remote := cfg.Remote
 	baseBranch, err := git.ResolveBaseBranch(ctx, runner, explicit, remote)
 	if err != nil {
-		fmt.Fprintf(os.Stderr, "Warning: could not resolve base branch for diff metadata. "+
-			"Set --metadata base_branch=<branch> if your repo uses a non-standard default branch.\n")
+		fmt.Fprintln(os.Stderr, "Warning: could not resolve base branch for diff metadata. "+
+			"Set --metadata base_branch=<branch> if your repo uses a non-standard default branch.")
 	} else {
 		debug.Printf("auto-detected base branch: %s", baseBranch)
 	}
--- a/internal/git/auto_metadata.go
+++ b/internal/git/auto_metadata.go
@@ -80,51 +80,14 @@ func collectCommitMetadata(ctx context.Context, runner GitRunner, metadata map[s
 		return
 	}
 
-	meta, ok := parseCommitRecord(output)
+	record := strings.TrimRight(output, recordSeparator+"\n ")
+	meta, ok := parseRecord(record)
 	if !ok {
+		debug.Printf("Warning: git log returned unparseable output; skipping commit metadata")
 		return
 	}
 
 	mergeNonEmpty(metadata, meta.ToMap())
-}
-
-// parseCommitRecord parses a single git log record (using MetadataFormat
-// separators) into a CommitMetadata struct. Returns false if the record
-// is empty or has too few fields. Reuses the same parsing logic as
-// FetchBulkMetadata.
-func parseCommitRecord(output string) (CommitMetadata, bool) {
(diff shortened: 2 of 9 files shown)


gchan:
Evidence of this working in a local environment: https://github.com/buildkite/buildkite/pull/29016#issuecomment-4257634296 (private repo)

internal/command/plan.go
```diff
@@ -147,6 +153,31 @@ func createTestPlan(ctx context.Context, cfg *config.Config, files []string, api
...
+		return
+	}
+
+	// Use user-provided base_branch from --metadata if present
+	explicit := cfg.Metadata["base_branch"]
+	remote := cfg.Remote
+	baseBranch, err := git.ResolveBaseBranch(ctx, runner, explicit, remote)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Warning: could not resolve base branch for diff metadata. "+
+			"Set --metadata base_branch=<branch> if your repo uses a non-standard default branch.\n")
```
nprizal: There's no formatting here, maybe use `Fprintln`? Then we can remove the `\n`.
```suggestion
		fmt.Fprintln(os.Stderr, "Warning: could not resolve base branch for diff metadata. "+
			"Set --metadata base_branch=<branch> if your repo uses a non-standard default branch.")
```

nprizal:
@codex review

nprizal:
@codex review

internal/git/auto_metadata.go
```diff
@@ -0,0 +1,193 @@
...
+// git log call with the same format as FetchBulkMetadata, parses it into
+// a CommitMetadata struct, and flattens it into the metadata map via ToMap.
+func collectCommitMetadata(ctx context.Context, runner GitRunner, metadata map[string]string) {
+	output, err := runner.Output(ctx, "log", "-1", fmt.Sprintf("--format=%s", MetadataFormat))
+	if err != nil {
+		debug.Printf("Warning: git log failed, skipping commit metadata: %v", err)
+		return
+	}
+
+	record := strings.TrimSpace(strings.TrimSuffix(output, recordSeparator))
```
nprizal: 👋 Claude here — I was reviewing this PR and think this trim order silently corrupts every commit message on the plan path.

Real `git log` output ends in `\x1e\n` (git always appends a newline). `TrimSuffix` for `\x1e` sees the trailing `\n` and no-ops; `TrimSpace` then strips the `\n` but leaves `\x1e` behind — which rides along into `fields[8]` and survives `TrimSpace` (not Unicode whitespace). Every message ends up as `"...\n\u001e"` in the API payload, while the backfill path stores the same commit cleanly.

Tests miss it because `buildRecord` ends at `\x1e` with no trailing `\n` — the one shape where the current order happens to work.

```suggestion
	record := strings.TrimSuffix(strings.TrimSpace(output), recordSeparator)
```
A fixture ending in `\x1e\n` would lock it in.

nprizal: I'm not familiar with `git log` and the exising metadata backfill, so I'm not sure if this is valid. Feel free to ignore if you think its irelevant.

gchan: Seems legit:

image

internal/command/plan.go
gchan: Good call — switched to `Fprintln` and dropped the trailing `\n` in 51cc9d901de0f8b764b65f20ca156bd61a449caa.

internal/git/auto_metadata.go
```diff
@@ -0,0 +1,193 @@
...
+		{value: os.Getenv("BUILDKITE_PULL_REQUEST_BASE_BRANCH"), source: "BUILDKITE_PULL_REQUEST_BASE_BRANCH"},
+	}
+
+	for _, c := range candidates {
+		if c.value == "" {
+			continue
+		}
+		ref := c.value
+		if !strings.HasPrefix(ref, remote+"/") {
+			ref = remote + "/" + ref
```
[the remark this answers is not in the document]
gchan: Fixed in a780239eb1b5e335fd8e7dd59a98e607ce14c4c0. 

Rather than a `contains("/")` heuristic (which would misclassify legitimate bare-branch patterns like `release/v2` or `feature/foo` as qualified), `ResolveBaseBranch` now probes each candidate verbatim first, then falls back to `<remote>/<candidate>` if the literal value does not resolve. This handles every shape without heuristics:

- `main` — verbatim fails, `origin/main` succeeds
- `release/v2` — verbatim fails, `origin/release/v2` succeeds (old behaviour preserved)
- `upstream/main` — verbatim succeeds (previously was rewritten to `origin/upstream/main` and silently dropped)
- `refs/heads/release` — verbatim succeeds (previously was rewritten and dropped)
- `origin/main` — verbatim succeeds

Regression tests added: `TestResolveBaseBranch_ExplicitQualifiedDifferentRemote`, `TestResolveBaseBranch_ExplicitFullyQualifiedRef`, `TestResolveBaseBranch_EnvVarQualifiedRef`.

internal/git/auto_metadata.go
gchan: Nice catch — confirmed and fixed in eae355fc966ecbc1eb306c8b5a31808cda467e08.

We caught production evidence of this in the Kafka message stream: every plan-path commit message ended in the ASCII record separator byte (U+001E), JSON-encoded into the downstream payload. Exactly the failure mode you described:

1. `TrimSuffix` with the record separator saw the trailing `\n` at the end of real git output, no-oped.
2. `TrimSpace` then stripped the `\n`, leaving the record separator trapped inside `fields[8]`.
3. Per-field `TrimSpace` does not touch that byte (U+001E is not Unicode whitespace), so it rode into the commit message value and out the API payload.

Swapped the order to `strings.TrimSuffix(strings.TrimSpace(output), recordSeparator)` and added `TestCollectPlanMetadata_TrailingNewline` with a fixture ending in `recordSeparator + "\n"` (matching real git output) plus an explicit `!strings.Contains(got["message"], recordSeparator)` assertion. Verified the new test fails against the old code and passes against the fix.

Thanks for the detailed analysis, especially the note about the test fixture being the one shape where the buggy order happens to work.

gchan: For the record — the concern was real. Production Kafka data confirmed the trailing ASCII record separator byte (U+001E) on every plan-path commit message. Fixed in eae355fc966ecbc1eb306c8b5a31808cda467e08 with a regression test using a fixture that matches real git log output.

internal/git/auto_metadata.go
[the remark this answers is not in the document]
gchan: ^ This was AI

internal/git/auto_metadata.go
gchan: ^ Last two messages was AI.

nprizal (approved):
🚀

PR #487, opened 2026-04-23
[TE-5669] Consolidate duplicate handleError implementations in bktec

meghan-kradolfer:
### Description

`run.go` and `plan.go` each had their own handleError implementation with the same error-classification logic (retry timeout → fallback, billing error → fallback, everything else → hard fail). This PR extracts a single shared handleError into errors.go so future error branches only need to be added once.

This PR also updates all errors to be `stderr`, where previously it was a mix of `stdout` and `stderr`. I don't think this should be a problem as the job logs output both `stdout` and `stderr` so nothing should change

### Context

[TE-5669](https://linear.app/buildkite/issue/TE-5669/consolidate-duplicate-handleerror-implementations-in-bktec)

### Changes

- `errors.go` (new): shared `handleError(err error)` error — classifies API errors and prints warnings to stderr. Returns `nil` for recoverable errors (caller falls back), or the original error for hard failures.
- `errors_test.go` (new): direct unit tests covering all three branches (retry timeout, billing error, unknown error).
- `plan.go`: removed duplicate `handleError`, removed unused errors import.
- `run.go`: replaced inline closure with calls to the shared function. Callers construct fallback plans themselves when handleError returns `nil`.

### Testing

- `errors_test.go` (new): direct unit tests covering all three branches — retry timeout returns nil with stderr warning, billing error returns nil with `stderr` warning, unknown error passes through unchanged with no stderr output.
- Existing integration tests (`TestFetchOrCreateTestPlan_*`, `TestPlanJSON_*`) continue to pass and cover the end-to-end behaviour.

### DIFF
--- a/internal/command/errors.go
+++ b/internal/command/errors.go
@@ -1,0 +1,27 @@
+package command
+
+import (
+	"errors"
+	"fmt"
+	"os"
+
+	"github.com/buildkite/test-engine-client/internal/api"
+)
+
+// handleError classifies API errors and prints user-facing warnings to stderr.
+// Returns nil for recoverable errors (caller should fall back to non-intelligent splitting),
+// or the original error for unrecoverable failures.
+func handleError(err error) error {
+	if errors.Is(err, api.ErrRetryTimeout) {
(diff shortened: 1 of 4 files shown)


malclocke (approved):
Nice tidy up 👍

PR #491, opened 2026-04-28
[TE-5667] Replace generic fallback warning with differentiated failure messages

meghan-kradolfer:
### Description

A lot of fallback scenarios currently surface the same generic message: *"Could not fetch or create plan from server, falling back to non-intelligent splitting."* Users have no way to tell whether they're hitting a timeout, a billing issue, a missing suite, or an auth failure, and no indication of what to do next.

This PR replaces that catch-all with differentiated messages per failure mode. The API client now classifies non-2xx responses into typed errors (`AuthError`, `ForbiddenError`, `NotFoundError`, `BadRequestError`, `UnprocessableEntityError`, plus the existing `BillingError`). 
`handleError` maps each type to either:

- A **recoverable warning** with a fallback notice (timeout, billing, 422, 404. Build proceeds with non-intelligent splitting), or
- A **fatal error** that stops the build (401 auth, 403 access denied, 400 invalid request) so the user fixes the underlying problem.

Output is also colorized (yellow ⚠️ for warnings, red ❌ for fatal errors), with `NO_COLOR` honoured.

### Context

- Linear: [TE-5667](https://linear.app/buildkite/issue/TE-5667)
- Investigation doc: [Improving Test Splitting Communication](https://www.notion.so/buildkite/Improving-Test-Splitting-Communication-348b8dbc2c8980c78143c2e648d9edd3)

### Changes

- `internal/api/client.go`: new typed errors (`AuthError`, `ForbiddenError`, `NotFoundError`, `BadRequestError`, `UnprocessableEntityError`); `DoWithRetry` switches on `resp.StatusCode` to return the right type.
- `internal/api/fetch_test_plan.go`: uses `errors.As` instead of inspecting `resp.StatusCode`.
- `internal/command/errors.go`: `handleError` classifies each typed error and prints a labelled, coloured message; new `warn()` and `fatal()` helpers; `warnErrorPlan()` consolidates the empty-task-list fallback message.
- `internal/command/run.go`, `plan.go`: use the shared helpers; `Run`/`Plan` no longer wrap the returned error so the formatted `❌` message reaches the user verbatim.
- Tests added for each typed error path in both `internal/api` and `internal/command`.

#### Example output

**Error**
Screenshot 2026-04-28 at 3 33 25 PM

**Warning**
Screenshot 2026-04-28 at 3 33 41 PM

### Testing

- Unit tests added for each typed error in `internal/api/client_test.go`, `internal/command/errors_test.go`, and `internal/command/run_test.go`.
- Manual: triggered each failure mode against a local mock server and confirmed messages render correctly with and without `NO_COLOR`

### DIFF
--- a/internal/command/errors.go
+++ b/internal/command/errors.go
@@ -13,7 +13,7 @@ import (
 // or the original error for unrecoverable failures.
 func handleError(err error) error {
 	if errors.Is(err, api.ErrRetryTimeout) {
-		fmt.Fprintln(os.Stderr, "⚠️ Could not fetch or create plan from server, falling back to non-intelligent splitting. Your build may take longer than usual.")
+		fmt.Fprintln(os.Stderr, "⚠️ Test Engine API timed out after 130s. Falling back to non-intelligent splitting. Your build may take longer than usual.")
 		return nil
 	}
 
@@ -21,6 +21,26 @@ func handleError(err error) error {
 		fmt.Fprintln(os.Stderr, billingError.Message+"\n")
 		fmt.Fprintln(os.Stderr, "⚠️ Falling back to non-intelligent splitting. Your build may take longer than usual.")
 		return nil
+	}
+
+	if authError := new(api.AuthError); errors.As(err, &authError) {
+		fmt.Fprintln(os.Stderr, "❌ Invalid API token. Check BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN is set and valid. See https://buildkite.com/docs/apis/managing-api-tokens")
+		return err
+	}
+
+	if forbiddenError := new(api.ForbiddenError); errors.As(err, &forbiddenError) {
+		fmt.Fprintf(os.Stderr, "❌ Access denied: %s. Check your API token has the required scopes and organization access. See https://buildkite.com/docs/apis/managing-api-tokens\n", forbiddenError.Message)
+		return err
+	}
+
+	if errors.As(err, new(*api.NotFoundError)) {
+		fmt.Fprintln(os.Stderr, "⚠️ Suite not found. Check BUILDKITE_TEST_ENGINE_SUITE_SLUG is correct. Falling back to non-intelligent splitting.")
+		return nil
+	}
+
+	if badRequestError := new(api.BadRequestError); errors.As(err, &badRequestError) {
+		fmt.Fprintf(os.Stderr, "❌ Invalid request: %s\n", badRequestError.Message)
+		return err
 	}
 
 	return err
--- a/internal/command/errors.go
+++ b/internal/command/errors.go
@@ -25,19 +25,23 @@ func ansi(code string) string {
 	return code
 }
 
-// warn prints a recoverable warning to stderr followed by an optional hint
+// printWarn prints a recoverable warning to stderr followed by an optional hint
 // and the fallback notice.
-func warn(label, message string, hints ...string) {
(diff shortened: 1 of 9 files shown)


internal/command/errors.go
```diff
@@ -8,20 +8,77 @@ import (
...
+	colorBold   = ansi("\033[1m")
+	colorReset  = ansi("\033[0m")
+)
+
+func ansi(code string) string {
+	if os.Getenv("NO_COLOR") != "" {
+		return ""
+	}
+	return code
+}
```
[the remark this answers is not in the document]
meghan-kradolfer: I don't think this matters for bktec as the env is set before the process begins

internal/command/errors.go
```diff
@@ -8,20 +8,77 @@ import (
...
+
+// warn prints a recoverable warning to stderr followed by an optional hint
+// and the fallback notice.
+func warn(label, message string, hints ...string) {
```
nprizal: [nit] just to clarify that this function doesn't actually return anything, and only print warning

```suggestion
func printWarn(label, message string, hints ...string) {
```

internal/command/errors.go
```diff
@@ -8,20 +8,77 @@ import (
...
+	for _, h := range hints {
+		fmt.Fprintln(os.Stderr, h)
+	}
+	fmt.Fprintf(os.Stderr, "%s%s%s\n", colorYellow, fallbackExtra, colorReset)
+}
+
+// fatal formats an unrecoverable error message with red bold styling.
+func fatal(label, message string) error {
+	return fmt.Errorf("%s❌ %s:%s %s", colorRed+colorBold, label, colorReset, message)
+}
```
nprizal: [nit]
```suggestion
func fatal(label, a any) error {
	return fmt.Errorf("%s❌ %s:%s %v", colorRed+colorBold, label, colorReset, message)
}
```
I think you can do this, so you can use this function on line 79

```
# line 79
return fatal("Unexpected error", err)
```

internal/command/errors.go
```diff
@@ -8,20 +8,77 @@ import (
...
 	if billingError := new(api.BillingError); errors.As(err, &billingError) {
-		fmt.Fprintln(os.Stderr, billingError.Message+"\n")
-		fmt.Fprintln(os.Stderr, "⚠️ Falling back to non-intelligent splitting. Your build may take longer than usual.")
+		warn("Billing Error", billingError.Message)
 		return nil
 	}
 
-	return err
+	if unprocessableError := new(api.UnprocessableEntityError); errors.As(err, &unprocessableError) {
+		warn("Unavailable", unprocessableError.Message)
```
nprizal: I'm not quite sure about this. "Unavailable" can mean that the server is offline, but I don't think thats what "Unprocessable Entity" means. should we let this handled as "Unexpected error"

meghan-kradolfer: The only 422 that I can find that bktec hits on test plan endpoints comes from `verify_api_active`: when `ANALYTICS_DISABLE_API` feature flag is active for the org, which doesn't feel quite right, but something I don't really want to change.
I could update the warning label to be something like "Test Engine unavailable", or I can just let it fall through to the "Unexpected error", although this is a fatal error so not sure if we want that or if we want it as warning + fallback. What do you think?

nprizal: Hmm, it's a shame that we return `422` for that situation. I think `403` better suit that situation. 

Anyway, lets keep it as it is, and probably add a comment why we treat "Unprocessable entity" as "Unavailable".

nprizal (approved):

Issue #485, opened 2026-04-20
Retry constructs invalid pytest node ID for collection errors (empty scope)


jasonwbarnett:
## Summary

When bktec retries a pytest collection error (e.g. an import failure), it constructs an invalid node ID because the scope is empty. This causes `no tests ran` on the retry attempt.

The test still correctly fails overall — the collection error is reported as a failure and bktec exits non-zero. But the retry is wasted and the error message is confusing.

## Root cause

In [`internal/runner/pytest.go` line 106](https://github.com/buildkite/test-engine-client/blob/v2.4.0/internal/runner/pytest.go#L106), bktec constructs retry node IDs by joining scope and name:

```go
Path: fmt.Sprintf("%s::%s", test.Scope, test.Name)
```

For normal test failures this produces valid node IDs like `tests/foo.py::test_bar`.

For collection errors, the scope is empty — and correctly so. A collection error's node ID is just the file path (e.g. `tests/foo.py`) with no `::` separator. When the test-collector-python plugin splits on `::`, it gets `scope=""` and `name="tests/foo.py"`. The unconditional join produces:

```
::tests/foo.py
```

This is not a valid pytest node ID, so pytest reports `no tests ran`.

## Suggested fix

Handle empty scope when constructing the retry path:

```go
if test.Scope != "" {
    path = fmt.Sprintf("%s::%s", test.Scope, test.Name)
} else {
    path = test.Name
}
```

This would produce `tests/foo.py` for collection errors, which is a valid pytest node ID. Pytest would attempt to collect the file again — it would still fail (the import error can't be fixed by re-running), but the error message would be clear and the retry behavior would be consistent.

## Context

This issue was discovered while investigating [buildkite/test-collector-python#106](https://github.com/buildkite/test-collector-python/issues/106), where collection errors were silently dropped from the JSON report. The fix in [buildkite/test-collector-python#108](https://github.com/buildkite/test-collector-python/pull/108) adds collection errors to the report, which surfaces this pre-existing issue in bktec's retry logic.

## Reproduction

https://github.com/altana-ai/bktec-collection-error-repro — the third scenario (`03-bktec-retry`) demonstrates the invalid node ID on retry after applying the test-collector-python fix.

PR #492, opened 2026-04-28
Skip retries for pytest collection errors

jasonwbarnett:
### Description

When a test file fails to import (e.g. `ModuleNotFoundError`), pytest fires `pytest_collectreport` instead of the normal `pytest_runtest_*` hooks. The [test-collector-python plugin](https://github.com/buildkite/test-collector-python/pull/108) now tags these entries with `test.pytest_collection_error=true` in the JSON report.

Previously, bktec would treat these as regular test failures and attempt to retry them. This is pointless — the underlying issue is a broken import or syntax error, not a flaky test. Retrying just wastes CI time.

This PR parses the `tags` field from the JSON report and excludes collection errors from retry candidates while still counting them as failures in the run status.

### Context

- Companion PR: https://github.com/buildkite/test-collector-python/pull/108
- Issue: https://github.com/buildkite/test-collector-python/issues/106

### Changes

- Add `Tags` field to `TestEngineTest` to parse tags from the JSON report
- Add `CollectionError` flag to `TestResult` and `RecordCollectionError()` method to `RunResult`
- Exclude collection errors from `FailedTests()` and `FailedMutedTests()` (no retries)
- `Status()` still reports `RunStatusFailed` when collection errors are present
- Add "Collection Errors (will not be retried)" section to the report output
- Unit tests for all new behavior, including JSON tag parsing and pytest runner integration

### Testing

- All existing tests pass (failures in `go test ./...` are pre-existing: missing rspec, gotestsum, cypress, pact_ffi)
- New unit tests cover: collection errors excluded from retries, excluded from muted retries, status still reports failed, JSON tag parsing
- Pytest runner integration test uses a fixture with both a real failure and a tagged collection error, verifying only the real failure appears in `FailedTests()`

### DIFF
--- a/internal/runner/run_result.go
+++ b/internal/runner/run_result.go
@@ -76,17 +76,42 @@ func (r *RunResult) RecordTestResult(testCase plan.TestCase, status TestStatus) 
 	}
 }
 
+// RecordCollectionError records a test collection error (e.g. import failure).
+// Collection errors are recorded as failed but excluded from retries.
+func (r *RunResult) RecordCollectionError(testCase plan.TestCase) {
+	test := r.getTest(testCase)
+	test.Status = TestStatusFailed
+	test.CollectionError = true
+	test.ExecutionCount++
+	if r.mutedTestLookup[mutedTestIdentifier(testCase)] {
+		test.Muted = true
+	}
+}
+
 // FailedTests returns a list of test cases that failed.
+// Collection errors are excluded because retrying them is pointless
+// (the underlying issue is a broken import or syntax error, not a flaky test).
 func (r *RunResult) FailedTests() []plan.TestCase {
 	var failedTests []plan.TestCase
 
 	for _, test := range r.tests {
-		if test.Status == TestStatusFailed && !test.Muted {
+		if test.Status == TestStatusFailed && !test.Muted && !test.CollectionError {
 			failedTests = append(failedTests, test.TestCase)
 		}
 	}
 
 	return failedTests
+}
+
+// CollectionErrors returns test cases that failed due to collection errors (e.g. import failures).
+func (r *RunResult) CollectionErrors() []plan.TestCase {
+	var errors []plan.TestCase
+	for _, test := range r.tests {
+		if test.CollectionError && !test.Muted {
+			errors = append(errors, test.TestCase)
+		}
+	}
+	return errors
 }
 
 func (r *RunResult) MutedTests() []TestResult {
@@ -115,7 +140,7 @@ func (r *RunResult) FailedMutedTests() []plan.TestCase {
 	var failedTests []plan.TestCase
 
 	for _, test := range r.tests {
-		if test.Status == TestStatusFailed && test.Muted {
+		if test.Status == TestStatusFailed && test.Muted && !test.CollectionError {
 			failedTests = append(failedTests, test.TestCase)
 		}
 	}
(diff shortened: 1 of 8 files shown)


jasonwbarnett:
@niceking This is 100% vibe coding with a single prompt (with the other context from https://github.com/buildkite/test-collector-python/pull/108). I am reviewing it now.

jasonwbarnett:
I'm going to be honest, this touches quite a lot of existing concepts that I'm not super familiar with. Ideally someone who is opinionated and knows bktec can provide meaningful feedback.

internal/runner/run_result.go
```diff
@@ -174,11 +219,11 @@ func (r *RunResult) Status() RunStatus {
 		return RunStatusUnknown
 	}
 
-	if len(r.FailedTests()) > 0 {
+	if len(r.FailedTests()) > 0 || r.collectionErrorCount() > 0 {
```
jasonwbarnett, line 222: I really don't know if this makes sense or not. It's likely it does given a collection error typically means one or more tests (even if muted) can't be run so it's probably a good idea to bail early and treat collection errors as things that can't be muted/ignored.

internal/runner/run_result.go
```diff
@@ -174,11 +219,11 @@ func (r *RunResult) Status() RunStatus {
...
 		return RunStatusFailed
 	}
 
-	if r.passedTestsCount()+r.skippedTestsCount()+len(r.FailedMutedTests()) == len(r.tests) {
+	if r.passedTestsCount()+r.skippedTestsCount()+len(r.FailedMutedTests())+r.mutedCollectionErrorCount() == len(r.tests) {
```
jasonwbarnett, line 226: I don't think it's even possible to mute an entire set of tests so I don't think this makes sense, but I could be wrong. In other words, a collection error could mean one or more tests that may or may not be muted. I think it's far simpler to simply treat all collection failures as true failures and to exclude this mutedCollectionErrorCount from this completely.

niceking:
@jasonwbarnett i've also released python collector v1.4.3 so if you could test against the changes. I'll tag the team in for a bktec review!

internal/runner/run_result.go
niceking: Yeah I think that's right

internal/runner/run_result.go
niceking: Yeah we can't mute at the test file level. And yeah we should treat collection failures as true failures and exclude the mutedCollectionErrorCount from this completely.

internal/runner/run_result.go
```diff
@@ -241,8 +286,9 @@ type TestEngineTest struct {
 	Name     string
 	Scope    string
 	Location string
-	FileName string `json:"file_name,omitempty"`
+	FileName string            `json:"file_name,omitempty"`
```
niceking, line 289: 🤔

internal/runner/run_result.go
```diff
@@ -241,8 +286,9 @@ type TestEngineTest struct {
...
 	Result   TestStatus
+	Tags     map[string]string `json:"tags,omitempty"`
```
niceking, line 291: 🤔

niceking (commented):
can we change all `CollectionError` to `PytestCollectionError` 🙏

jasonwbarnett:
@pda I saw you merged https://github.com/buildkite/test-engine-client/pull/553

I wasn't sure if you thought this PR still has any merit or if I should just close it. Let me know!

jasonwbarnett:
Superseded by #553 (merged), which skips retries for tagged pytest collection errors and also fixes the invalid `::file.py` retry selector (#485). Closing.

PR #493, opened 2026-04-29
Remove Pact contract testing

pda:
### Description

Remove the Pact consumer-side integration from bktec.

Pact hasn't paid for itself: it's the only Pact integration in the buildkite stack, no one is actively driving it, and the API it covers (test plan, test plan metadata, file timings, filter tests) is small and slow-moving. The signal we actually rely on is elsewhere — bktec is versioned with a support window, the Test Engine API has its own request specs, and real integration issues surface through dogfooding and customer reports faster than a contract test would catch them.

This PR removes the consumer side. The server-side changes will be removed separately.

### Changes

- `internal/api/*_test.go`: replace pact-go consumer tests with plain `net/http/httptest` fixtures. Each test still asserts the URL, method, headers, request body, and response decoding. `*config.Config` is compared with `cmpopts.IgnoreUnexported` because `Config` has an unexported `errs` field.
- `internal/api/pacts/`: deleted (generated `TestEngineClient-TestEngineServer.json`).
- `go.mod` / `go.sum`: drop `github.com/pact-foundation/pact-go/v2`.
- `bin/publish-pact`, `bin/release-pact-version`, `bin/pact-record-support-ended`: deleted. The last one was a bit bitrotten — it still targeted the legacy `TestSplitterClient` pacticipant.
- `bin/setup`, `.buildkite/steps/tests.sh`: drop `pact-go` install steps.
- `.buildkite/steps/publish-pacts.sh`: deleted (already orphaned).
- `.buildkite/pipeline.yml`: drop `internal/api/pacts/*` artifact path.
- `CHANGELOG.md`: note the change under `Unreleased`.

### Testing

```
go test ./internal/api/...   # all pass
go build ./...               # clean
go vet ./...                 # clean
```

The five rewritten tests assert the same URL/method/headers/body/response-decoding as before, just without a broker.

### Related

- Resolves TE-5717
- Parent: TE-5716


internal/api/create_test_plan_test.go
```diff
@@ -46,98 +36,74 @@ func TestCreateTestPlan(t *testing.T) {
...
+			t.Errorf("Content-Type header = %q", got)
+		}
+
+		var got TestPlanParams
+		if err := json.NewDecoder(r.Body).Decode(&got); err != nil {
+			t.Fatalf("decoding request body: %v", err)
+		}
+		if diff := cmp.Diff(got, params); diff != "" {
+			t.Errorf("request body diff (-got +want):\n%s", diff)
+		}
```
gchan: I wonder if we should compare against a JSON so the request format is asserted. The current approach decodes back into TestPlanParams, which means JSON tag changes round-trip cleanly and slip through. e.g. renaming json:"target_time" to json:"targetTime" would still pass this test.

internal/api/fetch_files_timing_test.go
```diff
@@ -2,72 +2,65 @@ package api
...
+	svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodPost {
+			t.Errorf("request method = %q, want %q", r.Method, http.MethodPost)
+		}
+		if r.URL.Path != "/v2/analytics/organizations/buildkite/suites/rspec/test_files" {
+			t.Errorf("request path = %q", r.URL.Path)
+		}
+		if got := r.Header.Get("Authorization"); got != "Bearer asdf1234" {
+			t.Errorf("Authorization header = %q", got)
+		}
```
gchan, line 27: Nit: This and TestFilterTests_SlowFiles doesn't assert Content-Type: application/json but TestCreateTestPlan and TestPostTestPlanMetadata do.

internal/api/fetch_test_plan_test.go
```diff
@@ -3,152 +3,106 @@ package api
...
+	c := NewClient(ClientConfig{
+		AccessToken:      "asdf1234",
+		OrganizationSlug: "buildkite",
+		ServerBaseUrl:    svr.URL,
 	})
 
+	got, err := c.FetchTestPlan(context.Background(), "rspec", "abc123", 0)
 	if err != nil {
-		t.Error("Error mocking provider", err)
+		t.Errorf("FetchTestPlan() error = %v", err)
```
gchan: Use t.Fatal for the error check so next execution (where got != nil) is skipped

gchan (approved):
A few minor suggestions/nits but LGTM.

Should we remove pact from the .gitignore file too?

pda:
🤖 Removed `internal/api/pacts/` from `.gitignore` in c789cbd.

internal/api/fetch_test_plan_test.go
pda: 🤖 Done in 03efa14 — switched to `t.Fatalf` so the `got != nil` check below is skipped on error.

internal/api/fetch_files_timing_test.go
pda: 🤖 Done in 9fea2e7 — added `Content-Type: application/json` assertions to both `TestFetchFilesTiming` and `TestFilterTests_SlowFiles`.

internal/api/create_test_plan_test.go
pda: 🤖 Done in 58bdd30. Added an `assertJSONBody` helper in [client_test.go](https://github.com/buildkite/test-engine-client/blob/58bdd30/internal/api/client_test.go) that compares the captured body against a literal JSON string (parsed into `any` so whitespace/key-order don't matter, but tag names do). Applied it to all six tests that assert request bodies — `TestCreateTestPlan`, `TestCreateTestPlan_SplitByExample`, `TestCreateTestPlan_MutedTests`, `TestFilterTests_SlowFiles`, `TestFetchFilesTiming`, and `TestPostTestPlanMetadata`.

The two tests that embed `*config.Config` (`TestFilterTests_SlowFiles`, `TestPostTestPlanMetadata`) now spell out every `BUILDKITE_*` env field on the wire — verbose, but it means a tag rename or a new env var is a deliberate test update, which is the point.

PR #494, opened 2026-05-04
[TE-5706] Print split summary to stderr from plan and run commands

meghan-kradolfer:
### Description

After a test plan is resolved (fetched or newly created), print a human-readable split summary to stderr from both plan and run commands, so users can see at a glance how the plan was built and how much of it relied on historical timing data.

Output examples:

Screenshot 2026-05-01 at 2 32 55 PM

All-default case:

Screenshot 2026-04-29 at 4 49 52 PM

At parallelism=1 the server skips per-format timing data for performance reasons, so the breakdown falls back to the plan-level `known_timings_ratio`.

### Context

- [TE-5706](https://linear.app/buildkite/issue/TE-5706/print-split-summary-to-stderr-from-plan-and-run-commands)
- Parent: [TE-5668](https://linear.app/buildkite/issue/TE-5668/surface-a-split-summary-after-plan-resolution) (Surface a split summary after plan resolution)
- Depends on the API changes that emit `timing_metadata` and per-case `timing_sample_size` - done in https://github.com/buildkite/buildkite/pull/29320

### Changes

- New `internal/plan/summary.go` with `PrintSplitSummary` covering single-format, mixed-format, and parallelism=1 fallback paths
- New schema fields on `internal/plan/type.go`: `TestPlan.TimingMetadata`, `TestPlan.KnownTimingsRatio`, `TestCase.TimingSampleSize` (omitempty)
- Wire `PrintSplitSummary` into `internal/command/plan.go` and `internal/command/run.go` after plan resolution
- Skipped for fallback plans and when neither `timing_metadata` nor `known_timings_ratio` is present (older/cached plans)

### Testing

- Unit tests in `internal/plan/summary_test.go` cover mixed history, all-no-history, null median, parallelism=1 ratio path, ratio=0, example-format, mixed file+example, and fallback skip
- Round-trip and unmarshal tests in `internal/plan/type_test.go` cover the new fields with and without timing metadata

### DIFF
--- a/internal/command/plan.go
+++ b/internal/command/plan.go
@@ -72,6 +72,8 @@ func Plan(ctx context.Context, cfg *config.Config, testFileList string, outputFo
 		debug.Printf("Test plan created. Identifier: %q, Parallelism: %d", testPlan.Identifier, testPlan.Parallelism)
 	}
 
+	plan.PrintSplitSummary(os.Stderr, testPlan)
+
 	switch outputFormat {
 
 	case PlanOutputJSON:
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -68,6 +68,8 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 
 	debug.Printf("My favourite ice cream is %s", testPlan.Experiment)
 
+	plan.PrintSplitSummary(os.Stderr, testPlan)
+
 	// get plan for this node
 	thisNodeTask := testPlan.Tasks[strconv.Itoa(cfg.NodeIndex)]
 
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -68,7 +68,7 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 
 	debug.Printf("My favourite ice cream is %s", testPlan.Experiment)
 
-	plan.PrintSplitSummary(os.Stderr, testPlan)
+	plan.PrintSplitSummary(os.Stdout, testPlan)
 
 	// get plan for this node
 	thisNodeTask := testPlan.Tasks[strconv.Itoa(cfg.NodeIndex)]
--- a/internal/plan/summary.go
+++ b/internal/plan/summary.go
@@ -1,0 +1,81 @@
+package plan
+
+import (
+	"fmt"
+	"io"
+	"strconv"
+)
+
+// PrintSplitSummary writes a human-readable summary of the resolved test plan
+// to w (typically os.Stderr). It is a no-op when the plan does not carry
+// timing metadata (fallback plans, error plans, or cached plans created
+// before the server began emitting timing_metadata).
+func PrintSplitSummary(w io.Writer, p TestPlan) {
+	if p.Fallback || p.TimingMetadata == nil {
+		return
+	}
+
+	total := 0
+	known := 0
+	exampleMode := false
+	for _, task := range p.Tasks {
+		for _, tc := range task.Tests {
+			total++
+			if tc.TimingSampleSize > 0 {
+				known++
+			}
+			if tc.Format == TestCaseFormatExample {
+				exampleMode = true
+			}
+		}
+	}
+	if total == 0 {
+		return
+	}
+
+	noun := "files"
+	if exampleMode {
+		noun = "examples"
+	}
+
+	nodes := p.Parallelism
+	if nodes == 0 {
+		nodes = len(p.Tasks)
+	}
+
+	fmt.Fprintln(w, "\n+++ Buildkite Test Engine Client: 📊 Split summary")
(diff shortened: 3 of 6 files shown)


meghan-kradolfer:
@codex review

internal/command/plan.go
```diff
@@ -72,6 +72,8 @@ func Plan(ctx context.Context, cfg *config.Config, testFileList string, outputFo
 		debug.Printf("Test plan created. Identifier: %q, Parallelism: %d", testPlan.Identifier, testPlan.Parallelism)
 	}
 
+	plan.PrintSplitSummary(os.Stderr, testPlan)
```
nprizal, line 75: Not sure if this is intentional, but maybe we should print to the stdout instead?
```suggestion
	plan.PrintSplitSummary(os.Stdout, testPlan)
```

internal/command/run.go
```diff
@@ -68,6 +68,8 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 
 	debug.Printf("My favourite ice cream is %s", testPlan.Experiment)
 
+	plan.PrintSplitSummary(os.Stderr, testPlan)
```
nprizal: Same as above

internal/plan/summary.go
```diff
@@ -0,0 +1,171 @@
...
+	}
+	if known < 0 {
+		known = 0
+	}
+	printFormatBreakdown(w, total, known, noun, nil, false)
+}
+
+// countByFormat returns (total, known) for cases of the given format. The
+// empty (default) Format value is treated as TestCaseFormatFile.
+func countByFormat(p TestPlan, format TestCaseFormat) (total, known int) {
```
nprizal, line 73: TIL `go` has _named return_ i.e. (`total` & `known`)

internal/plan/summary.go
```diff
@@ -0,0 +1,171 @@
...
+
+	total := fileTotal + exampleTotal
+	if total == 0 {
+		return
+	}
+
+	nodes := p.Parallelism
+	if nodes == 0 {
+		nodes = len(p.Tasks)
+	}
```
nprizal: Is there a scenario where `p.Parallelism` is `0` but the `Tasks` is not empty?

internal/plan/summary.go
```diff
@@ -0,0 +1,171 @@
...
+// printRatioBreakdown renders a single-node summary using the plan-level
+// known_timings_ratio. Per-format metadata is unavailable at parallelism == 1,
+// so it delegates to printFormatBreakdown with meta == nil; the breakdown text
+// is rendered without parenthesised duration values.
+func printRatioBreakdown(w io.Writer, total int, ratio float64, noun string) {
+	known := int(float64(total)*ratio + 0.5)
+	if known > total {
+		known = total
+	}
+	if known < 0 {
```
nprizal: In what situation will this be true?

internal/command/plan.go
meghan-kradolfer: It looks like we use stderr in other places when printing the `+++ Buildkite Test Engine Client: bktec` banner messages on the `plan`, but the `run` uses stdout. I am not sure the reasoning between why we use a mix here. Asking my LLM it says:

> stdout from plan is consumed by downstream tools (buildkite-agent env set for JSON mode, buildkite-agent pipeline upload via stdin for upload mode), so anything human-readable has to go to stderr to avoid corrupting that stream

Should we update the `run` to use stdout and keep the plan as-is to prevent that downstream JSON issue?

nprizal: Ah! I remember. We don't want to output to `stdout` for `plan` because the output of `plan` command will be parsed by a script to trigger the dynamic pipeline so the output should be clean.<br><br>Can we follow the same pattern then? print to `stderr` for `plan` and `stdout` for `run`? We need to tidy this up in the future tho.<br><br>cc @malclocke

internal/plan/summary.go
meghan-kradolfer: I don't think this will ever happen, will remove

internal/plan/summary.go
meghan-kradolfer: Also just think was the LLM being defensive, happy to remove

internal/command/plan.go
malclocke: Yeh `plan` needs to print informational stuff on `Stderr` because it has a mode to output JSON to `Stdout`.

internal/command/run.go
malclocke: I think leave `plan` printing to `Stderr` but change this to `Stdout` :sweat_smile: 

```suggestion
	plan.PrintSplitSummary(os.Stdout, testPlan)
```

nprizal (approved):
:shipit:

PR #495, opened 2026-05-05
Deduplicate flag definitions

malclocke:
Removes duplication of the flag definitions for `plan` and `run` commands.

### DIFF
--- a/cli.go
+++ b/cli.go
@@ -447,6 +447,37 @@ var uploadFlag = &cli.StringFlag{
 	Destination: &cfg.UploadFile,
 }
 
+// Groupings common to multiple commands
+var buildEnvironmentFlags = []cli.Flag{
+	organizationSlugFlag,
+	buildIDFlag,
+	jobIDFlag,
+	stepIDFlag,
+	branchFlag,
+	retryCountFlag,
+	parallelJobFlag,
+}
+
+var testEngineFlags = []cli.Flag{
+	accessTokenFlag,
+	suiteSlugFlag,
+	baseURLFlag,
+}
+
+var runnerEnvironmentFlags = []cli.Flag{
+	testCommandFlag,
+	testFilePatternFlag,
+	testFileExcludePatternFlag,
+	testRunnerFlag,
+	resultPathFlag,
+	splitByExampleFlag,
+	locationPrefixFlag,
+	// Runner Retry Flags
+	disableRetryMutedFlag,
+	retryCommandFlag,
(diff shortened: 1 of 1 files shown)


cli.go
```diff
@@ -465,71 +496,29 @@ func runCommandFlags() []cli.Flag {
...
-		// Runner Retry Flags
-		disableRetryMutedFlag,
-		retryCommandFlag,
-		testEngineRetryCountFlag,
 	}
+	flags = append(flags, buildEnvironmentFlags...)
+	flags = append(flags, testEngineFlags...)
+	flags = append(flags, runnerEnvironmentFlags...)
+	flags = append(flags, parallelismFlag)
+	flags = append(flags, failOnNoTestsFlag)
```
malclocke, line 504: These 2 are "runner flags" that are only available on the `run` command and not on `plan`

gchan: Nit: Would you say `parallelismFlag` is a build environment flag (an env var set by the pipeline)? It's only relevant to the `run` command, not `plan` as you mentioned.

I'm asking because `parallelJobFlag` and `parallelismFlag` has been paired together historically and `parallelismFlag` remains in the `BUILD ENVIRONMENT` category.

gchan (approved):
LGTM with an optional nit

cli.go
malclocke: Yeh, you're right, sorry.

As you say `parallelismFlag` is grouped in the `BUILD ENVIRONMENT` category, and it appears there regardless of the order it's declared here. So the only "nit" correction required would be my comment on the pull request I think.

gchan: I think we can leave it as is.

I wanted to point out that `buildEnvironmentFlags` doesn't contains all the `BUILD ENVIRONMENT` category flags. This makes sense as `parallelismFlag` is not a common flag. My initial expectation was `buildEnvironmentFlags` contained all flags, but the comment  `Groupings common to multiple commands` clarifies the intention of the variable.

New question: Is `parallelJobFlag` only relevant to the the `run` command too?

malclocke: > New question: Is parallelJobFlag only relevant to the the run command too?

Yes, I think it is only relevant to the `run` command.

I've not tried to change anything about which flags are on which command in this PR, it's just deduplicating. I think `parallelJobFlag` falls under the umbrella of this comment in `planCommandFlags`:

```golang
	// Some of these flags are not strictly required for planning,
	// we will remove these in future iterations.
```

PR #496, opened 2026-05-06
Centralize buildCommand method for runners

malclocke:
Principal motivation for this change is to extract all of the duplicated logic for creating an `exec.Cmd` when executing the test runners. We'll be adding some additional behaviour to this in a follow up and this will avoid duplicating the new behaviour across 10 different runners.

- Removes duplicate `TestRunner` interface from `command` package
- Adds `CommandNameAndArgs` to `runner.TestRunner` interface
- Adds `LocationPrefix` to `runner.TestRunner` interface
- Extracts duplicated logic from all runners into `buildCommand`

### DIFF
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -19,20 +19,6 @@ import (
 	"github.com/olekukonko/tablewriter"
 )
 
-type TestRunner interface {
-	// Run takes testCases as input, executes the test against the test cases, and mutates the runner.RunResult with the test results.
-	Run(result *runner.RunResult, testCases []plan.TestCase, retry bool) error
-	// GetExamples discovers all tests within given files.
-	// This function is only used for split by example use case. Currently only supported by RSpec.
-	GetExamples(files []string) ([]plan.TestCase, error)
-	// GetFiles discover all test files that the runner should execute.
-	// This is sent to server-side when creating test plan.
-	// This is also used to obtain a fallback non-intelligent test splitting mechanism.
-	GetFiles() ([]string, error)
-	Name() string
-	LocationPrefix() string
-}
-
 const Logo = `
 ______ ______ _____
 ___  /____  /___  /____________
@@ -225,7 +211,7 @@ func sendMetadata(ctx context.Context, apiClient *api.Client, cfg *config.Config
 // For next reader, there is a small caveat with current implementation:
 // - testCases and timeline are both expected to be mutated.
 // - testCases in this case serve both as input and output -> we should probably change it.
-func runTestsWithRetry(testRunner TestRunner, testsCases *[]plan.TestCase, maxRetries int, mutedTests []plan.TestCase, timeline *[]api.Timeline, retryForMutedTest bool, failOnNoTests bool) (runner.RunResult, error) {
+func runTestsWithRetry(testRunner runner.TestRunner, testsCases *[]plan.TestCase, maxRetries int, mutedTests []plan.TestCase, timeline *[]api.Timeline, retryForMutedTest bool, failOnNoTests bool) (runner.RunResult, error) {
 	attemptCount := 0
 
 	// Create a new run result with muted tests to keep track of the results.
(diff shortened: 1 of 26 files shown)


internal/command/run.go
```diff
@@ -19,20 +19,6 @@ import (
...
-	Run(result *runner.RunResult, testCases []plan.TestCase, retry bool) error
-	// GetExamples discovers all tests within given files.
-	// This function is only used for split by example use case. Currently only supported by RSpec.
-	GetExamples(files []string) ([]plan.TestCase, error)
-	// GetFiles discover all test files that the runner should execute.
-	// This is sent to server-side when creating test plan.
-	// This is also used to obtain a fallback non-intelligent test splitting mechanism.
-	GetFiles() ([]string, error)
-	Name() string
-	LocationPrefix() string
```
malclocke, line 33: We have 2 interfaces with the same name and basically identical method definitions, `command.TestRunner` (this one) and `runner.TestRunner`. I don't know how we've ended up with 2 but it seems unnecessary, so I'm removing `command.TestRunner` here.

internal/runner/detector.go
```diff
@@ -7,30 +7,12 @@ import (
...
-	TagFilters             string
-	TestCommand            string
-	TestFileExcludePattern string
-	TestFilePattern        string
-}
-
-func (c RunnerConfig) LocationPrefix() string {
-	return c.locationPrefix
-}
-
```
malclocke, line 28: Extracted this to a separate file `internal/runner/runner_config.go`

internal/runner/gotest.go
```diff
@@ -110,10 +114,10 @@ func (g GoTest) GetFiles() ([]string, error) {
 	return validPackages, nil
 }
 
-func (p GoTest) commandNameAndArgs(cmd string, testCases []plan.TestCase) (string, []string, error) {
```
malclocke, line 113: This change is an outlier to the rest. All of the existing methods had this method sig:

```golang
commandNameAndArgs(cmd string, testCases []string)
```

But this method took `[]plan.TestCase` as the second arg. I've lifted the `[]plan.TestCase -> []string` conversion out of this method into the caller (`Run`) so that this method sig can be changed to match the other runners.

internal/runner/gotest.go
```diff
@@ -124,7 +128,7 @@ func (p GoTest) commandNameAndArgs(cmd string, testCases []plan.TestCase) (strin
 		cmd = cmd + " " + concatenatedPackages
 	}
 
-	cmd = strings.Replace(cmd, "{{resultPath}}", p.ResultPath, 1)
```
malclocke, line 127: I guess this was cut'n'paste from the python runner, changing this `p` to `g` while I'm here.

internal/runner/command.go
```diff
@@ -8,6 +8,15 @@ import (
 	"syscall"
 )
 
+func buildCommand(runner TestRunner, testCases []string, retry bool) (*exec.Cmd, error) {
```
malclocke, line 11: This is the crux of this PR. We currently duplicate this code in each runner - e.g. 10 times. This PR pulls this duplicate code into a single method.

internal/runner/cypress.go
```diff
@@ -38,13 +37,12 @@ func (c Cypress) Run(result *RunResult, testCases []plan.TestCase, retry bool) e
 	for i, tc := range testCases {
 		testPaths[i] = tc.Path
 	}
-	cmdName, cmdArgs, err := c.commandNameAndArgs(c.TestCommand, testPaths)
+
+	cmd, err := buildCommand(c, testPaths, retry)
```
nprizal, line 45:
```suggestion
	cmd, err := buildCommand(c, testPaths, false)
```
I don't think Cypress support retry, so maybe we want to be explicit about it here?

internal/runner/playwright.go
```diff
@@ -106,7 +104,12 @@ func (p Playwright) getTestResultsFromSuite(suite PlaywrightReportSuite, suiteNa
 	return testResults
 }
 
-func (p Playwright) commandNameAndArgs(cmd string, testCases []string) (string, []string, error) {
+func (p Playwright) CommandNameAndArgs(testCases []string, retry bool) (string, []string, error) {
+	cmd := p.TestCommand
+	if retry {
+		cmd = p.RetryTestCommand
```
nprizal, line 114: The `RetryTestCommand` is nil by default, we need to fallback to `TestCommand` if `RetryTestCommand` is nil

malclocke: I've added a commit to add the following to the `New` method for both cypress and playwright to match the existing behaviour of a bunch of the other runners

https://github.com/buildkite/test-engine-client/blob/87bd73c0ad3186c8cb1ce867c6f078eff48b48ad/internal/runner/gotest.go#L26-L28

nprizal (approved):

PR #500, opened 2026-05-08
Rework `runner.buildCommand()` method signature

malclocke:
This change reworks `runner.buildCommand` and `runner.CommandNameAndArgs` from taking a `[]string` argument to a `[]plan.TestCase` argument.

All the runners currently convert an array of `plan.TestCase` structs into an array of `string` and pass the string array into the methods above. This works for all the runners **except** for Jest, which needs access to an array of `plan.TestCase` instances to use the generic `runner.buildCommand` method.

The change in this PR will allow a rework of the Jest `Run` method to use `runner.buildCommand` in the same way as all the other runners.

### DIFF
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -61,16 +61,12 @@ func (j Jest) GetFiles() ([]string, error) {
 func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) error {
 	var cmd *exec.Cmd
 
-	testPaths := make([]string, len(testCases))
-	for i, testCase := range testCases {
-		testPaths[i] = testCase.Path
-	}
-
+	testPaths := pathsFromTestCases(testCases)
 	slices.Sort(testPaths)
 	testPaths = slices.Compact(testPaths)
 
 	if !retry {
-		commandName, commandArgs, err := j.CommandNameAndArgs(testPaths, false)
+		commandName, commandArgs, err := j.CommandNameAndArgs(testCases, false)
 		if err != nil {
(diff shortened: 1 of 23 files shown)


internal/runner/jest.go
```diff
@@ -61,16 +61,12 @@ func (j Jest) GetFiles() ([]string, error) {
...
-		testPaths[i] = testCase.Path
-	}
-
+	testPaths := pathsFromTestCases(testCases)
 	slices.Sort(testPaths)
 	testPaths = slices.Compact(testPaths)
 
 	if !retry {
-		commandName, commandArgs, err := j.CommandNameAndArgs(testPaths, false)
+		commandName, commandArgs, err := j.CommandNameAndArgs(testCases, false)
```
nprizal, line 69: The `!retry` case now uses the raw `testCases` without the sort/dedupe. I think that's fine since the `testCases` for initial run will be a unique test file paths from the test plan.

I'm not sure why the sort/dedupe was done for the non `!retry` case in the initial implementation, but I don't think that's necessary. Could we move the sort/compact inside the `else` so it's clear the dedupe is only needed when retrying failed tests (which can share a file)?

nprizal (commented):
Looks good, just one comment about Jest

internal/runner/jest.go
malclocke: I'm working on a follow up to port Jest to use `buildCommand`, I'll address this in that branch.

nprizal:
We talked about it in-sync. @malclocke will address it in next PR

nprizal (approved):

PR #501, opened 2026-05-08
Bind BUILDKITE_PARALLEL_JOB_COUNT in `bktec plan` (TE-5745)

niceking:
### Description

`bktec plan` was silently dropping `BUILDKITE_PARALLEL_JOB_COUNT` because `parallelismFlag` was only registered in `runCommandFlags()`, not in `planCommandFlags()`. As a result, `cfg.Parallelism` stayed at its zero value, and the entire `Config` struct sent as the `env` payload to `filter_tests` had `"BUILDKITE_PARALLEL_JOB_COUNT": 0`. That trips the server's `return [] if @parallelism <= 0` guard in `TestFilesFilter#slow_files`, so split-by-example could never identify slow files when planning.

The fix is to add `parallelismFlag` to `planCommandFlags()` so the env var / `--parallelism` flag binds to `cfg.Parallelism` for `bktec plan`, the same way it does for `bktec run`.

### Context

- Linear: [TE-5745](https://linear.app/buildkite/issue/TE-5745/circle-buildkite-test-engine-split-by-example-not-producing-example)

### Changes

- `cli.go`: append `parallelismFlag` to `planCommandFlags()`.
- `cli_test.go`: add `TestPlanCommandIncludesParallelismFlag` regression test.
- `CHANGELOG.md`: note the fix under Unreleased.

### Testing

- `go build ./...`
- `go test ./...` — `TestPlanCommandIncludesParallelismFlag` passes; existing tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - Remove Pact contract testing for the Test Engine API. Consumer tests now use plain `httptest` fixtures (no behaviour change for users); the `internal/api/pacts/` directory, `pact-go` dependency, and `bin/{publish-pact,release-pact-version,pact-record-support-ended}` scripts have been deleted.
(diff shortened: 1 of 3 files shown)


niceking:
@codex review

nprizal (approved):
Looks good to me

PR #502, opened 2026-05-11
Port Jest to buildcommand

malclocke:
Updates `runner.Jest` to use `buildCommand` in line with the other runners.

Removes the method `retryCommandNameAndArgs` method which was unique to the `Jest` runner, moving it's behaviour into `CommandNameAndArgs`.

### DIFF
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -4,7 +4,6 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"regexp"
 	"slices"
@@ -59,14 +58,10 @@ func (j Jest) GetFiles() ([]string, error) {
 }
 
 func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) error {
-	var cmd *exec.Cmd
-
-	commandName, commandArgs, err := j.CommandNameAndArgs(testCases, retry)
+	cmd, err := buildCommand(j, testCases, retry)
 	if err != nil {
-		return fmt.Errorf("failed to build command: %w", err)
+		return err
 	}
-
-	cmd = exec.Command(commandName, commandArgs...)
 
 	cmdErr := runAndForwardSignal(cmd)
 
--- a/internal/runner/jest.go
+++ b/internal/runner/jest.go
@@ -61,29 +61,12 @@ func (j Jest) GetFiles() ([]string, error) {
 func (j Jest) Run(result *RunResult, testCases []plan.TestCase, retry bool) error {
 	var cmd *exec.Cmd
 
-	testPaths := pathsFromTestCases(testCases)
-	slices.Sort(testPaths)
-	testPaths = slices.Compact(testPaths)
-
-	if !retry {
-		commandName, commandArgs, err := j.CommandNameAndArgs(testCases, false)
-		if err != nil {
-			return fmt.Errorf("failed to build command: %w", err)
-		}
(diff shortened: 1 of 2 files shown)


internal/runner/jest.go
```diff
@@ -191,58 +169,51 @@ func (j Jest) CommandNameAndArgs(testCases []plan.TestCase, retry bool) (string,
 	}
 
 	testPaths := pathsFromTestCases(testCases)
+	slices.Sort(testPaths)
```
malclocke, line 172: Moving this `slices.Sort` from the `Run` method into `CommandNameAndArgs` means it again sorts the paths when `retry` is `true` and when it is `false`.

internal/runner/jest.go
```diff
@@ -191,58 +169,51 @@ func (j Jest) CommandNameAndArgs(testCases []plan.TestCase, retry bool) (string,
...
+			words = append(words, testPaths...)
+		} else {
+			words = slices.Replace(words, idx, idx+1, testPaths...)
+		}
 	}
 
 	outputIdx := slices.Index(words, "{{resultPath}}")
 	if outputIdx < 0 {
-		err := fmt.Errorf("couldn't find '{{resultPath}}' sentinel in retry command, exiting")
+		err := fmt.Errorf("couldn't find '{{resultPath}}' sentinel in command, exiting")
```
malclocke, line 206: A tiny change here, this error message becomes the same for both retry and non retry cases.

internal/runner/jest.go
```diff
@@ -191,58 +169,51 @@ func (j Jest) CommandNameAndArgs(testCases []plan.TestCase, retry bool) (string,
...
+		}
 
-	escapedTestCases := make([]string, len(testCases))
-	for i, testCase := range testCases {
-		escapedTestCases[i] = regexp.QuoteMeta(testCase)
-	}
+		escapedTestCases := make([]string, len(testCases))
+		for i, testCase := range testNames {
+			escapedTestCases[i] = regexp.QuoteMeta(testCase)
+		}
```
gchan, line 185: Nit: Can we fold the `testCase` loop above into this loop? It would save us from setting `testNames`

```suggestion
		escapedTestCases := make([]string, len(testCases))
		for i, testCase := range testCases {
			escapedTestCases[i] = regexp.QuoteMeta(fmt.Sprintf("%s %s", testCase.Scope, testCase.Name))
		}
```

internal/runner/jest_test.go
```diff
@@ -351,7 +351,7 @@ func TestJestCommandNameAndArgs_WithInterpolationPlaceholder(t *testing.T) {
 	}
 
 	wantName := "jest"
-	wantArgs := []string{"spec/user.spec.js", "spec/billing.spec.js", "--outputFile", "jest.json"}
+	wantArgs := []string{"spec/billing.spec.js", "spec/user.spec.js", "--outputFile", "jest.json"}
```
gchan, line 354: Observation: test commands are now sorted.

gchan (approved):
LGTM, can merge as is.

internal/runner/jest_test.go
malclocke: There is no functional change here because the sort was previously done in the caller (`Jest.Run`). The sort has moved from `Run` to the method under test here (`CommandNameAndArgs`) so these tests need to be updated. 

There has been a bit of churn on this behaviour in a couple of recent PRs but basically the overall sorting is unchanged at runtime.

PR #507, opened 2026-05-12
Use OIDC when auth is missing

malclocke:
`BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN` is the token used by `bktec` to communicate to the test planning API.

`BUILDKITE_ANALYTICS_TOKEN` is used by the language specific test collector libraries to upload test execution results to Test Engine.

Previously these both had to be set manually to a user API token and a suite token respectively, but support has been added recently to allow a single short lived OIDC token generated by `buildkite-agent` to be used in place of both tokens.

With the changes in this PR, if either `BUILDKITE_TEST_ENGINE_API_ACCESS_TOKEN` or `BUILDKITE_ANALYTICS_TOKEN` are blank, they are populated with a temporary OIDC token generated via `buildkite-agent oidc request-token`. If both are missing the same OIDC token is used for both.

The token lifetime defaults to 2 hours but can be overridden with `--oidc-lifetime`.


internal/config/mock-buildkite-agent
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
```
malclocke, line 1: This script is only used by the test suite.

internal/config/validate.go
```diff
@@ -33,6 +37,17 @@ func (c *Config) validate() error {
 		}
 	}
 
+	if c.AccessToken == "" {
+		token, err := c.generateOidcToken()
```
malclocke, line 41: Implementing this in `config.validate` is the lowest touch approach AFAICT. This does mean that `bktec` will potentially be shelling out to `buildkite-agent` to fetch a token before the configuration has been completely validated.

It would be preferable IMO to generate a token "just in time" on the first use, but that would be a larger rework.

cli.go
```diff
@@ -154,11 +155,38 @@ var parallelismFlag = &cli.IntFlag{
...
+var uploadTokenFlag = &cli.StringFlag{
+	Name:        "upload-token",
+	Category:    "TEST ENGINE",
+	Usage:       "Buildkite collector upload token. If not set, an OIDC token will be generated using buildkite-agent",
+	Sources:     cli.EnvVars("BUILDKITE_ANALYTICS_TOKEN"),
+	Destination: &cfg.UploadToken,
+	Hidden:      true,
+}
+
+var oidcFlag = &cli.BoolWithInverseFlag{
```
malclocke, line 172: Adding a flag here as a kill switch for the OIDC token generation, in case it causes problems for users. The boolean flag `--oidc` defaults to `true`, but can be turned off with `--no-oidc`.

cli.go
```diff
@@ -332,6 +360,12 @@ var debugFlag = &cli.BoolFlag{
 	Destination: &cfg.DebugEnabled,
 }
 
+var buildkiteAgentCommandFlag = &cli.StringFlag{
```
malclocke, line 363: `bktec plan` uses `buildkite-agent` to run `buildkite-agent pipeline upload`. This flag allows the user to set the command to run `buildkite-agent`, but I haven't included the changes to make `bktec plan` honour the provided value as it requires a wee bit of rework in the tests.

So if a user runs `bktec --buildkite-agent-command /path/to/buildkite-agent`:

- OIDC tokens will be generated with `/path/to/buildkite-agent oidc request-token`
- Dynamic pipelines will be uploaded with `buildkite-agent pipeline upload`

We can leave this as is, or I can add support for the latter here or in a follow up at the reviewers discretion.

We could also make the flag hidden, as it's currently only used for setting a `mock-buildkite-agent` as the command when running tests.

gchan: Maybe make it a hidden flag for now until we make `bktec plan` respect this flag? Perhaps add a comment here for future travellers that `plan` doesn't honour this flag yet?

gchan: If we plan to not make it hidden, we may want to document it with `Category`, `Usage`, etc.

internal/config/validate.go
gchan: I'm feeling a bit nervous about the `validate` function modifying the configuration. However, there is already some precedent with the `Identifier` being synthesised from other fields.

gchan: JIT sounds great, perhaps a lazy `GetAcessToken` function on Config.
A non-JIT approach may include a 'resolve' step/function between parsing and validation.

internal/config/validate_test.go
```diff
@@ -566,3 +566,23 @@ func TestConfigValidate_TagFiltersOnlyWorksWithPytest(t *testing.T) {
 		}
 	})
 }
+
+func TestConfigValidate_OidcTokens(t *testing.T) {
```
gchan, line 570: Should we add a test which ensures `AccessToken` and `UploadToken` isn't overwritten when one or the other is missing?

internal/config/validate.go
```diff
@@ -90,6 +105,24 @@ func (c *Config) validate() error {
...
+			c.UploadToken = c.AccessToken
+		} else {
+			// If OIDC was *not* used to generate the bktec API access token then we need
+			// to generate a token for collector uploads.
+			token, err := c.generateOidcToken()
+
+			if err != nil {
+				c.errs.appendFieldError("BUILDKITE_ANALYTICS_TOKEN", "%v", err)
+			}
+			c.UploadToken = token
```
gchan, line 122:
```suggestion
			if err != nil {
				c.errs.appendFieldError("BUILDKITE_ANALYTICS_TOKEN", "%v", err)
			} else {
			   c.UploadToken = token
			}
```

internal/config/validate.go
```diff
@@ -90,6 +105,24 @@ func (c *Config) validate() error {
...
+	// Upload token could come from the env BUILDKITE_ANALYTICS_TOKEN, but may be blank ...
+	if c.UploadToken == "" {
+		if c.accessTokenIsOIDC {
+			// If OIDC was used to generate the bktec API access token then the same token
+			// can be used for collector uploads.
+			c.UploadToken = c.AccessToken
+		} else {
+			// If OIDC was *not* used to generate the bktec API access token then we need
+			// to generate a token for collector uploads.
+			token, err := c.generateOidcToken()
```
gchan, line 117: Do we want to shout loudly if "" is returned when `--no-oidc` is set and there is no upload token?

At the moment, I believe we will set `UploadToken` to "" and set the envar  BUILDKITE_ANALYTICS_TOKEN to ""

internal/config/validate.go
```diff
@@ -201,3 +234,26 @@ func (c *Config) ValidateForPlan() error {
...
+	lifetime := strconv.Itoa(int(c.OidcLifetime.Seconds()))
+	// Skipping a security linter check here. The issue is "G204: Subprocess launched with a potential tainted input or cmd arguments"
+	// Given that running tainted input commands is bktec's raison d'etre this is acceptable.
+	cmd := exec.Command(c.BuildkiteAgentCommand, "oidc", "request-token", "--audience", suiteUrl, "--lifetime", lifetime) //nolint:gosec
+	cmd.Stderr = &errorWriter
+	cmd.Stdout = &tokenWriter
+	cmd.Env = os.Environ()
+
+	if err := cmd.Run(); err != nil {
+		return "", fmt.Errorf("Error generating token: %s: %v", errorWriter.String(), err)
```
gchan: Nit: error strings are typically not capitalised (ST1005). I wonder if our linter is working? Or whether this is just an optional convention?

internal/config/validate.go
```diff
@@ -90,6 +105,24 @@ func (c *Config) validate() error {
 func (c *Config) ValidateForRun() error {
 	_ = c.validate()
 
+	// Upload token could come from the env BUILDKITE_ANALYTICS_TOKEN, but may be blank ...
+	if c.UploadToken == "" {
+		if c.accessTokenIsOIDC {
+			// If OIDC was used to generate the bktec API access token then the same token
+			// can be used for collector uploads.
+			c.UploadToken = c.AccessToken
```
gchan, line 113: This assumes `validate` is already called and `c.AccessToken` has been set.

internal/config/validate.go
```diff
@@ -201,3 +234,26 @@ func (c *Config) ValidateForPlan() error {
 
 	return nil
 }
+
+func (c *Config) generateOidcToken() (token string, err error) {
```
gchan, line 238: Nit: should we capitalise `Oidc` to `OIDC`?

gchan (approved):
Nice small OIDC integration. I've left a few questions, suggested corrections and minor nits but have approved this PR. 

I wonder if we should add tests to ensure the tokens are not overwritten when one is absent. 

I noted whether we should fail loudly when --no-oidc is set and there is no token provided.

I assumed we have discussed the OIDC lifetime internally and decided 2 hours is sufficient for most users.

internal/config/validate.go
malclocke: `c.UploadToken` is set via the CLI flag from `BUILDKITE_ANALYTICS_TOKEN`. So with `--no-oidc` the behaviour is the same as today - `BUILDKITE_ANALYTICS_TOKEN` will be blank for the runner if it's blank in the environment.

internal/config/validate.go
malclocke: We don't seem to have an agreed convention, so I'll leave as is for now.

```
internal/api/client_test.go:func TestHttpClient_AttachAccessTokenToRequest(t *testing.T) {

internal/command/plan_test.go:func TestPlanJSON(t *testing.T) {
```

internal/config/validate.go
gchan: Yep agree it's the current behaviour today. My previous comment was a thought whether we want to 'fail-fast'er if `BUILDKITE_ANALYTICS_TOKEN` is not set and avoid setting the env var to a blank string. 

Feel free to ignore as this feels a little beyond the scope of this change.

internal/config/validate.go
malclocke: It looks like this linter should be enabled by default, not sure why it's not. I'll look at the linter in a follow up.

malclocke: Linter is working, this particular line seems to be hitting some internal exclude list 🤷 

```diff
diff --git a/internal/config/validate.go b/internal/config/validate.go
index 336e908..06a7169 100644
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -252,7 +252,7 @@ func (c *Config) generateOidcToken() (token string, err error) {
        cmd.Env = os.Environ()
 
        if err := cmd.Run(); err != nil {
-               return "", fmt.Errorf("error generating token: %s: %v", errorWriter.String(), err)
+               return "", fmt.Errorf("Not a great error: %s: %v", errorWriter.String(), err)
        }
 
        return strings.TrimSpace(tokenWriter.String()), nil
```

```
$ staticcheck  ./...
internal/config/validate.go:255:14: error strings should not be capitalized (ST1005)
```

```diff
diff --git a/internal/config/validate.go b/internal/config/validate.go
index 336e908..6403011 100644
--- a/internal/config/validate.go
+++ b/internal/config/validate.go
@@ -252,7 +252,7 @@ func (c *Config) generateOidcToken() (token string, err error) {
        cmd.Env = os.Environ()
 
        if err := cmd.Run(); err != nil {
-               return "", fmt.Errorf("error generating token: %s: %v", errorWriter.String(), err)
+               return "", fmt.Errorf("Error can be capitalized for some reason: %s: %v", errorWriter.String(), err)
        }
 
        return strings.TrimSpace(tokenWriter.String()), nil
```

```
$ staticcheck  ./...
$ echo $?
0
```

cli.go
malclocke: Updated to hidden

internal/config/validate_test.go
malclocke: Tests added.

internal/config/validate.go
malclocke: @gchan we can enable [ST1003](https://staticcheck.dev/docs/checks/#ST1003) and add OIDC to the [list of initializms](https://staticcheck.dev/docs/configuration/options/#initialisms) if you like, LMK.

gchan: Sounds good to me, if you'd like, I can possibly do this.

PR #508, opened 2026-05-14
[TE-5834] Drop /v2/access-token preflight from backfill-commit-metadata

gchan:
## Description

`bktec tools backfill-commit-metadata` had an unconditional `GET /v2/access-token` preflight that broke OIDC JWTs minted by `buildkite-agent oidc request-token`. That endpoint is the user-token introspection endpoint -- it doesn't accept OIDC tokens, which authenticate against a separate suite-scoped policy. Every OIDC-authenticated invocation failed before doing any real work with `token scope check failed: verifying token: Authentication required`.

Drop the preflight and rely on the natural error path of the suite-scoped endpoints bktec has to call anyway. `FetchCommitList` fast-fails for `read_suites`; `PresignUpload`, moved up to run before the git work, fast-fails for `write_suites`. The presigned URL response is held and reused at the actual upload site. If S3 rejects the upload with 403, request a fresh presigned URL and retry once.

Considered HEAD probes (HEAD only mirrors GET per RFC 9110 §9.3.2; can't validate the POST presigned-upload endpoint) and OPTIONS probes (Buildkite has no `rack-cors`; OPTIONS requests don't get auth-validating handlers). Rejected both -- the natural-error approach is consistent with how the main `bktec` (test-splitting) command already handles auth errors via `DoWithRetry`'s `AuthError` / `ForbiddenError` types.

The S3 403-retry classifier is **status-only** (any 403 -> refresh and retry once), not a message-text match. Verified in a sandbox AWS account that the cases we'd want to exclude (policy condition mismatch, signature mismatch) return 400 Bad Request, so the status check alone keeps them out of the retry path. This avoids coupling to AWS's error message wording, which AWS doesn't version.

## Context

Resolves [TE-5834](https://linear.app/buildkite/issue/TE-5834). Blocks TE-5833 and its consumer PR ([buildkite/test-prediction#38](https://github.com/buildkite/test-prediction/pull/38)), which is paused on a bktec release containing this fix. Adjacent to TE-5787 and #507 (in-process OIDC mint for the main command path) -- different scope, no merge conflicts.

## Changes

Best reviewed commit-by-commit; each commit independently builds and passes tests:

- `[TE-5834] Add PresignedURLExpiredError to upload package` -- typed error returned when S3 responds 403 with the expired marker, so callers can distinguish signature expiry from generic permission failures. Self-contained, no caller change yet. (Type later renamed to `S3ForbiddenError` in the final commit.)
- `[TE-5834] Move PresignUpload to a preflight + held-response retry helper` -- structural prep. `PresignUpload` now runs before the git work and its response is held for reuse at the upload site, routed through a new helper that handles the typed-error retry. `VerifyTokenScopes` is still in place at this point; both preflights run.
- `[TE-5834] Drop /v2/access-token preflight, rely on natural error path` -- the auth-model change. Deletes `internal/api/verify_token_scopes.go`, prunes 12 `/v2/access-token` test mocks, replaces three scope-check assertions with natural-error-path tests, and adds regression tests for the no-preflight contract and the `--output` no-upload path.
- `[TE-5834] Note backfill-commit-metadata OIDC support in CHANGELOG`.
- `[TE-5834] Match real S3 'Policy expired.' response, not the PUT shape` -- sandbox-verified matcher fix. The initial fixture used the presigned-PUT message; bktec uses POST, where the real response is different. Updates the matcher, source comments, and test fixtures to the literal sandbox responses for both the expiry case (403) and the boundary case (policy condition mismatch is 400, not 403).
- `[TE-5834] Classify any 403 as retryable, drop the message-text matcher` -- responding to review. The narrow message-text matcher coupled bktec to AWS's error wording, which AWS doesn't version. Broaden to "any 403 retries once". Renames `PresignedURLExpiredError` -> `S3ForbiddenError` and `uploadWithRefreshOnExpiry` -> `uploadWithRetryOn403` for accuracy. Softens the operator stderr message to "S3 rejected the upload (403); requesting a fresh presigned URL and retrying...".

## Verification

Backed by specs. Locally:

```bash
go test ./internal/api/... ./internal/command/... ./internal/upload/...
golangci-lint run ./internal/api/... ./internal/command/... ./internal/upload/...
```

The S3 response shapes the matcher and boundary tests use have been verified end-to-end in a Buildkite sandbox AWS account:

1. Signed a presigned POST with a 5-second policy expiry, slept past it, attempted the upload -- captured the literal 403 / `Policy expired.` response that the retry-success integration test uses.
2. Signed a presigned POST with a `key` condition, submitted with a different key -- captured the 400 / `Policy Condition failed` response that the boundary test uses. This is the case the status-only matcher must not retry on.
3. Signed a presigned POST, tampered with the signature field -- confirmed 400 (also falls through the matcher's 403 gate).
4. Posted to a nonexistent bucket -- confirmed 404 (also falls through).

The boundary tests pin against the literal sandbox response bodies, so a future change to either the matcher or the response shapes from real S3 surfaces as a test failure rather than a silent regression in production.

Once this lands and a bktec release is cut, TE-5833's PR #38 bumps `BKTEC_VERSION` in `.buildkite/steps/run_bk_bk_backfill.sh` and verifies end-to-end against the live `test-prediction-backfill` pipeline with an OIDC JWT.

AI assisted.

## Deployment

Low risk. The auth model change is bktec-side only -- no server-side coordination required. Classic API tokens continue to work unchanged because both `FetchCommitList` and `PresignUpload` were already in the flow; they're just doing the auth validation work the preflight used to do. Missing-scope users get a slightly different error message (surfaced from the failing API call rather than the preflight) but fail at the same point in the flow, with the same actionable content.

## Rollback

Yes. Single-PR revert is sufficient; no schema, no feature flag, no downstream coupling beyond the deferred TE-5833 release bump.


internal/command/backfill_commit_metadata_test.go
```diff
@@ -527,31 +527,23 @@ func TestBackfillCommitMetadata_OutputSkipsPresignPreflight(t *testing.T) {
...
 			w.Write([]byte(`<?xml version="1.0" encoding="UTF-8"?>
-<Error>
-  <Code>AccessDenied</Code>
-  <Message>Request has expired</Message>
-  <Expires>2026-05-14T10:00:00Z</Expires>
-  <ServerTime>2026-05-14T11:30:00Z</ServerTime>
-  <RequestId>EXAMPLE000</RequestId>
-  <HostId>EXAMPLE</HostId>
-</Error>`))
+<Error><Code>AccessDenied</Code><Message>Invalid according to Policy: Policy expired.</Message><RequestId>P7XG7F93RENQV3B6</RequestId><HostId>lZUQ47E9D7XbgAiipMfWIbgJj4WVl1/34BihhSMiDaHGEAjE0FtKZ9QAh4psrJ8S1tFRvaGcoBeuEUnICTnXu9c6uOWe7ELO</HostId></Error>`))
```
gchan, line 493: Verified in a sandbox account: https://linear.app/buildkite/issue/TE-5834/drop-v2access-token-preflight-from-bktec-tools-backfill-commit#comment-a7d856a4

gchan: This feels brittle, perhaps we could just retry once on any 403 errors?

gchan: https://github.com/buildkite/test-engine-client/pull/508/commits/051672861c6205947bf0dd59ee144f9ceeff25ce

gchan:
Verified the backfill command `backfill-commit-metadata` and the options --upload, --output all are functional.

internal/command/backfill_commit_metadata.go
```diff
@@ -23,6 +24,17 @@ import (
...
+// Auth model (TE-5834): bktec does not preflight token scopes via the
+// user-token introspection endpoint. That endpoint (GET /v2/access-token) does
+// not accept OIDC JWTs, so preflighting there breaks the agent-OIDC auth path.
+// Instead, this command relies on the natural error path of the suite-scoped
+// endpoints it has to call anyway: FetchCommitList fast-fails for read_suites
+// + suite policy, and PresignUpload (run before the git work) fast-fails for
+// write_suites + suite policy. The PresignUpload response is held and reused
+// at the actual upload site; if S3 rejects the upload because the presigned
+// URL has expired, a fresh presigned URL is fetched and the upload is retried
+// once.
```
malclocke: **question** Do we need a comment saying what the code isn't doing? 🤔

gchan: No, I don't I don't think it's particularly useful for humans (too much context). I'm going to remove it.

internal/command/backfill_commit_metadata.go
```diff
@@ -37,33 +49,12 @@ func BackfillCommitMetadata(ctx context.Context, cfg *config.Config, runner git.
...
-	}
-	debug.Println("Token scopes verified")
-
-	// 3. Fetch commit list from server
+	// 2. Fetch commit list from server.
+	// This is the first auth-validating call. Missing read_suites, an
+	// un-admitted OIDC pipeline, or a bad suite slug all surface here as a
+	// typed API error before any git work runs. This preserves the fast-fail
+	// UX that the dropped /v2/access-token preflight used to provide for the
+	// read-scope half of the check.
```
gchan: Drop this coment

internal/command/backfill_commit_metadata.go
```diff
@@ -76,6 +67,28 @@ func BackfillCommitMetadata(ctx context.Context, cfg *config.Config, runner git.
 		return nil
 	}
 
+	// 3. Preflight the upload by requesting a presigned URL up front.
+	// This is the write-scope half of what the dropped /v2/access-token
```
gchan: We don't need to talk about the dropped /v2/access-token preflight. Just keep why we are making this request before the git operation.

internal/command/backfill_commit_metadata_test.go
```diff
@@ -127,11 +132,46 @@ func TestBackfillCommitMetadata_HappyPath(t *testing.T) {
 	}
 }
 
+// TestBackfillCommitMetadata_NoAccessTokenPreflight pins the TE-5834 contract
+// that BackfillCommitMetadata never calls GET /v2/access-token. That endpoint
+// is the user-token introspection endpoint; calling it breaks OIDC JWTs minted
+// by agent.buildkite.com, which are authenticated against a separate
+// suite-scoped policy and would 401 here. Asserting the negative explicitly
+// prevents a future refactor from re-introducing the preflight.
```
gchan: Testing that we don't make a request to /v2/access-token seems excessive.

internal/command/backfill_commit_metadata_test.go
```diff
@@ -353,73 +378,132 @@ func TestBackfillCommitMetadata_Upload(t *testing.T) {
...
+	// would indicate the held-response optimisation has regressed.
+	if got := atomic.LoadInt32(&presignHits); got != 1 {
+		t.Errorf("expected 1 PresignUpload call (held URL reused on happy path), got %d", got)
+	}
 }
 
-func TestBackfillCommitMetadata_ScopeCheckFails(t *testing.T) {
+// TestBackfillCommitMetadata_WriteScopeMissing_FailsAtPreflight pins TE-5834's
+// fast-fail UX for the write-scope half of the check. Before this ticket, a
+// missing write_suites scope was caught by the GET /v2/access-token preflight.
```
gchan: Don't mention the history of GET /v2/access-token preflight, just state the intention and purpose of this test.

internal/command/backfill_commit_metadata_test.go
```diff
@@ -353,73 +378,132 @@ func TestBackfillCommitMetadata_Upload(t *testing.T) {
...
+	// missing write_suites fails before any git command runs. If this drops
+	// below the assertion, the preflight has regressed back to running after
+	// the git work.
+	if got := atomic.LoadInt32(&gitRunnerCalls); got != 0 {
+		t.Errorf("expected 0 git commands before preflight failure, got %d", got)
+	}
 }
 
-func TestBackfillCommitMetadata_ScopeCheckWarnsWithOutput(t *testing.T) {
+// TestBackfillCommitMetadata_ReadScopeMissing_FailsAtFetchCommitList pins
```
gchan: Don't mention the history of GET /v2/access-token preflight, just state the intention and purpose of this test.

internal/command/backfill_commit_metadata_test.go
```diff
@@ -353,73 +378,132 @@ func TestBackfillCommitMetadata_Upload(t *testing.T) {
...
 	}
 }
 
-func TestBackfillCommitMetadata_ScopeCheckFailsWithOutputMissingReadSuites(t *testing.T) {
+// TestBackfillCommitMetadata_OutputSkipsPresignPreflight pins the contract
+// that --output (write-tarball-locally) does not call PresignUpload. The
+// preflight is only relevant when bktec is about to upload; a user writing
+// locally has explicitly opted out of the upload step and should not be
+// blocked by a missing write_suites scope. The dropped /v2/access-token
+// preflight used to warn-only in this case; now we don't check at all.
```
gchan: Don't mention the history of GET /v2/access-token preflight, just state the intention and purpose of this test.

internal/command/backfill_commit_metadata.go
malclocke: Again, seems weird to add comments about removed code.

internal/command/backfill_commit_metadata_test.go
```diff
@@ -591,15 +775,24 @@ func TestBackfillCommitMetadata_UploadOnly_MissingSuiteSlugFailsBeforeNetwork(t
 	}
 }
 
-func TestBackfillCommitMetadata_UploadOnly_ScopeCheckFails(t *testing.T) {
+// TestBackfillCommitMetadata_UploadOnly_WriteScopeMissing_FailsAtPresignUpload
+// pins TE-5834's fast-fail UX for the upload-only path. Before this ticket,
+// missing write_suites was caught by the GET /v2/access-token preflight; now
```
gchan: Don't mention the history of GET /v2/access-token preflight, just state the intention and purpose of this test.

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # Changelog
 
 ## Unreleased
+- `bktec tools backfill-commit-metadata` no longer preflights token scopes via the user-token introspection endpoint (`GET /v2/access-token`), so it now works with OIDC JWTs minted by `buildkite-agent oidc request-token` against suite-scoped policies. The fast-fail UX for missing scopes is preserved: missing `read_suites` surfaces from `FetchCommitList`, and missing `write_suites` surfaces from a `PresignUpload` call that now runs as an explicit preflight before the git work. If the held presigned URL expires before the upload runs, bktec automatically requests a fresh URL and retries the upload once.
```
malclocke: **question** do we want this in the changelog?

malclocke (approved):

CHANGELOG.md
gchan: I'm planning to make a new release 2.5.1 release so something _needs_ to be present. But this LLM output is TMI. I'll update to put emphasis that the `backfill-commit-metadata` command now support OIDC tokens.

gchan: https://github.com/buildkite/test-engine-client/pull/508/changes/32b8802edb66da4bc1921da15bf27f9e27ea0937

PR #510, opened 2026-05-15
[TE-5843] Encode v2 major version in go.mod module path

gchan:
## Description

The `go.mod` in this repo declares `module github.com/buildkite/test-engine-client` at every v2.x tag (v2.0.0 through v2.5.0 and `main`). Go's [module versioning rules](https://go.dev/ref/mod#major-version-suffixes) require modules at v2+ to encode the major version in the module path, so v2.x tags should read `module github.com/buildkite/test-engine-client/v2`.

The current state makes v2.x releases unfetchable via `go get` and uninstallable as a `go tool`. The `+incompatible` escape hatch doesn't apply because that only works for modules without a `go.mod` file:

```
$ go get github.com/buildkite/test-engine-client@v2.5.0
go: ... invalid version: module contains a go.mod file, so module path must
    match major version ("github.com/buildkite/test-engine-client/v2")

$ go get github.com/buildkite/test-engine-client/v2@v2.5.0
go: ... invalid version: module contains a go.mod file, so module path must
    match major version ("github.com/buildkite/test-engine-client/v2")
```

Downstream Go projects can install bktec as a tool (via the `tool ( ... )` directive in `go.mod`) only against v1.x. `buildkite/agent` is pinned to `v1.6.0` for exactly this reason — that's the most recent v1.x release, and predates the `plan` subcommand (v2.0.0) and `--collect-git-metadata` (v2.4.0).

This PR:

1. Updates `go.mod` to `module github.com/buildkite/test-engine-client/v2`.
2. Rewrites the 56 internal import paths in `*.go` files from `github.com/buildkite/test-engine-client/internal/...` to `github.com/buildkite/test-engine-client/v2/internal/...`.
3. Updates the version ldflags path in `.goreleaser.yaml` from `.../internal/version.Version` to `.../v2/internal/version.Version` so release builds continue to stamp `internal/version.Version` correctly.

Once a v2.x tag is cut on top of this commit, `go get github.com/buildkite/test-engine-client/v2@vX.Y.Z` resolves and downstream `go tool` setups work. Historical v2.0.0–v2.5.0 tags don't need to move; downstream consumers should pin to the new tag (e.g. `v2.5.1`) and update their import paths to include the `/v2` suffix.

## Context

Resolves [TE-5843](https://linear.app/buildkite/issue/TE-5843/fix-test-engine-client-gomod-module-path-for-v2x-blocks-go-get-go-tool).

Surfaced in [TE-5842](https://linear.app/buildkite/issue/TE-5842) while trying to unify two bktec call sites in `buildkite/agent`:

- `.buildkite/steps/tests.sh` uses `go tool test-engine-client` against `v1.6.0` from `go.mod`.
- `.buildkite/steps/collect-commit-metadata.sh` (landed via [TE-5828](https://linear.app/buildkite/issue/TE-5828), [agent#3924](https://github.com/buildkite/agent/pull/3924)) downloads the `v2.5.0` binary from the GitHub release at step time.

## Changes

Single commit:

- `[TE-5843] Encode v2 major version in go.mod module path`

## Verification

Backed by the existing test suite. The internal import rewrite is mechanical (`sed`-driven) so the diff is large but uniform.

```bash
go build ./...
go vet ./...
go test ./...
```

`go build` and `go vet` are clean. `go test` is clean on every package except `internal/runner`, which has the same pre-existing local-env failures on `main` (missing `cypress`, `gotestsum`, and `cucumber` binaries) — not regressions from this change.

AI assisted.

## Deployment

Low risk. No behavioural change — only the module identifier and import paths move. The compiled binary is byte-equivalent modulo the `internal/version.Version` ldflag string, which is rewritten to match the new path in the same change.

## Unblocks

- [TE-5842](https://linear.app/buildkite/issue/TE-5842) — bump test-engine-client in `buildkite/agent` and unify both call sites on `go tool` once `v2.5.1` (or later) is tagged on top of this commit.

## Rollback

Yes. Revert the merge; downstream consumers haven't been migrated yet because no v2.x tag exists with the corrected path.


pda:
@codex review

pda (approved):
🚀 assuming Codex review doesn't spot anything unexpected.

PR #515, opened 2026-05-21
Add unit test to ensure run command env vars bind to config

nprizal:
### Description

We had a regression where `BUILDKITE_TEST_ENGINE_TEST_CMD` (and its `--test-command` flag) lost its binding to `cfg.TestCommand`, meaning custom test commands were silently ignored. That was fixed as a hotfix, but we had no test to prevent it happening again.

This adds `TestRunCommandEnvVarsBindToConfig` which runs the `bktec run` command with a no-op action and verifies that all 28 env vars wired to `runCommandFlags()` actually land in the `cfg` struct. If anyone accidentally drops a `Destination` field from a flag definition in the future, the test will catch it.

### Context

[TE-5180](https://linear.app/buildkite/issue/TE-5180)

### Testing

The new test covers all env var → `cfg` field bindings for the `run` command, including the inversion logic for `BUILDKITE_TEST_ENGINE_DISABLE_RETRY_FOR_MUTED_TEST` and the `BoolWithInverseFlag` binding for `BUILDKITE_TEST_ENGINE_OIDC`.


nprizal:
@buildsworth-bk re-review this PR

PR #524, opened 2026-05-27
Change pytest runner default output to JUnit; keep test-collector JSON as an option

malclocke:
Closes https://linear.app/buildkite/issue/TE-5956

## Summary

- When `buildkite-test-collector` is installed, keep existing behavior: use `--json={{resultPath}}` and parse Test Engine JSON output
- When not installed, fall back to `--junit-xml={{resultPath}}` and parse via the shared JUnit XML parser
- Logs a message indicating which path was taken so customers can see the behavior
- `--tag-filters` requires the plugin and exits with a clear error if used without it
- Adds `pytestNodeIDFromJUnit` to reconstruct pytest node IDs (scope + path) from JUnit `classname`/`name` attributes

## Test plan

- [ ] `TestPytestRun_JUnit_TestPassed` — JUnit path parses a passing result file
- [ ] `TestPytestRun_JUnit_TestFailed` — JUnit path parses a failing result file and maps node IDs correctly
- [ ] `TestPytestNodeIDFromJUnit` — unit tests for classname → node ID conversion (flat, nested dirs, class methods)
- [ ] Existing JSON-path tests (`TestPytestRun`, `TestPytestRun_TestFailed`, etc.) still pass
- [ ] Full suite: `go test ./...`

### DIFF
--- a/docs/pytest.md
+++ b/docs/pytest.md
@@ -1,19 +1,32 @@
 # Using bktec with pytest
-To integrate bktec with pytest, you need to [install and configure Buildkite Test Collector for pytest](https://buildkite.com/docs/test-engine/python-collectors#pytest-collector) first. Then set the `BUILDKITE_TEST_ENGINE_TEST_RUNNER` environment variable to `pytest`.
+Set the `BUILDKITE_TEST_ENGINE_TEST_RUNNER` environment variable to `pytest` to use bktec with pytest.
 
 ```sh
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=pytest
 bktec run
 ```
 
-## Configure test command
-By default, bktec runs pytest with the following command:
+bktec works with pytest in two modes depending on whether [Buildkite Test Collector for pytest](https://buildkite.com/docs/test-engine/python-collectors#pytest-collector) is installed:
 
+- **With `buildkite-test-collector`** (recommended): bktec uses the collector's JSON output for richer test result data and supports `--tag-filters`.
+- **Without `buildkite-test-collector`**: bktec falls back to JUnit XML output. This is sufficient for basic parallelisation and retry, but `--tag-filters` is not available.
+
+bktec logs which mode it is using on startup.
+
+## Configure test command
+By default, bktec runs pytest with one of the following commands depending on which mode is active:
+
+**With `buildkite-test-collector`:**
 ```sh
 pytest {{testExamples}} --json={{resultPath}}
 ```
 
-In this command, `{{testExamples}}` is replaced by bktec with the list of test files or tests to run, and `{{resultPath}}` is replaced with a unique temporary path created by bktec. `--json` option is a custom option added by Buildkite Test Collector to save the result into a JSON file at given path. You can customize this command using the `BUILDKITE_TEST_ENGINE_TEST_CMD` environment variable. 
+**Without `buildkite-test-collector` (JUnit fallback):**
+```sh
+pytest {{testExamples}} --junit-xml={{resultPath}}
+```
+
+In both commands, `{{testExamples}}` is replaced by bktec with the list of test files or tests to run, and `{{resultPath}}` is replaced with a unique temporary path created by bktec. You can customize this command using the `BUILDKITE_TEST_ENGINE_TEST_CMD` environment variable.
 
 To customize the test command, set the following environment variable:
 ```sh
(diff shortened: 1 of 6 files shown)


internal/runner/pytest.go
```diff
@@ -85,12 +99,16 @@ func (p Pytest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
 		return cmdErr
 	}
 
-	tests, parseErr := parseTestEngineTestResult(p.ResultPath)
+	if p.useJUnit {
+		return p.runParseJUnit(result, cmdErr)
+	}
+	return p.runParseJSON(result, cmdErr)
```
nprizal: [nit] can we avoid passing `cmdErr`?

```
if p.useJunit {
   parseError := p.runParseJunit(result)
} else {
   parseError := p.runParseJson(result)
}

if parseErr != nil {
   fmt.Println("....")
}

return cmdErr
```

internal/runner/pytest.go
```diff
@@ -85,12 +99,16 @@ func (p Pytest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
 		return cmdErr
 	}
 
-	tests, parseErr := parseTestEngineTestResult(p.ResultPath)
+	if p.useJUnit {
```
gchan, line 120: 🤔 I wonder if we should look at `c.TestCommand` as a heuristic?

It is possible for a user to not use the collector but set a `--json=` command OR use the collector but set a `--junit-xml` command. Possibly a rare case.

docs/pytest.md
```diff
@@ -1,27 +1,40 @@
...
+In both commands, `{{testExamples}}` is replaced by bktec with the list of test files or tests to run, and `{{resultPath}}` is replaced with a unique temporary path created by bktec. You can customize this command using the `BUILDKITE_TEST_ENGINE_TEST_CMD` environment variable.
 
 To customize the test command, set the following environment variable:
 ```sh
 export BUILDKITE_TEST_ENGINE_TEST_CMD="pytest --cache-clear --json={{resultPath}} {{testExamples}}"
 ```
 
 > [!IMPORTANT]
-> Make sure to include `--json={{resultPath}}` in your custom test command, as bktec requires this to read the test results for retries and verification purposes.
+> Make sure to include the appropriate result flag (`--json={{resultPath}}` or `--junit-xml={{resultPath}}`) in your custom test command, as bktec requires this to read the test results for retries and verification purposes.
```
nprizal: What happen if test-collector is installed, but the custom command use `--junit-xml` instead?

internal/runner/pytest.go
nprizal: +1 for this. But we should only parse json when the test collector is installed and the `test-cmd` has `--json`. We can't parse json that is not generated by the test collector. The later case is legit, test collector in installed, but the command uses `--junit-xml`.

internal/runner/pytest.go
```diff
@@ -106,10 +124,66 @@ func (p Pytest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
...
+
+	// Everything up to the first uppercase component is the module (file) path.
+	// Class names in Python use CamelCase; module/directory names use snake_case.
+	moduleEnd := len(parts)
+	for i, p := range parts {
+		if len(p) > 0 && p[0] >= 'A' && p[0] <= 'Z' {
+			moduleEnd = i
+			break
+		}
+	}
```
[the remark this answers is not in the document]
nprizal: the second sound like a valid case

file path: `tests/MyTest/test_substract.py`

junit
```
    <testcase classname="tests.MyTest.test_substract.TestSubstract" name="test_positive"
      time="0.000" />
```

json:
```
  {
    "scope": "tests/MyTest/test_substract.py::TestSubstract",
    "name": "test_positive",
    "location": "tests/MyTest/test_substract.py:1",
    "file_name": "tests/MyTest/test_substract.py",
    ...
  },
```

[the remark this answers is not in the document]
nprizal: But the logic should works for above case, can we add a test case for that?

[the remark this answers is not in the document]
nprizal: Ah yes, that logic wont work for this

```
<testcase classname="tests.MyTest.test_add" name="test_add" time="0.000" />
```

json:
```
  {
    "scope": "tests/MyTest/test_add.py",
    "name": "test_add",
    "location": "tests/MyTest/test_add.py:2",
    "file_name": "tests/MyTest/test_add.py",
     ...
  }
```

internal/runner/pytest.go
```diff
@@ -59,6 +57,20 @@ func NewPytest(c RunnerConfig) Pytest {
 		}
 	}
 
+	// Detect output format from the result flag in the command, so a custom command
+	// with --junit-xml works even when buildkite-test-collector is installed (and vice versa).
+	if strings.Contains(c.TestCommand, "--junit-xml") {
+		useJUnit = true
+	} else if strings.Contains(c.TestCommand, "--json=") {
+		useJUnit = false
```
nprizal, line 70: we should return an error here when `--json` is set, but no test collector is installed. So we don't attempt to parse json generated by other json reporter

internal/runner/pytest_test.go
```diff
@@ -463,6 +463,20 @@ func TestPytestNodeIDFromJUnit(t *testing.T) {
 			wantScope: "test_something",
 			wantPath:  "test_something",
 		},
+		{
```
nprizal, line 466: 👍

nprizal (approved):
Looks good to me. One issue is that the existing JUnit parser doesn't handle `<error>` element at the moment. But it is out of scope of this PR. I've created a ticket for that.

PR #527, opened 2026-05-28
Upload pytest test results to Upload API

malclocke:
## Summary

- Implements `ResultFormat()` on the `Pytest` runner to hook into the existing `uploadResults` flow in `run.go`
- Returns `"junit"` when using JUnit XML output (the default), triggering an upload to the Upload API
- Returns `""` when using JSON output (buildkite-test-collector plugin), skipping the upload to avoid duplicates — the collector handles it itself

Closes TE-5957
Depends on TE-5956 (already merged)

## Test plan

- [ ] `TestPytestResultFormat_JUnit` — verifies `"junit"` is returned when `useJUnit: true`
- [ ] `TestPytestResultFormat_JSON` — verifies `""` is returned when `useJUnit: false`

🤖 Generated with [Claude Code](https://claude.ai/claude-code)

### DIFF
--- a/internal/runner/pytest.go
+++ b/internal/runner/pytest.go
@@ -90,6 +90,13 @@ func NewPytest(c RunnerConfig) Pytest {
 	}
 }
 
+func (p Pytest) ResultFormat() string {
+	if p.useJUnit {
+		return "junit"
+	}
+	return ""
+}
+
 func (p Pytest) SupportedFeatures() SupportedFeatures {
 	return SupportedFeatures{
 		SplitByFile:     true,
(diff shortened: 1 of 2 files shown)


internal/runner/pytest.go
```diff
@@ -90,6 +90,13 @@ func NewPytest(c RunnerConfig) Pytest {
 	}
 }
 
+func (p Pytest) ResultFormat() string {
+	if p.useJUnit {
+		return "junit"
+	}
+	return ""
```
nprizal, line 97: 👍

nprizal (approved):

PR #531, opened 2026-06-04
[PS-2065] Fall back to mute when skip is unsupported

gchan:
> **Demonstration PR, not for merge.** Opened as a draft to show this approach as an option for PS-2065 / PS-2061.

For runners that can mute but cannot skip (e.g. Jest), bktec now falls back to muting the tests Test Engine wanted to skip, so their failures no longer fail the build.

Test Engine excludes skipped tests from a node's task list. File-splitting runners like Jest still receive the whole file and run every test in it, including the ones marked for skipping. Jest has no way to skip an individual test, so the test executes and its failure fails the build — the behaviour Highspot hit in PS-2061.

### Decisions Made

**Mute rather than re-run-and-discard.** Muting is matched by scope and name and reuses the existing muted-failure suppression (`OnlyMutedFailures`), so no new result-handling path is introduced. The skipped tests are left in the task list and simply registered as muted; they were already running.

**Runners that genuinely support skip are untouched.** RSpec splits by example and never receives the skipped cases, so the fallback only applies when `SupportedFeatures{Skip: false, Mute: true}`.

## Context

Linear: PS-2065, related to PS-2061

## Verification

`go test ./internal/command/` — backed by specs. New test `TestSkipToMuteFallback_FailedSkippedTestDoesNotFailBuild` asserts a failing skipped-then-muted test yields `RunStatusPassed` with the failure recorded as muted. AI assisted.

## Rollback

Yes. Single self-contained change behind a runner-capability check; reverting the commit restores prior behaviour.

### DIFF
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -60,6 +60,22 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 	// get plan for this node
 	thisNodeTask := testPlan.Tasks[strconv.Itoa(cfg.NodeIndex)]
 
+	// When a runner can mute tests but cannot skip them, fall back to muting the
+	// tests Test Engine wanted to skip.
+	//
+	// Test Engine excludes skipped tests from a node's task list, but runners that
+	// split by file (e.g. Jest) still receive the whole file and run every test in
+	// it, including the ones marked for skipping. Such runners have no way to skip
+	// an individual test, so the skipped test executes anyway and its failure would
+	// fail the build. Treating it as muted lets it run while suppressing its
+	// failure. Muting is matched by scope and name, so the skipped test is muted
+	// regardless of which file or node it belongs to.
+	features := testRunner.SupportedFeatures()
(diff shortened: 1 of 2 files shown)


internal/command/run.go
```diff
@@ -60,6 +60,22 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
...
+	//
+	// Test Engine excludes skipped tests from a node's task list, but runners that
+	// split by file (e.g. Jest) still receive the whole file and run every test in
+	// it, including the ones marked for skipping. Such runners have no way to skip
+	// an individual test, so the skipped test executes anyway and its failure would
+	// fail the build. Treating it as muted lets it run while suppressing its
+	// failure. Muting is matched by scope and name, so the skipped test is muted
+	// regardless of which file or node it belongs to.
+	features := testRunner.SupportedFeatures()
+	if !features.Skip && features.Mute && len(testPlan.SkippedTests) > 0 {
```
gchan, line 74: I don't think that actually works as `testPlan.SkippedTests` wouldn't have the list of skipped tests.

PR #540, opened 2026-06-10
Treat go test build failures as terminal; don't retry

pda:
### Description

When a Go package fails to compile, gotestsum reports it in JUnit XML as a synthetic testcase named `TestMain` with an empty classname. bktec recorded that as a retryable test failure, but the empty classname means an empty package path, so it was silently dropped from the retry command. If the remaining failures were flaky and passed on retry, the final attempt exited 0 and the job passed despite the compile error — even though bktec's own report said tests had failed.

Two changes:

- The gotest runner detects the build-failure signature (`TestMain` + empty classname + `[build failed]` in the failure text) and records it as an error outside of the tests, using the same `RunResult.error` mechanism rspec, jest, and playwright already use for things like syntax errors.
- The retry loop treats an error outside of the tests as terminal and returns the current attempt's command error instead of retrying. Retrying can't fix such errors, and a passing retry could previously mask them — this also closes the same latent hole for the rspec/jest/playwright error conditions when they coexist with retryable failures.

The `[build failed]` content check distinguishes build failures from other package-level failures that share the same testcase shape (a crashed `TestMain`, test timeouts); those keep their existing behaviour, as does `[setup failed]`. The string is stable: it's been a literal in cmd/go since at least Go 1.5 and is asserted byte-for-byte by Go's own script tests, and gotestsum's synthesis is unchanged since before v1.8.0.

Out of scope for now: the broader question of deriving bktec's final exit status from the run result, so that *any* failed test that never makes it into the final attempt fails the step. This closes the build-failure instance of that class.

### Testing

Added `testdata/go/broken/`, a deliberately uncompilable package, so the regression tests exercise the real go toolchain and gotestsum rather than canned XML — if a future Go or gotestsum version changes the build-failure output, these tests fail at version-bump time rather than in a user's pipeline.

Tests were written first and confirmed to fail. The loop-level test reproduces the original incident shape (a broken package alongside a genuinely failing test): before the fix, the retry visibly drops the broken package and re-runs only the other package.

### DIFF
--- a/internal/runner/gotest.go
+++ b/internal/runner/gotest.go
@@ -81,6 +81,19 @@ func (g GoTest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
 	}
 
 	for _, test := range testResults {
+		// A package that fails to build is reported by gotestsum as a synthetic
+		// testcase named "TestMain" with an empty classname, whose failure
+		// message contains "[build failed]" (printed by the go tool itself).
+		// A build failure cannot be fixed by retrying, so treat it as an error
+		// outside of the tests instead of recording it as a test result.
+		// Recording it would select it for retry, where its empty package path
+		// would silently drop it from the retry command, allowing the build
+		// failure to go unnoticed.
+		if isBuildFailure(test) {
+			result.error = fmt.Errorf("go test failed to build %s", test.SuiteName)
+			continue
+		}
+
 		result.RecordTestResult(plan.TestCase{
 			Format: plan.TestCaseFormatExample,
 			Scope:  test.Classname,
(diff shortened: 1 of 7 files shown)


internal/runner/gotest.go
```diff
@@ -81,6 +81,19 @@ func (g GoTest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
...
+		// A package that fails to build is reported by gotestsum as a synthetic
+		// testcase named "TestMain" with an empty classname, whose failure
+		// message contains "[build failed]" (printed by the go tool itself).
+		// A build failure cannot be fixed by retrying, so treat it as an error
+		// outside of the tests instead of recording it as a test result.
+		// Recording it would select it for retry, where its empty package path
+		// would silently drop it from the retry command, allowing the build
+		// failure to go unnoticed.
+		if isBuildFailure(test) {
+			result.error = fmt.Errorf("go test failed to build %s", test.SuiteName)
```
gchan, line 93: Noting the last-write-wins behavour: If there are multiple build failures, only the last failure will be reported. 

Also noting result.error is sticky which is fine as the run is terminal.

gchan (approved):

nprizal (approved):
Make sense to me

PR #543, opened 2026-06-11
Retry uploading test results on transient failures (take 2)

nprizal:
### Description

This re-applies #538, which added retry logic to `UploadTestResults` but had to be reverted in #541.

The original goal is unchanged: add retry logic to `UploadTestResults` so transient failures are retried before giving up.

### Why it was reverted

#538 changed how the response body is read. It made `doWithRetry` return the response so the caller reads the body, where before the body was read inside `doWithRetry`. The per-attempt `context.WithTimeout` was already there, but **now the read happened after that context had been cancelled**. For small buffered responses the read still succeeded from the transport buffer, so tests passed. Against a live connection though, the read failed and `bktec` exited with `1` on test plan calls:

> ❌ Unexpected error: reading response body: context canceled

### What's different this time

Each attempt is now bounded by `httpClient.Timeout` instead of a per-attempt request context. Go keeps that timer running after `Do` returns and only cleans it up when the body is closed, so the caller can safely read the response body. I also added a regression test that streams the response body in chunks with a flush and a delay, forcing the read to happen against a live connection so the old bug would fail the test.

### Changes

- Re-apply #538.
- Replace the per-attempt `context.WithTimeout` with `httpClient.Timeout`. Both cap how long a single attempt can take, so the timeout behavior is the same. The difference is that `httpClient.Timeout` keeps running until the body is closed instead of being cancelled when the attempt returns, so the caller can still read the response body.

### Testing

- `go test ./internal/api/...` passes.
- Added a streamed-response-body regression test for the revert bug, plus retry-then-succeed and no-retry-on-4xx tests for the upload path.

### DIFF
--- a/internal/api/client.go
+++ b/internal/api/client.go
@@ -140,13 +140,24 @@ type httpRequest struct {
 	Body   any
 }
 
-// DoWithRetry sends http request with retries.
-// Successful API response (status code 200) is JSON decoded and stored in the value pointed to by v.
-// The request will be retried when the server returns 429 or 5xx status code, or when there is a network error.
-// After reaching the retry timeout, the function will return ErrRetryTimeout.
-// The request will not be retried when the server returns 4xx status code,
-// and the error message will be returned as an error.
-func (c *Client) DoWithRetry(ctx context.Context, reqOptions httpRequest, v interface{}) (*http.Response, error) {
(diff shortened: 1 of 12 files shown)


meghan-kradolfer (approved):

nprizal:
@buildsworth-bk review

PR #551, opened 2026-06-17
Decouple CLI config from the API env payload

pda:
### Description

`config.Config` was doing two unrelated jobs: the client's internal CLI/env configuration *and* the `env` payload serialized into API requests. The same struct was passed as the `Env` field of the `filter_tests` and `test_plan_metadata` requests, and which fields were sent was decided implicitly by JSON struct tags.

Because Go serializes exported fields by default, the safe default was backwards — a new config field would be sent to the API unless someone remembered to tag it `json:"-"`.

This introduces a dedicated `config.EnvPayload` type holding exactly the fields we send today, plus a `Config.EnvPayload()` mapper that is the single deliberate boundary between internal config and the request body. `Config` is now fully severed from JSON serialization, so future config fields are private by default — sending one requires an explicit addition to `EnvPayload`.

Alternatives considered: a custom `MarshalJSON` on `Config` (keeps one struct doing two jobs) and an allowlist map (loses type safety and a clear contract).

### Changes

- Add `config.EnvPayload` type and `Config.EnvPayload()` mapper.
- Switch `FilterTestsParams.Env` and `TestPlanMetadataParams.Env` to `config.EnvPayload`; update the two call sites.
- Tag every `Config` field `json:"-"` so `Config` is never serialized.

Before:

```mermaid
flowchart LR
    C1["config.Config<br/>(all CLI/env config)"]
    E1["&quot;env&quot;: { ... }<br/>API request body"]
    C1 -- "serialized directly<br/>(json tags)" --> E1
```

After:

```mermaid
flowchart LR
    C2["config.Config<br/>(all CLI/env config)<br/>json:&quot;-&quot; — never serialized"]
    P2["config.EnvPayload<br/>(explicit wire type)"]
    E2["&quot;env&quot;: { ... }<br/>API request body"]
    C2 -- "EnvPayload()<br/>explicit mapper" --> P2 --> E2
```

### Testing

- `go build ./...` and `go vet` clean.
- The existing golden-body tests for `filter_tests` and `test_plan_metadata` pass unchanged, confirming the wire format is identical byte for byte. No behavior change.

### Note

A possible follow-up is to prune `EnvPayload` down to the subset the server actually consumes; that's out of scope here. This change just draws the line so the two concepts can evolve independently.

### DIFF
--- a/internal/api/filter_tests.go
+++ b/internal/api/filter_tests.go
@@ -10,8 +10,8 @@ import (
 )
 
 type FilterTestsParams struct {
-	Files []plan.TestCase `json:"files"`
-	Env   *config.Config  `json:"env"`
+	Files []plan.TestCase   `json:"files"`
+	Env   config.EnvPayload `json:"env"`
 }
 
 type FilteredTest struct {
--- a/internal/api/filter_tests_test.go
+++ b/internal/api/filter_tests_test.go
@@ -25,7 +25,7 @@ func TestFilterTests_SlowFiles(t *testing.T) {
 			{Path: "./dog_spec.rb"},
 			{Path: "./turtle_spec.rb"},
 		},
-		Env: &cfg,
+		Env: cfg.EnvPayload(),
 	}
 
 	svr := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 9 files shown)


nprizal (approved):
Nice

PR #555, opened 2026-06-24
feat(gotest): add opt-in Go JSONL result uploads

mattensor:
## Summary

Adds an opt-in Go JSONL mode for the `gotest` runner so TA ingestion can use `GoJSONTestParser` for automatic tagging, while keeping the default JUnit XML flow unchanged.

`bktec` now detects `gotestsum --jsonfile` or `go test -json`, parses the JSONL for retry/status handling, and uploads it as `go-jsonl`. Direct `go test -json` output is captured to `{{resultPath}}` while still streaming to the build log.

Also updates the Buildkite self-test step and docs to use/explain the JSONL mode.

## Testing

```sh
go test ./internal/runner -run 'TestGotest' -count=1
go test ./internal/command -run 'TestRunTestsWithRetry_GoTest' -count=1

### DIFF
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -30,11 +30,10 @@ echo '+++ Running tests'
 
 export BUILDKITE_TEST_ENGINE_SUITE_SLUG=bktec
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=gotest
-export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
+export BUILDKITE_TEST_ENGINE_RESULT_PATH="gotest-${BUILDKITE_JOB_ID}.jsonl"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
-export BUILDKITE_TEST_ENGINE_TEST_CMD='gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out -failfast -race {{packages}}'
+export BUILDKITE_TEST_ENGINE_TEST_CMD='gotestsum --jsonfile={{resultPath}} -- -count=1 -coverprofile=cover.out -failfast -race {{packages}}'
 export BUILDKITE_TEST_ENGINE_UPLOAD_RESULTS=true
-export BUILDKITE_TEST_ENGINE_CLIENT_GOTEST_UPLOAD_GO_JSONL=true
 
 "${bktec_bin}" run
 
--- a/.buildkite/steps/tests.sh
+++ b/.buildkite/steps/tests.sh
@@ -34,6 +34,7 @@ export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
 export BUILDKITE_TEST_ENGINE_TEST_CMD='gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out -failfast -race {{packages}}'
 export BUILDKITE_TEST_ENGINE_UPLOAD_RESULTS=true
+export BUILDKITE_TEST_ENGINE_CLIENT_GOTEST_UPLOAD_GO_JSONL=true
 
 "${bktec_bin}" run
 
--- a/internal/runner/gotest.go
+++ b/internal/runner/gotest.go
@@ -178,7 +178,7 @@ type goJSONLTestEvent struct {
 }
 
 func (g GoTest) parseGoJSONLResults(result *RunResult) error {
-	events, err := loadGoJSONLTestEvents(g.ResultFilePath())
+	events, err := loadGoJSONLTestEvents(g.ResultPath)
 	if err != nil {
 		return err
 	}
--- a/internal/runner/gotest.go
+++ b/internal/runner/gotest.go
@@ -213,7 +213,9 @@ func (g GoTest) parseGoJSONLResults(result *RunResult) error {
 		}
 
 		key := event.Package + "/" + event.Test
-		testStatuses[key] = status
+		if testStatuses[key] != TestStatusFailed {
+			testStatuses[key] = status
+		}
 		if status == TestStatusFailed {
 			packageHasFailedTest[event.Package] = true
 		}
@@ -272,6 +274,9 @@ func loadGoJSONLTestEvents(path string) ([]goJSONLTestEvent, error) {
 	}
 	if err := scanner.Err(); err != nil {
 		return nil, fmt.Errorf("failed to read go jsonl: %w", err)
+	}
+	if len(events) == 0 {
+		return nil, fmt.Errorf("failed to parse go jsonl: no events found")
 	}
 
 	return events, nil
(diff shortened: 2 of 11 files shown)

### CODE REVIEW

File: internal/runner/gotest.go [hunk 1]
```diff
@@ -169,17 +357,57 @@ func (g GoTest) CommandNameAndArgs(testCases []plan.TestCase, retry bool) (strin
...
+}
+
+func (g GoTest) withGoJSONLStdoutCapture(args []string) []string {
+	if !isGoTestJSONCommand(args) || goJSONLFileArg(args) != "" {
+		return args
+	}
+
+	wrapped := []string{"-o", "pipefail", "-c", `"$@" | tee "$0"`, g.ResultPath}
+	wrapped = append(wrapped, args...)
+	return append([]string{"bash"}, wrapped...)
```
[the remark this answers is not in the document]
mttns: Good point, I've addressed this issue

.buildkite/steps/tests.sh
```diff
@@ -30,9 +30,9 @@ echo '+++ Running tests'
 
 export BUILDKITE_TEST_ENGINE_SUITE_SLUG=bktec
 export BUILDKITE_TEST_ENGINE_TEST_RUNNER=gotest
-export BUILDKITE_TEST_ENGINE_RESULT_PATH="junit-${BUILDKITE_JOB_ID}.xml"
+export BUILDKITE_TEST_ENGINE_RESULT_PATH="gotest-${BUILDKITE_JOB_ID}.jsonl"
 export BUILDKITE_TEST_ENGINE_RETRY_COUNT=1
-export BUILDKITE_TEST_ENGINE_TEST_CMD='gotestsum --junitfile={{resultPath}} -- -count=1 -coverprofile=cover.out -failfast -race {{packages}}'
+export BUILDKITE_TEST_ENGINE_TEST_CMD='gotestsum --jsonfile={{resultPath}} -- -count=1 -coverprofile=cover.out -failfast -race {{packages}}'
```
nprizal, line 35: nice 👍

internal/runner/gotest.go
```diff
@@ -49,7 +66,35 @@ func (g GoTest) Name() string {
 }
 
 func (g GoTest) ResultFormat() string {
-	return "junit"
+	return g.resultFormat
+}
+
+func (g GoTest) ResultFilePath() string {
+	if g.resultFormat == goTestResultFormatGoJSONL {
+		return g.goJSONLResultPathFromCommand()
```
nprizal: What was the reasoning behind this? Why can't we return the configured result path?

```
gotestsum --jsonfile={{resultPath}}
```

or
```
go test -json // we direct the output to resultPath
``

Comment (mttns) [reply internal/runner/gotest.go]: This was leftover from an earlier mixed-output approach, I chose to simplify it & now we just support one format at a time. I'll clean this up

File: internal/runner/gotest.go [hunk 4]
```diff
@@ -169,9 +352,7 @@ func (g GoTest) CommandNameAndArgs(testCases []plan.TestCase, retry bool) (strin
 		cmd = cmd + " " + concatenatedPackages
 	}
 
-	cmd = strings.Replace(cmd, "{{resultPath}}", g.ResultPath, 1)
-
-	args, err := shellquote.Split(cmd)
+	args, err := g.commandArgsWithoutPackages(cmd)
```
nprizal: [nit] Maybe we don't need this changes anymore now we only allow one format at a time

internal/runner/gotest.go
```diff
@@ -103,8 +166,128 @@ func (g GoTest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
...
+type goJSONLTestEvent struct {
+	Action      string `json:"Action"`
+	Package     string `json:"Package"`
+	Test        string `json:"Test"`
+	Output      string `json:"Output"`
+	FailedBuild string `json:"FailedBuild"`
+}
+
+func (g GoTest) parseGoJSONLResults(result *RunResult) error {
+	events, err := loadGoJSONLTestEvents(g.ResultFilePath())
```
nprizal: We can now just call use the `ResultPath` from the struct.
```suggestion
	events, err := loadGoJSONLTestEvents(g.ResultPath)
```

meghan-kradolfer: Do we have any validation around if this is an empty file?

internal/runner/gotest.go
```diff
@@ -103,8 +166,128 @@ func (g GoTest) Run(result *RunResult, testCases []plan.TestCase, retry bool) er
...
+			continue
+		}
+
+		status, ok := goJSONLActionStatus(event.Action)
+		if !ok {
+			continue
+		}
+
+		key := event.Package + "/" + event.Test
+		testStatuses[key] = status
```
nprizal: This is a _last-wins_. If the tests run with `--count=3` and the status is `failed, passed, passed` we mark it as `pass`. However, we should mark it as `fail` instead.
```suggestion
        // Failed status should not be overwritten
        if testStatuses[key] != TestStatusFailed {
           		testStatuses[key] = status
        }
```

nprizal (changes requested):

internal/runner/gotest.go
mttns: Good point, I think it will pass silently currently. I'll add some validations 🙏

internal/runner/gotest.go
mttns: Thanks! I'll fix this one up

internal/runner/gotest.go
mttns: Thanks Naufan! I'll add a regression test around this case too

internal/runner/gotest.go
```diff
@@ -49,7 +66,15 @@ func (g GoTest) Name() string {
...
 
 func (g GoTest) ResultFormat() string {
-	return "junit"
+	return g.resultFormat
+}
+
+func commandProducesGoJSONL(command string) bool {
+	args, err := shellquote.Split(command)
+	if err != nil {
+		return strings.Contains(command, "--jsonfile") || strings.Contains(command, "go test -json")
```
malclocke: **nit** the check against `go test -json` is a little brittle. `go test` takes a number of args and a user might potentially configure their test command so as to break this check, e.g.

```
export BUILDKITE_TEST_ENGINE_TEST_CMD="go test -v -json {{packages}}"
```

Suggest we make the check less specific


```suggestion
		return strings.Contains(command, "--jsonfile") || strings.Contains(command, "-json")
```

.buildkite/steps/tests.sh
malclocke, line 35: Out of scope of this PR, but it would be good to introduce another build step that uses `go test`.

internal/runner/gotest.go
nprizal: can we use `isGoTestJSONCommand`?

mttns: Good point! Thanks Malc, committing this suggestion 🙏

internal/runner/gotest_test.go
```diff
@@ -210,6 +211,50 @@ func TestGotestRun_GoJSONLPackageLevelFailure(t *testing.T) {
...
+	})
+	result := NewRunResult([]plan.TestCase{})
+	err = gotest.parseGoJSONLResults(result)
+
+	if err == nil || !strings.Contains(err.Error(), "no events found") {
+		t.Errorf("GoTest.parseGoJSONLResults() error = %v, want no events found", err)
+	}
+}
+
+func TestGotestRun_GoJSONLFailureIsNotOverwrittenByLaterPass(t *testing.T) {
```
nprizal, line 231: 💚

nprizal (approved):

PR #557, opened 2026-06-26
chore: Add manual version override to release pipeline

mattensor:
## Summary

This adds a manual version override to the release pipeline.

We hit this while trying to cut `v2.9.0-rc.3`: the release prompt only showed the versions suggested by `svu`, and because there was a `feat:` conventional commit, `svu` suggested the next minor pre-release instead of the RC version we wanted.

With this change, the release block still shows the normal generated version options, but also includes a “Version override” option. When selected, the release job uses the exact version entered in the override field.

The release script also validates the final version and fails early if the override is missing, invalid, or already exists as a tag.

## Testing

- Ran shell syntax checks for the release scripts
- Verified the diff has no whitespace errors
- Spot-checked version validation for valid and invalid RC examples

## UI Screenshot

image

### DIFF
--- a/.buildkite/steps/release.sh
+++ b/.buildkite/steps/release.sh
@@ -1,7 +1,40 @@
 #!/usr/bin/env sh
 set -euo pipefail
 
-tag=$(buildkite-agent meta-data get "release-version")
+override_option_value="__override__"
+version_number='(0|[1-9][0-9]*)'
+prerelease_identifier='(0|[1-9][0-9]*|[0-9A-Za-z-]*[A-Za-z-][0-9A-Za-z-]*)'
+version_pattern="^v${version_number}\\.${version_number}\\.${version_number}(-${prerelease_identifier}(\\.${prerelease_identifier})*)?$"
+
+selected_tag=$(buildkite-agent meta-data get "release-version")
+override_tag=$(buildkite-agent meta-data get "release-version-override" 2>/dev/null || true)
+override_tag=$(printf '%s' "${override_tag}" | sed 's/^[[:space:]]*//;s/[[:space:]]*$//')
+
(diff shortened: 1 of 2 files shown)


.buildkite/steps/release.sh
```diff
@@ -1,7 +1,38 @@
 #!/usr/bin/env sh
 set -euo pipefail
 
-tag=$(buildkite-agent meta-data get "release-version")
+override_option_value="__override__"
+version_pattern='^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$'
```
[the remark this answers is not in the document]
mttns: Fair, I'll tighten the validation

.buildkite/steps/release.sh
```diff
@@ -1,7 +1,40 @@
...
+  echo "Invalid release version: ${tag}"
+  exit 1
+fi
+
+if ! printf '%s\n' "${tag}" | grep -Eq "${version_pattern}"; then
+  echo "Invalid release version: ${tag}"
+  exit 1
+fi
+
+if git rev-parse -q --verify "refs/tags/${tag}" >/dev/null; then
```
nprizal, line 34: [nit] CMIIW, but I don't think we need this. `git tag` will fail when tag already exists
```
$ git tag v2.9.0-rc.2
fatal: tag 'v2.9.0-rc.2' already exists
```

nprizal (approved):

.buildkite/steps/release.sh
mttns: Yeah `git tag v2.9.0-rc.2` would fail currently as it already exists. I think the main benefit here is a clear error message before attempting to create the tag

PR #563, opened 2026-06-29
[TE-6222] Add GetSelectors to test runners for selector based splitting

nprizal:
### Description

We're introducing selector based splitting in bktec. Right now Go supports selector based splitting (when you turn it on with the CLI option / env var), and every other runner still goes through file based splitting. Eventually we want to move all runners over to selector based splitting and drop file based entirely.

While we support both, the code gets confusing because Go's `GetFiles` doesn't actually return files. It returns package import paths, since "file" isn't a first class concept in Go. This PR is a refactor to make that less confusing by giving runners an explicit `GetSelectors` method, separate from `GetFiles`.

Selector based splitting is opt-in (gated behind `BUILDKITE_TEST_ENGINE_SELECTOR_SPLITTING`) and only Go supports it for now. This change should be a no-op for existing behavior.

One thing to note: when selector splitting is on, `--files` stops working. That's intentional. `--files` shouldn't be used with selector splitting, so I'm going to introduce a `--selectors` flag for that purpose in a follow up.

### Context

- [TE-6222](https://linear.app/buildkite/issue/TE-6222/support-other-runners-for-selector-splitting-in-bktec)
- Follows on from the Go-only POC in [TE-6156](https://linear.app/buildkite/issue/TE-6156)
- We need to avoid regressing the file based runners that rely on the existing MV until the new selector MV lands ([TE-6201](https://linear.app/buildkite/issue/TE-6201), [TE-6200](https://linear.app/buildkite/issue/TE-6200))

### Changes

- Added `GetSelectors()` to the `TestRunner` interface
- Go runner now returns package import paths via `GetSelectors`, and `GetFiles` just delegates to it
- Every other runner returns a "not supported" error from `GetSelectors` for now
- `run.go` chooses selectors or files based on whether selector splitting is enabled and the runner supports it

### Testing

Existing tests pass. Behavior is unchanged unless selector splitting is explicitly turned on, which only Go supports today.

### DIFF
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -35,9 +35,19 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 		return fmt.Errorf("unsupported value for BUILDKITE_TEST_ENGINE_TEST_RUNNER: %w", err)
 	}
 
-	files, err := getTestFiles(testListFilename, testRunner)
-	if err != nil {
-		return err
+	var testTargets []string
+	if shouldUseSelectorSplitting(cfg, testRunner) {
+		selectors, err := testRunner.GetSelectors()
+		if err != nil {
+			return err
+		}
+		testTargets = selectors
+	} else {
+		files, err := getTestFiles(testListFilename, testRunner)
+		if err != nil {
+			return err
+		}
+		testTargets = files
 	}
 
 	// get plan
@@ -48,7 +58,7 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 		OrganizationSlug: cfg.OrganizationSlug,
 	})
 
-	testPlan, err := fetchOrCreateTestPlan(ctx, apiClient, cfg, files, testRunner)
+	testPlan, err := fetchOrCreateTestPlan(ctx, apiClient, cfg, testTargets, testRunner)
 	if err != nil {
 		return err
 	}
--- a/internal/command/run.go
+++ b/internal/command/run.go
@@ -35,19 +35,9 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
 		return fmt.Errorf("unsupported value for BUILDKITE_TEST_ENGINE_TEST_RUNNER: %w", err)
 	}
 
-	var testTargets []string
-	if shouldUseSelectorSplitting(cfg, testRunner) {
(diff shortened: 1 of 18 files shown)


internal/command/run.go
```diff
@@ -35,9 +35,19 @@ func Run(ctx context.Context, cfg *config.Config, testListFilename string) error
...
+			return err
+		}
+		testTargets = selectors
+	} else {
+		files, err := getTestFiles(testListFilename, testRunner)
+		if err != nil {
+			return err
+		}
+		testTargets = files
 	}
```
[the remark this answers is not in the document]
nprizal: This is intentional. I'm going to introduce `--selectors` option to provide list of selectors instead of reusing `--files` which is confusing.

internal/command/test_target.go
```diff
@@ -0,0 +1,22 @@
...
+)
+
+func getTestTargets(cfg *config.Config, runner runner.TestRunner, testFileList string) ([]string, error) {
+	if shouldUseSelectorSplitting(cfg, runner) {
+		selectors, err := runner.GetSelectors()
+		if err != nil {
+			return nil, err
+		}
+		return selectors, nil
+	} else {
```
gchan: Nit - given the the branches above early return, the `else` is redundant and we could treat the first branch as a guard clause.

internal/runner/cucumber.go
```diff
@@ -150,6 +150,10 @@ func mapScenarioToTestCase(featureName string, featureURI string, scenario Cucum
 	}
 }
 
+func (c Cucumber) GetSelectors() ([]string, error) {
+	return nil, fmt.Errorf("not supported in Cucumber")
+}
```
gchan: Nitpick - Can these 'not supported' errors be extracted to a default function in RunnerConfig? 

Here's what's an LLM suggested:

```go
// GetSelectors returns a "not supported" error by default. Selector based
// splitting is currently only implemented by the Go runner, which overrides
// this. Every other runner inherits this default until it gains selector
// support.
func (rc RunnerConfig) GetSelectors() ([]string, error) {
	return nil, fmt.Errorf("selector based splitting is not supported for the %s runner", rc.TestRunner)
}
```

gchan (approved):
Approved with a couple of nits.

mattensor (approved):

PR #566, opened 2026-06-29
[TE-6228] Add --plan-identifier flag to the plan command

gchan:
### Description

Add the `--plan-identifier` flag (env `BUILDKITE_TEST_ENGINE_PLAN_IDENTIFIER`) to `bktec plan`, so a plan can be generated under a caller-supplied identifier rather than only consumed by `bktec run`. The identifier is the plan's server-side cache key; a non-blank value also skips the `BUILDKITE_BUILD_ID`/`BUILDKITE_STEP_ID` guards in `validate()`, which lets `plan` run off-agent (locally, outside a Buildkite build).

This **reverses a decision made in [#102](https://github.com/buildkite/test-engine-client/pull/102)** (`tat-209`), which removed the ability to override the identifier to tie plan identity to the build. Re-exposing it on `plan` is a deliberate, scoped reopening of that override for the off-agent use case, and needs the client owner's sign-off. The flag itself already existed on `run`; this PR only registers it on `plan` and broadens its usage text.

**Decision — identity over faking build vars:** chosen over making callers set a dummy `BUILDKITE_BUILD_ID`/`BUILDKITE_STEP_ID`, which would leak a fake build identity downstream.

### Context

- Linear: [TE-6228](https://linear.app/buildkite/issue/TE-6228) (sub-issue of [TE-6168](https://linear.app/buildkite/issue/TE-6168))
- Reverses: [#102](https://github.com/buildkite/test-engine-client/pull/102)

### Changes

- `cli.go`: register `planIdentifierFlag` on `planCommandFlags()`; rewrite its usage to describe both `run` and `plan`.
- Tests: flag registration on `plan`; identifier flows to the `test_plan` create request param; the `validate()` identity branch (identifier precedence over build/step, off-agent validation, and the blank-identifier failure).
- Docs: README `Plan identifier` section and an unreleased changelog entry.

### Testing

Backed by specs; AI assisted. `go build ./...`, `go vet ./...`, and `go test . ./internal/command/ ./internal/config/` all pass.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,12 @@ need to set `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, which are otherwise
 required. Use a unique value per request (a UUIDv7 works well) so plans don't
 collide or silently return a stale cached plan.
 
+> [!IMPORTANT]
+> The identifier must be unique per **step**, not just per build. It is the
+> cache key for the plan, so two steps sharing an identifier get the *same*
+> cached plan, and the second step runs the first step's test split instead of
+> its own. The default `${BUILDKITE_BUILD_ID}/${BUILDKITE_STEP_ID}` includes the
+> step id for exactly this reason.
 ```sh
 ./bktec plan --json --plan-identifier 01919f1e-0000-7000-8000-000000000000
 ```
--- a/README.md
+++ b/README.md
@@ -92,6 +92,30 @@ export BUILDKITE_TEST_ENGINE_TAGS="env=production,region=us-east-1"
 
 Test collectors are available for many languages and frameworks. Some collectors also provide richer data collection such as execution-level tagging and span tracing. See the [test collector docs](https://buildkite.com/docs/test-engine/test-collection) for details on what's available for your framework.
 
+### Plan identifier
+
+`--plan-identifier` (or `BUILDKITE_TEST_ENGINE_PLAN_IDENTIFIER`) sets the
+identifier the `plan` command generates the plan under. The identifier is the
+plan's server-side cache key: distinct values produce distinct plans, while
+reusing a value returns the previously cached plan for that identifier.
+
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -92,6 +92,30 @@ export BUILDKITE_TEST_ENGINE_TAGS="env=production,region=us-east-1"
...
+reusing a value returns the previously cached plan for that identifier.
+
+Inside a Buildkite build you don't need to set this; the identifier defaults to
+`${BUILDKITE_BUILD_ID}/${BUILDKITE_STEP_ID}`. Supply `--plan-identifier`
+explicitly when generating a plan **off-agent** (for example, running it in your
+local development environment or on your own machine). Doing so also removes the
+need to set `BUILDKITE_BUILD_ID` and `BUILDKITE_STEP_ID`, which are otherwise
+required. Use a unique value per request (a UUIDv7 works well) so plans don't
+collide or silently return a stale cached plan.
+
```
nprizal, line 108: Can we add note about it has to be unique per step rathern than per build to prevent 2 different test steps sharing a plan.

```suggestion

> [!IMPORTANT]
> The identifier must be unique per **step**, not just per build. It is the
> cache key for the plan, so two steps sharing an identifier get the *same*
> cached plan, and the second step runs the first step's test split instead of
> its own. The default `${BUILDKITE_BUILD_ID}/${BUILDKITE_STEP_ID}` includes the
> step id for exactly this reason.
```

nprizal (commented):
Looks good to me. The identifier only used for caching. This doesn't affect the assocation between a plan and a test run (we use build id and step id from the metadata reported by `bktec` post-run for this).

The original decision was made to prevent different test steps (e.g monorepo build with multiple test steps) sharing a test plan. Can we make it clear that is has to be unique per test step?

nprizal (approved):

PR #568, opened 2026-06-30
Support selector-based splitting for the custom runner

nprizal:
### Description

Selector-based splitting is currently only supported by the Go runner. We want to bring it to the other runners too, and this PR is the first step, starting with the custom runner. This is still experimental and may change.

Splitting by selector lets you divide a suite by something other than a test file, like a tag or package. The custom runner can't auto-discover selectors the way the Go runner does with `go list`, so instead you point bktec at a file with one selector per line. bktec distributes those selectors across parallel nodes the same way it distributes test files, and the selectors assigned to each node replace `{{testExamples}}` in `BUILDKITE_TEST_ENGINE_TEST_CMD`.

To use it:

```sh
export BUILDKITE_TEST_ENGINE_TEST_RUNNER=custom
export BUILDKITE_TEST_ENGINE_TEST_CMD="bin/test {{testExamples}}"
export BUILDKITE_TEST_ENGINE_SELECTOR_SPLITTING=true
export BUILDKITE_TEST_ENGINE_SELECTOR_LIST_PATH="path/to/selectors.txt"
bktec run
```

Since it's still experimental, the new `--selectors` flag (`BUILDKITE_TEST_ENGINE_SELECTOR_LIST_PATH`) is hidden for now.

### Context

- Linear: [TE-6225](https://linear.app/buildkite/issue/TE-6225/support-selector-based-splitting-for-custom-runner)
- Parent: TE-6222 (added `GetSelectors` and selector splitting for the Go runner)

### Changes

- Added the `--selectors` flag / `BUILDKITE_TEST_ENGINE_SELECTOR_LIST_PATH` env var for providing a selector list file. When a selector file is given it takes precedence over a runner's own `GetSelectors`, so this also lets the Go runner read selectors from a file instead of auto-discovering them
- Taught the custom runner to support selector splitting, sourcing selectors from the file
- Required `--selector-splitting` to be enabled when a selector list is provided, with validation
- Documented selector splitting in the custom runner docs

Also, while working on this I moved the custom runner test fixtures over to bats and made sure bats gets installed in the pipeline Dockerfile and `bin/setup`.

### Testing

Covered by unit tests for the new validation and the supported-features changes, plus the bats-based custom runner test fixtures.

### DIFF
--- a/.buildkite/Dockerfile
+++ b/.buildkite/Dockerfile
@@ -8,6 +8,9 @@ COPY --from=ruby / /
 COPY --from=cypress / /
 COPY --from=python / /
 
+RUN git clone --depth 1 https://github.com/bats-core/bats-core.git /tmp/bats-core \
+  && /tmp/bats-core/install.sh /usr/local \
+  && rm -rf /tmp/bats-core
 RUN gem install rspec cucumber base64
 RUN gem install bigdecimal -v 3.2.0
 RUN yarn global add jest
--- a/cli.go
+++ b/cli.go
@@ -283,11 +283,11 @@ var filesFlag = &cli.StringFlag{
 }
 
 var selectorsFlag = &cli.StringFlag{
-	Name:        "selectors",
+	Name:        "selector-file",
 	Category:    "TEST RUNNER",
 	Value:       "",
-	Usage:       "A file path containing a list of selectors to run (one per line). Must be used alongside `selector-splitting`",
-	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_SELECTOR_LIST_PATH"),
+	Usage:       "Path to a file containing a list of selectors to run (one per line). Must be used alongside `selector-splitting`",
+	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_SELECTOR_FILE"),
 	Destination: &cfg.SelectorListPath,
 	Hidden:      true,
 }
--- a/cli.go
+++ b/cli.go
@@ -282,6 +282,16 @@ var filesFlag = &cli.StringFlag{
 	Sources:  cli.EnvVars("BUILDKITE_TEST_ENGINE_FILES"),
 }
 
+var selectorsFlag = &cli.StringFlag{
+	Name:        "selectors",
+	Category:    "TEST RUNNER",
+	Value:       "",
+	Usage:       "A file path containing a list of selectors to run (one per line). Must be used alongside `selector-splitting`",
+	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_SELECTOR_LIST_PATH"),
+	Destination: &cfg.SelectorListPath,
+	Hidden:      true,
+}
+
 var tagFiltersFlag = &cli.StringFlag{
 	Name:        "tag-filters",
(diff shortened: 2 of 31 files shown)


.buildkite/Dockerfile
```diff
@@ -8,6 +8,9 @@ COPY --from=ruby / /
 COPY --from=cypress / /
 COPY --from=python / /
 
+RUN git clone --depth 1 https://github.com/bats-core/bats-core.git /tmp/bats-core \
```
nprizal, line 11: I'm changing the custom runner test data with bats tests.

cli.go
```diff
@@ -282,6 +282,16 @@ var filesFlag = &cli.StringFlag{
 	Sources:  cli.EnvVars("BUILDKITE_TEST_ENGINE_FILES"),
 }
 
+var selectorsFlag = &cli.StringFlag{
+	Name:        "selectors",
```
nprizal: Feedback from @pda during demo

```suggestion
	Name:        "selectors-file",
```
or
```suggestion
	Name:        "selectors-path",
```

internal/runner/custom.go
```diff
@@ -41,6 +41,7 @@ func (c Custom) SupportedFeatures() SupportedFeatures {
 		AutoRetry:       false,
 		Mute:            true,
 		Skip:            false,
+		SplitBySelector: true,
```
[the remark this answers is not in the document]
nprizal: Good catch. Fixed in 444a86f7. I threaded `SelectorSplitting` and `SelectorListPath` into `RunnerConfig` and added a `splitBySelectorList()` helper, then relaxed both checks so the file pattern is only required when we're not splitting by a provided selector list. `ValidateForRun` and `NewCustom` now both let the documented selector-only config through. Added tests for both.

cli.go
```diff
@@ -282,6 +282,16 @@ var filesFlag = &cli.StringFlag{
 	Sources:  cli.EnvVars("BUILDKITE_TEST_ENGINE_FILES"),
 }
 
+var selectorsFlag = &cli.StringFlag{
+	Name:        "selector-file",
+	Category:    "TEST RUNNER",
+	Value:       "",
+	Usage:       "Path to a file containing a list of selectors to run (one per line). Must be used alongside `selector-splitting`",
+	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_SELECTOR_FILE"),
```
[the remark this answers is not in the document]
nprizal: Done in d9353b03. Set `BUILDKITE_TEST_ENGINE_SELECTOR_FILE` in `TestRunCommandEnvVarsBindToConfig` and asserted it binds to `cfg.SelectorListPath`.

docs/custom-test-runner.md
```diff
@@ -27,6 +27,39 @@ You can exclude specific files or directories that match a certain pattern using
...
+search
+account
+```
+
+The actual command that bktec runs on each node will look like this:
+
+```sh
+bin/test checkout search
+```
+
```
meghan-kradolfer: Should we be adding this to the docs now? It may make it easier to find and we haven't turned the feature flag on/add all tag attribution yet

meghan-kradolfer (approved):

PR #570, opened 2026-07-01
[TE-6254] Add --plan-out output mode to bktec plan

gchan:
### Description

Add `--plan-out` to `bktec plan`: it writes the full test plan without running any tests, so users can pull down a copy of the plan instead of curling the unpublished `test_plan` endpoint by hand.

`bktec plan` already creates a plan without running tests, but `--json` emits only the plan identifier and parallelism. `--plan-out` writes the whole plan — tasks, the per-node test breakdown, muted and skipped tests, and timing metadata — as the server's response, byte-for-byte, only re-indented. It takes a destination: `-` for stdout, or a file path. It is an additive third arm of the existing, mutually-exclusive PLAN OUTPUT group; `--json` and `--pipeline-upload` are unchanged, so existing call sites keep working.

Emitting the server's raw JSON (rather than decoding into `plan.TestPlan` and re-marshalling) means any field the client does not model is preserved, and the output cannot drift from what a fetch of the endpoint returns. `--plan-out` also does not substitute a local fallback for a server *error* plan: an empty plan (`{"tasks": {}}`) is passed through as-is, with a warning on stderr. Only when the server is unreachable does it emit a minimal locally-generated plan (identifier + parallelism, no tasks — not a computed split), noted on stderr. Fatal API errors (auth, forbidden, bad request) exit non-zero with no output.

**Naming:** this started as `--full-json` (see the branch/early commits) and became `--plan-out <path|->` after review — a destination-taking flag is more flexible (writes straight to a file) and reads better next to `--json`/`--pipeline-upload`. A `--plan-out-mode` flag was considered and dropped: with only a `raw` mode it would be inert, and it can be added non-breakingly later.

### Context

Linear: [TE-6254](https://linear.app/buildkite/issue/TE-6254), a sub-issue of [TE-6168](https://linear.app/buildkite/issue/TE-6168) (self-serve evaluation of Test Prediction). This productises the manual `curl GET .../test_plan` capture the evaluation flow relies on, alongside the sibling `--plan-identifier` work (TE-6228, #566).

### Changes

The branch evolved through review (full-json → raw passthrough → --plan-out); best reviewed as the net diff rather than commit-by-commit. The final shape:

- `cli.go` / `main.go`: `--plan-out` flag in the PLAN OUTPUT mutually-exclusive group, bound to `cfg.PlanOut`; dispatched via `cmd.IsSet`.
- `internal/api/create_test_plan.go`: `CreateTestPlanRaw` returns the decoded plan plus the raw JSON response.
- `internal/command/plan.go`: `planOut` writes the raw response to the destination (`planOutWriter`: stdout or a file); empty/unreachable handling as above.
- Tests in `plan_test.go` (stdout + file destinations, parallelism 0, error-plan passthrough, unreachable fallback, fatal error) and `cli_test.go`; README + changelog.

### Testing

`go test ./internal/command/... ./internal/api/... .` passes. Manually verified with `go run . plan`: `--plan-out` appears in `--help`; combining it with `--json` errors; omitting all output flags errors with the required-group message. Buildkite CI (build #2363) passed on an earlier head; will re-run on the latest.

The `internal/runner` Jest test fails locally without a node/jest environment; that failure is present on `main` and is unrelated to this change.

AI assisted.

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,8 +136,8 @@ the muted and skipped tests, and the timing metadata. It takes a destination:
 `-` for stdout, or a file path.
 
 ```sh
-./bktec plan --plan-out - > plan.json    # stdout
-./bktec plan --plan-out plan.json        # a file
+./bktec plan --plan-out -            # stdout
+./bktec plan --plan-out plan.json    # a file
 ```
 
 The human-readable split summary and any warnings are written to stderr, so
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ tests, the muted and skipped tests, and the timing metadata. Redirect stdout to
 keep a copy:
 
 ```sh
-./bktec plan --full-json --plan-identifier "$(uuidgen)" > plan.json
+./bktec plan --full-json ... > plan.json
 ```
 
 The human-readable split summary and any warnings are written to stderr, so
--- a/README.md
+++ b/README.md
@@ -143,6 +143,11 @@ The human-readable split summary and any warnings are written to stderr, so
 stdout carries only the JSON. `--json`, `--full-json` and `--pipeline-upload`
 are mutually exclusive; choose one.
 
+`--full-json` prints what the server returns. If the server cannot generate a
+plan it returns an empty plan, which is emitted as-is (a warning is printed to
+stderr). Only when the server cannot be reached at all does `bktec` fall back to
+a locally-computed plan, again noting this on stderr.
+
 ### Preview: Test Selection
 You can pass test selection strategy configuration and additional change context to the test plan API request.
 This preview is enabled only when `BKTEC_PREVIEW_SELECTION` is truthy (`1`, `true`, `yes`, or `on`).
--- a/README.md
+++ b/README.md
@@ -127,6 +127,22 @@ identifier, or, on a cache miss, creates and caches one under it. A reused
 identifier therefore returns the previously cached plan even if the inputs have
 changed, so keep the value unique per distinct plan.
 
+### Inspecting the full plan
+
+`--full-json` makes `bktec plan` print the full test plan to stdout as JSON,
+without running any tests. Unlike `--json`, which emits only the plan identifier
+and parallelism, `--full-json` includes the tasks, the per-node breakdown of
+tests, the muted and skipped tests, and the timing metadata. Redirect stdout to
+keep a copy:
+
+```sh
+./bktec plan --full-json --plan-identifier "$(uuidgen)" > plan.json
+```
+
+The human-readable split summary and any warnings are written to stderr, so
+stdout carries only the JSON. `--json`, `--full-json` and `--pipeline-upload`
+are mutually exclusive; choose one.
+
 ### Preview: Test Selection
 You can pass test selection strategy configuration and additional change context to the test plan API request.
 This preview is enabled only when `BKTEC_PREVIEW_SELECTION` is truthy (`1`, `true`, `yes`, or `on`).
--- a/README.md
+++ b/README.md
@@ -129,24 +129,25 @@ changed, so keep the value unique per distinct plan.
 
 ### Inspecting the full plan
 
-`--full-json` makes `bktec plan` print the full test plan to stdout as JSON,
-without running any tests. Unlike `--json`, which emits only the plan identifier
-and parallelism, `--full-json` includes the tasks, the per-node breakdown of
-tests, the muted and skipped tests, and the timing metadata. Redirect stdout to
-keep a copy:
+`--plan-out` makes `bktec plan` write the full test plan without running any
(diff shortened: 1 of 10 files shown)


cli_test.go
```diff
@@ -74,6 +74,72 @@ func TestPlanCommandIncludesPlanIdentifierFlag(t *testing.T) {
 	}
 }
 
+func TestPlanCommandIncludesFullJSONOutputFlag(t *testing.T) {
+	planCmd := findCommand(cliCommand, "plan")
```
gchan: Do we need this findCommand function?

cli_test.go
```diff
@@ -74,6 +74,72 @@ func TestPlanCommandIncludesPlanIdentifierFlag(t *testing.T) {
...
+	if planCmd == nil {
+		t.Fatal("cliCommand missing the plan subcommand")
+	}
+
+	if !mutuallyExclusiveGroupHasFlag(planCmd, "PLAN OUTPUT", "full-json") {
```
gchan: Same question here, do we need `mutuallyExclusiveGroupHasFlag`?

README.md
```diff
@@ -127,6 +127,27 @@ identifier, or, on a cache miss, creates and caches one under it. A reused
...
+### Inspecting the full plan
+
+`--full-json` makes `bktec plan` print the full test plan to stdout as JSON,
+without running any tests. Unlike `--json`, which emits only the plan identifier
+and parallelism, `--full-json` includes the tasks, the per-node breakdown of
+tests, the muted and skipped tests, and the timing metadata. Redirect stdout to
+keep a copy:
+
+```sh
+./bktec plan --full-json --plan-identifier "$(uuidgen)" > plan.json
```
gchan: Let's simplify the example and replace the other options with `...` or something similar 

```
bktec plan --full-json ... > plan.json
```

internal/command/plan.go
```diff
@@ -134,33 +245,61 @@ func makePipelineUploadCommand(template string) *exec.Cmd {
...
 	if len(testPlan.Tasks) == 0 {
 		warnErrorPlan()
-		return fallbackPlan, nil
+		return makeFallbackPlan(cfg, testTargets), nil
 	}
 
 	return testPlan, nil
 }
 
+func makeFallbackPlan(cfg *config.Config, testTargets []string) plan.TestPlan {
```
gchan: The previous code simply returned:

```
fallbackPlan := plan.TestPlan{
		Identifier:  cfg.Identifier,
		Parallelism: cfg.MaxParallelism,
		Fallback:    true,
	}
```

Can we continue with this approach?

internal/command/plan.go
```diff
@@ -126,6 +134,109 @@ func Plan(ctx context.Context, cfg *config.Config, testFileList string, outputFo
...
+// empty error plan) is passed through verbatim, so the output faithfully
+// reflects what the server has. A locally-computed fallback is used only when
+// the server cannot be reached at all, since there is nothing to pass through.
+func fullJSONPlan(ctx context.Context, cfg *config.Config, testTargets []string, apiClient *api.Client, testRunner runner.TestRunner) error {
+	params, err := createRequestParam(ctx, cfg, testTargets, *apiClient, testRunner)
+	if err != nil {
+		if handledErr := handleError(err); handledErr != nil {
+			return handledErr
+		}
+		return emitLocalFallback(cfg, testTargets)
```
gchan: Can we just emit a dumb fallback as per the previous behaviour?

gchan:
@buildsworth-bk review

internal/command/plan.go
```diff
@@ -126,6 +134,99 @@ func Plan(ctx context.Context, cfg *config.Config, testFileList string, outputFo
...
+		return nil, nil, fmt.Errorf("opening --plan-out file: %w", err)
+	}
+	return f, func() { f.Close() }, nil
+}
+
+// emitLocalFallback writes a locally-computed fallback plan as JSON, used by
+// --plan-out when the server cannot be reached and there is nothing to pass
+// through. The fallback carries no tasks, so only the identifier and
+// parallelism are emitted.
+func emitLocalFallback(out io.Writer, cfg *config.Config) error {
```
gchan, line 199: I don't like this. I think it should hard fail. But I'm sticking to the existing conventions of `makeFallbackPlan` for the other modes for now.

I do want to make a pitch for better handling when the server can't be reached (out of scope of this PR).

cli.go
```diff
@@ -467,6 +467,12 @@ var jsonFlag = &cli.BoolFlag{
 	Usage: "JSON format output",
 }
 
+var planOutFlag = &cli.StringFlag{
+	Name:        "plan-out",
```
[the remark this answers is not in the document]
gchan: We've renamed it

pda (approved):

PR #575, opened 2026-07-02
[TE-6238] Prepend and trim location prefix when split by selector

nprizal:
### Description
This is a spike for one of the options we're weighing for how location prefix should interact with selector based splitting (see Context). Specifically, it's what `bktec` needs to look like under the option where `test.selector.primary` includes the location prefix.

Right now, selector based splitting sends discovered test targets straight to the Test Plan API as raw selectors, with no location prefix applied. If ingestion ends up setting `test.selector.primary` with the prefix included, our selector won't match what's stored, and the test plan will come back with no historical duration.

This PR makes `bktec` prepend the location prefix to a selector before sending it, but only when the selector is actually a file path (e.g. Rspec). Some runners use selectors that aren't file paths at all (e.g. Go package names), so we check with `os.Stat` first and leave those untouched.

I also fixed `trimTaskLocationPrefix` in `run.go`, which previously skipped selector-format tests entirely when trimming the prefix from the plan we get back. Now it trims the prefix from `test.Value` for those too, so the runner gets a selector it can actually use.

### Context
We hit this while testing selector based splitting for rspec in `buildkite/buildkite` ([build](https://buildkite.com/buildkite/buildkite/builds/199535/canvas?jid=019f1bfd-f21f-4ef5-aa50-b273b4c1012c&tab=output)) and the plan came back with no historical duration.

We're still deciding how ingestion should set `test.selector.primary` when location prefix is configured: leave the prefix out, or include it. This PR is a spike for what `bktec` looks like if we go with including it. Wrote up the full problem in [this doc](https://app.notion.com/p/Selector-Splitting-Location-Prefix-Strikes-Back-390b8dbc2c898009bb2ce151a83bdf76) if you want more background.

Linear: [TE-6238](https://linear.app/buildkite/issue/TE-6238/support-selector-based-splitting-for-rspec)

### Testing
Covered by unit tests.

### DIFF
--- a/internal/command/request_param.go
+++ b/internal/command/request_param.go
@@ -3,6 +3,7 @@ package command
 import (
 	"context"
 	"fmt"
+	"os"
 	"strings"
 
 	"github.com/buildkite/test-engine-client/v2/internal/api"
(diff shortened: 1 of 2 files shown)


nprizal:
We have decided to exclude location prefix in the selector. Therefore this is no longer necessary.

PR #580, opened 2026-07-03
TE-6261 - Add a vitest runner

mcncl:
### Description

Currently folks need to use the Jest runner with Vitest, as Vitest supports the Jest shape of JSON, however Vitest is (arguably) becoming more popular and offers some benefits over Jest, like explicit `skipped` status.

### Context

This was raised by a customer as an enhancement on the engine, rather than using Jest.

### Changes

- adds a Vitest test runner which can be used with `BUILDKITE_TEST_ENGINE_RUNNER=vitest`
    - a lot of the runner is similar/the same as Jest, Vitest just handles things like skipping in a more convenient way
- creates a `ParseJestReport` function for Jest and Vitest to share, with more generic erroring (removed `Jest`)
- adds `vitest` to config locations

### Testing

- `go test ./...` with targeted testing passes, full test suite fails locally as it looks to use some CI specific function


internal/runner/supported_features_test.go
```diff
@@ -39,6 +40,7 @@ func TestSupportedFeatures_SplitBySelectorSupportedRunners(t *testing.T) {
 		gotest,
 		cucumber,
 		nunit,
+    vitest,
```
nprizal:
```suggestion
		vitest,
```

internal/runner/supported_features_test.go
```diff
@@ -27,6 +27,7 @@ func TestSupportedFeatures_SplitBySelectorSupportedRunners(t *testing.T) {
 	gotest := NewGoTest(runnerConfig)
 	cucumber := NewCucumber(runnerConfig)
 	nunit := NewNUnit(runnerConfig)
+  vitest := NewVitest(runnerConfig)
```
nprizal:
```suggestion
	vitest := NewVitest(runnerConfig)
```

internal/runner/vitest.go
```diff
@@ -0,0 +1,206 @@
...
+		Skip:            false,
+	}
+}
+
+func (v Vitest) Name() string {
+	return "Vitest"
+}
+
+func (v Vitest) ResultFormat() string {
+	return "jest-json"
```
nprizal, line 60: 👍 I've checked the `jest-json` parser in the ingestion and it already handle `skipped` result.

internal/runner/vitest.go
```diff
@@ -0,0 +1,206 @@
...
+// Vitest's JSON reporter is Jest-compatible, so the result file is parsed using
+// the shared JestReport types. The main differences from the Jest runner are
+// the default test/retry commands and the result format identifier.
+type Vitest struct {
+	RunnerConfig
+}
+
+func NewVitest(v RunnerConfig) Vitest {
+	if v.TestCommand == "" {
+		v.TestCommand = "npx vitest run {{testExamples}} --reporter=json --outputFile {{resultPath}}"
```
nprizal:
```suggestion
		v.TestCommand = "npx vitest run {{testExamples}} --reporter=default --reporter=json --outputFile {{resultPath}}"
```
Vitest doesn't print to stdout when `--reporter=json` is set. We need to set both like we do for rspec.

internal/runner/vitest.go
```diff
@@ -0,0 +1,206 @@
...
+	}
+
+	if v.TestFilePattern == "" {
+		v.TestFilePattern = "**/{__tests__/**/*,*.spec,*.test}.{ts,js,tsx,jsx}"
+	}
+
+	if v.RetryTestCommand == "" {
+		v.RetryTestCommand = "npx vitest run --testNamePattern '{{testNamePattern}}' --reporter=json --outputFile {{resultPath}}"
```
nprizal:
```suggestion
		v.RetryTestCommand = "npx vitest run --testNamePattern '{{testNamePattern}}' --reporter=default --reporter=json --outputFile {{resultPath}}"
```

nprizal (commented):
Looks good to me. Can we update the default test command to include both default and json reporter?

mcncl:
Thanks @nprizal! Applied the suggestions

nprizal:
>  The CI check is currently blocked/pending, so this is from the diff rather than a completed build.

Interesting that the last few commits were pushed by @buildkite-systems  🤔  therefore the build is blocked.

mcncl:
@nprizal yeah, that's on me; I was setting something up in another tab and commented on your comments as buildkite-systems 😬

mcncl:
Closing this because of the Buildkite-systems stuff, will re-open a PR that's not as janked

PR #583, opened 2026-07-07
Support selector based splitting for jest, playwright, cypress, cucumber, and pytest

nprizal:
We're moving towards selector based splitting as the general way bktec splits test suites, eventually replacing file based splitting entirely. Each runner decides what a selector means for it. For `go` that's a package, for `custom` it's a user provided list, and `rspec` already added selector support where a selector is a file.

This PR adds the same file-based selector support to `jest`, `playwright`, `cypress`, `cucumber`, and `pytest`. For all of these, a selector is still just a file, so `GetSelectors` delegates straight to `GetFiles`.

Two runners aren't part of this PR:
- `nunit` is going to use class name as the selector instead of file, so it needs its own implementation.
- `pytest-pants` isn't included either. We're likely going to deprecate it soon, so it's not worth adding selector support for now.

For now selector based splitting runs alongside the existing file based splitting, and they need to behave the same way. That's what makes it easy to eventually deprecate the file based path and just use selector based splitting by default.

Note this is still experimental. It requires a feature flag on the server for intelligent test plan, so it won't do anything without that flag enabled.

### Changes

- Added `SplitBySelector: true` to jest, playwright, cypress, cucumber, and pytest's supported features
- Added `GetSelectors` to each of those runners, delegating to `GetFiles`

### DIFF
--- a/internal/command/request_param.go
+++ b/internal/command/request_param.go
@@ -119,9 +119,6 @@ func createRequestParam(ctx context.Context, cfg *config.Config, testTargets []s
 }
 
 func shouldUseSelectorSplitting(cfg *config.Config, runner runner.TestRunner) bool {
-	if cfg.TagFilters != "" {
-		return false
-	}
 	// Pytest tag filters are applied during local example collection.
 	// Selector requests skip that step, so fall back to example splitting.
 	if cfg.TagFilters != "" {
--- a/internal/command/request_param.go
+++ b/internal/command/request_param.go
@@ -122,7 +122,11 @@ func shouldUseSelectorSplitting(cfg *config.Config, runner runner.TestRunner) bo
 	if cfg.TagFilters != "" {
 		return false
 	}
-
+// Pytest tag filters are applied during local example collection.
+// Selector requests skip that step, so fall back to example splitting.
+if cfg.TagFilters != "" {
+	return false
+}
 	return cfg.SelectorSplitting && runner.SupportedFeatures().SplitBySelector
 }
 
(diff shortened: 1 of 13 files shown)


internal/command/request_param.go
```diff
@@ -119,6 +119,10 @@ func createRequestParam(ctx context.Context, cfg *config.Config, testTargets []s
 }
 
 func shouldUseSelectorSplitting(cfg *config.Config, runner runner.TestRunner) bool {
+	if cfg.TagFilters != "" {
+		return false
+	}
+
```
mttns: A comment here would be valuable for context

```suggestion
// Pytest tag filters are applied during local example collection.
// Selector requests skip that step, so fall back to example splitting.
if cfg.TagFilters != "" {
	return false
}
```

mattensor (approved):
Looks good!

internal/command/request_param.go
nprizal: Good idea!

internal/command/request_param.go
```diff
@@ -119,6 +119,11 @@ func createRequestParam(ctx context.Context, cfg *config.Config, testTargets []s
 }
 
 func shouldUseSelectorSplitting(cfg *config.Config, runner runner.TestRunner) bool {
+	// Pytest tag filters are applied during local example collection.
+	// Selector requests skip that step, so fall back to example splitting.
+	if cfg.TagFilters != "" {
```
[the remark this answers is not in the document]
nprizal: yes, if tag filters is enabled we should use that "mode" instead. we won't be splitting by files for this "mode"

PR #586, opened 2026-07-07
Support skipped tests in selector split plans

meghan-kradolfer:
### Description

Supports skipped-test handling when selector splitting is enabled for file-backed runners that already support Test Engine skipping.

For RSpec and Cucumber, selector targets are still discovered as file paths. This change filters those selector-backed files through `filter_tests`; files that need skipped-test filtering are expanded into examples, while the remaining files are sent as selectors for normal selector bin-packing.

### Context

Linear: https://linear.app/buildkite/issue/TE-6199/support-skipped-tests-in-selector-split-test-plans

Related server/API change: https://github.com/buildkite/buildkite/pull/30886

### Changes

- Adds a skipped-test selector-splitting path for file-backed runners with selector, example, file-filtering, and skip support.
- Preserves selector-only request params for non-skipped selector runners such as Go/Jest/etc.
- Extracts shared filter-and-expand logic for file and selector-backed-file flows.
- Makes `./` location-prefix application idempotent so real RSpec example paths are not double-prefixed.
- Adds request-param coverage for filtered selector-backed files, no filtered files, location prefixes, and feature gating.

### DIFF
--- a/internal/command/path.go
+++ b/internal/command/path.go
@@ -3,6 +3,7 @@ package command
 import (
 	"fmt"
 	"path/filepath"
+	"strings"
 )
 
 // prefixFilePaths prepends the given prefix to the file paths of the test cases.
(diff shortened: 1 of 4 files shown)


nprizal (approved):
👍

PR #590, opened 2026-07-10
feat(vitest): add vitest runner

mcncl:
### Description

Currently folks need to use the Jest runner with Vitest, as Vitest supports the Jest shape of JSON, however Vitest is (arguably) becoming more popular and offers some benefits over Jest, like explicit `skipped` status.

### Context

This was raised by a customer as an enhancement on the engine, rather than using Jest.

### Changes

- adds a Vitest test runner which can be used with `BUILDKITE_TEST_ENGINE_RUNNER=vitest`
    - a lot of the runner is similar/the same as Jest, Vitest just handles things like skipping in a more convenient way
- creates a `ParseJestReport` function for Jest and Vitest to share, with more generic erroring (removed `Jest`)
- adds `vitest` to config locations

### Testing

- `go test ./...` with targeted testing passes, full test suite fails locally as it looks to use some CI specific function


nprizal (approved):
Looks good!

For posperity this PR is reopened from https://github.com/buildkite/test-engine-client/pull/580.

mcncl:
Hey @nprizal, would you mind unblocking CI for this?

PR #592, opened 2026-07-13
[TE-6293] Support location prefix in selector split requests

meghan-kradolfer:
### Description

Add the configured location prefix to selector-splitting test plan requests so the API can scope selector timing history correctly in monorepos. Selector values remain unchanged and executable as-is.

### Context

[TE-6293](https://linear.app/buildkite/issue/TE-6293/support-location-prefix-in-bktec-selector-splitting-requests)

### Changes

- Add optional top-level `location_prefix` to test plan request parameters.
- Populate it for both direct selector and hybrid selector/example splitting requests when configured.
- Preserve raw, unprefixed selector values.
- Cover request construction and JSON serialization for RSpec and Go selectors.

### Testing

- `go test ./internal/api`
- `go test ./internal/command -run 'TestCreateRequestParams_(GoTestSelectorSplitting|RSpecSelectorSplitting)' -count=1`

The complete `internal/command` package currently has two environment-dependent Pytest failures because the locally installed Pytest does not recognize `--tag-filters`; the selector-focused tests pass.


nprizal (commented):
~Should we do this for `FilterTests` endpoint as well?~

Apparently we are already sending the prefixed filename to the `FilterTests` endpoint when selector splitting is on. So we don't need to send that again.

nprizal (approved):

meghan-kradolfer:
> ~Should we do this for `FilterTests` endpoint as well?~
> 
> Apparently we are already sending the prefixed filename to the `FilterTests` endpoint when selector splitting is on. So we don't need to send that again.

Yup `filter_tests` receives location-prefixed file paths. We haven't really touched that in the selector work as it only supports file-base runners anyway

mattensor (approved):

PR #596, opened 2026-07-15
[TE-6379] Auto-collect git metadata during bktec run for ordered selection

gchan:
### Description

Auto-collect git metadata during `bktec run` (not only `bktec plan`) so selection prediction can be driven straight from the existing `run` command, with no separate `plan` step. `command.Run` accepted the selection flags but never collected the git commit and diff metadata the plan API needs, so selection could previously only be exercised via `plan`.

The intended use is ordered selection: run the full suite but order it so higher-scoring tests run first. This change does not itself force ordering or drop any tests — it makes the selection options usable from `run`. The behaviour of the resulting plan (ordering versus a score cutoff / deselection) is determined by the strategy and params the caller passes to the API, exactly as with `plan`.

`autoCollectGitMetadata` is called on the plan-creation path in `fetchOrCreateTestPlan`, gated on `cfg.SelectionStrategy != ""`, mirroring `command.Plan`. Collection only runs on a plan-cache miss: when the server already has a cached plan for the identifier, `run` reuses it and skips collection.

Trade-off worth noting for the parallel-node case. A typical `run` has many nodes starting at roughly the same time, each independently inferring its test files and calling `FetchTestPlan`. There is no client-side coordination, so whichever nodes miss the cache before the first `CreateTestPlan` populates it will each run the git-metadata collection and submit their own create request; in the worst case (all nodes start together and all miss) every node runs the git commands. This is acceptable because the cost is low — a handful of quick `git` invocations, dominated by the plan API round-trip and the test run itself. The only potential hotspot is the full `git diff base...HEAD` on very large diffs. `runDiffCommands` has a `skipDiffs` mode that omits the full patch, but it is not reachable here yet: the auto-collection path (`collectDiffMetadata`) always passes `skipDiffs=false`, and the `--skip-diffs` flag is currently only registered on the `backfill-commit-metadata` command. Wiring it into the `run`/`plan` auto-collection is left as follow-up.

`--collect-git-metadata` is intentionally left unwired on `run` for now, so the trigger is the selection strategy alone. The behaviour stays behind the `BKTEC_PREVIEW_SELECTION` preview flag that gates the selection flags on both commands.

### Context

[TE-6379](https://linear.app/buildkite/issue/TE-6379/auto-collect-git-metadata-during-bktec-run-for-ordered-selection)

### Changes

- `internal/command/run.go`: call `autoCollectGitMetadata` on the cache-miss path in `fetchOrCreateTestPlan` when a selection strategy is set.
- `internal/command/run_test.go`: cover the collect and no-collect cases, driven by an injectable fake git runner asserting on the `CreateTestPlan` request body.

### Testing

Backed by specs. AI assisted. `go build ./...`, `go vet ./internal/command`, and `go test .  ./internal/command ./internal/config ./internal/api ./internal/git` all pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # Changelog
 ## Unreleased
+- Automatically collect git commit metadata on `bktec run` when `--selection-strategy` is set, matching `bktec plan`. Ordered selection can run straight from `bktec run` without a separate `plan` step. Preview; gated behind `BKTEC_PREVIEW_SELECTION`.
 - Add first-class support for the [Vitest](./docs/vitest.md) test runner (`BUILDKITE_TEST_ENGINE_TEST_RUNNER=vitest`), including test splitting by file, automatic retries, and muting. Vitest's JSON reporter is Jest-compatible, so results are parsed with the shared Jest report parser.
 - **Breaking:** Remove the `pytest-pants` runner. It had zero real usage over the last 3 months. Use the `custom` runner instead, which now supports selector-based test splitting (see `--selector-splitting` in 2.9.0).
 - **Breaking:** Remove the `nunit` runner. It saw almost no real usage, and had known issues splitting tests by file (result path handling, and class name collisions across namespaces).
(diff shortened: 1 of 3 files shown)


gchan:
@buildsworth-bk you can review and approve an L2, expanding the test selection functionality from plan to run is fine.

PR #598, opened 2026-07-16
[TE-6298] - stamp organization_uuid and suite_uuid into backfill tarball metadata.json from the presigned response

joshiain:
## Summary
- Add `organization_uuid` and `suite_uuid` to archive metadata and bump the metadata schema version to `2`.
- Reuse UUIDs from the presigned upload response when uploading, and fetch the suite by slug when writing locally with `--output`.
- Add coverage for suite lookup, UUID propagation, missing-UUID handling, and tarball metadata contents.
- For the `--output` path where no presigned url request is made, the GetSuite api is called instead to retrieve the OrganizationUUID and SuiteUUID. This maintains `read_suites` only scope requirement for the `--output` flag in the documentation

## Context
- Linear: [TE-6298](https://linear.app/buildkite/issue/TE-6298/bktec-stamp-organization-uuid-and-suite-uuid-into-backfill-tarball)

## Testing
- Added unit coverage for `GetSuite`, backfill UUID wiring, and tarball metadata serialization.
- Verified the backfill flow preserves UUIDs through both upload and local output paths.

### DIFF
--- a/internal/api/get_suite.go
+++ b/internal/api/get_suite.go
@@ -1,0 +1,30 @@
+package api
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"net/url"
+)
+
+// Suite identifies a Test Engine suite and its organization.
+type Suite struct {
+	ID             string `json:"id"`
+	OrganizationID string `json:"organization_id"`
+}
+
+// GetSuite retrieves a suite by slug.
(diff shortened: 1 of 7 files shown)


gchan (approved):
Looks good to me! 🚀

PR #601, opened 2026-07-16
Enable Skip feature for Pytest

nprizal:
### Description

Right now Pytest and Playwright both have `SplitByExample: true` but `Skip: false` in their `SupportedFeatures()`. Skipping tests relies on example-level granularity, which is the same mechanism that split by example uses. So any runner that supports split by example should also support skip. RSpec and Cucumber already follow that pattern.

I originally planned to flip `Skip` to `true` for both Pytest and Playwright. During review we found a complication with Playwright, so this PR only enables skip for Pytest and leaves Playwright for a follow-up.

**Why Playwright is not included**

The same Playwright test can run across multiple projects (for example Chromium and Firefox). Test Engine treats each project run as a separate example with its own scope, but all of them share the same `Path` (`spec.File:spec.Line`). The value we hand to Playwright when we run tests is just that `Path`, which is not project-specific. So if the plan skips only the Chromium example and keeps the Firefox one, we would still invoke Playwright with the shared line selector and rerun the test for every project. Enabling skip safely needs a project-scoped selector first (for example passing `--project`), so Playwright skip is deferred.

**Pytest scope fix**

While enabling skip for Pytest, I hit a scope mismatch that had to be fixed first. Skip works by matching the tests we collect locally (via `GetExamples`) against the results the server has on record. For Pytest those two come from different sources, and they were producing different scopes:

- JUnit XML ingestion records the scope as the raw classname (e.g. `tests.test_auth.TestLogin`).
- Local collection via `pytest --collect-only` was recording the scope as the file path (e.g. `tests/test_auth.py::TestLogin`).

Because the scopes did not match, skip would not line up. So this PR makes local collection derive a JUnit-style scope when we are in JUnit mode, so it matches what ingestion stores. I also dropped the now-unused `scope` return from `pytestNodeIDFromJUnit`.

### Context

[TE-6388](https://linear.app/buildkite/issue/TE-6388/enable-skip-feature-for-pytest-and-playwright)

### Changes

- Set `Skip: true` for Pytest
- Fixed Pytest local collection to derive a JUnit-style scope when running in JUnit mode, so it matches the scope recorded by JUnit XML ingestion
- Kept `Skip: false` for Playwright and added a comment explaining the project-selector limitation

### Testing

Covered by the existing runner unit tests plus a new `TestParsePytestCollectOutput_JUnit` case. The Pytest suite passes locally.

### DIFF
--- a/internal/runner/playwright.go
+++ b/internal/runner/playwright.go
@@ -55,7 +55,7 @@ func (p Playwright) SupportedFeatures() SupportedFeatures {
 		FilterTestByTag: false,
 		AutoRetry:       true,
 		Mute:            true,
-		Skip:            false,
+		Skip:            true,
 		SplitBySelector: true,
 	}
 }
--- a/internal/runner/playwright.go
+++ b/internal/runner/playwright.go
@@ -55,7 +55,13 @@ func (p Playwright) SupportedFeatures() SupportedFeatures {
 		FilterTestByTag: false,
 		AutoRetry:       true,
 		Mute:            true,
-		Skip:            true,
+		// Skip stays disabled for now. The same test can run across multiple
+		// Playwright projects (e.g. Chromium and Firefox) as separate examples
+		// that share the same Path (file:line). The value we hand to Playwright
+		// is just that Path, which is not project-specific, so skipping one
+		// project's example while keeping another's would still rerun both.
+		// Enabling skip needs a project-scoped selector first.
+		Skip:            false,
 		SplitBySelector: true,
 	}
 }
--- a/internal/runner/pytest.go
+++ b/internal/runner/pytest.go
@@ -311,7 +311,7 @@ func (p Pytest) GetExamples(files []string) ([]plan.TestCase, error) {
 		return nil, fmt.Errorf("pytest collection failed: %w", err)
 	}
 
-	return parsePytestCollectOutput(string(output))
+	return p.parsePytestCollectOutput(string(output))
 }
 
 // parsePytestCollectOutput parses the output of `pytest --collect-only -q`
@@ -324,7 +324,7 @@ func (p Pytest) GetExamples(files []string) ([]plan.TestCase, error) {
 //	test_auth.py::test_param[value1]
 //
 //	3 tests collected in 0.05s
-func parsePytestCollectOutput(output string) ([]plan.TestCase, error) {
+func (p Pytest) parsePytestCollectOutput(output string) ([]plan.TestCase, error) {
 	var testCases []plan.TestCase
 
 	for _, line := range strings.Split(output, "\n") {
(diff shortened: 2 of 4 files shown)


internal/runner/pytest.go
```diff
@@ -366,6 +370,23 @@ func mapNodeIDToTestCase(nodeID string) plan.TestCase {
...
+// test-collector reports the scope as the file path (e.g. tests/test_auth.py::TestLogin),
+// while JUnit XML reports the scope as the raw classname (e.g. tests.test_auth.TestLogin).
+// Remove `.py` extension, and replace `::` with `.` from scope.
+// Example:
+//
+//	test_sample.py -> test_sample
+//	tests/test_auth.py::TestLogin -> tests.test_auth.TestLogin
+//	tests/MyTest/test_subtract.py::TestClass -> tests.MyTest.test_subtract.TestClass
+func collectorScopeToJunitScope(scope string) string {
+	classname := strings.Replace(scope, ".py", "", 1)
```
mttns: We replace the first instance of `.py`, if there was a dir with .py this logic would break.

```
scope:  tests/pkg.py/test_auth.py::TestLogin
actual: tests/pkg/test_auth.py::TestLogin
```

Could we split out the file path and use TrimSuffix on that component instead?

internal/runner/playwright.go
```diff
@@ -55,7 +55,7 @@ func (p Playwright) SupportedFeatures() SupportedFeatures {
 		FilterTestByTag: false,
 		AutoRetry:       true,
 		Mute:            true,
-		Skip:            false,
+		Skip:            true,
```
mttns: What happens when the same test runs across multiple Playwright projects/browsers? Chromium and Firefox are separate examples/scopes, but may share the same Path 🤔 What if we wanted to skip the Chromium test but still run Firefox? Would passing the shared path to Playwright run both?

nprizal: The project is included in the `scope`. Test Engine and `bktec` identifiy a test by `scope` and `name`. So test for Chromium and Firefox is identified separately.

nprizal: https://github.com/buildkite/test-engine-client/blob/ce258c0569a7fc124c63c19ca42511546d09a0a6/internal/runner/playwright.go#L263

internal/runner/pytest.go
nprizal: That would break yes. But having a dir with `.py` in a python project would be a very rare case I believe.

I've fixed it to be safe.

internal/runner/playwright.go
nprizal: Ah! I see what you mean.

nprizal: Yeah maybe we shouldn't enable skipping for playwright yet. There is no easy way to do this.

nprizal: I've reverted the Playright `Skip` back to `false`

mattensor (approved):

PR #600, opened 2026-07-16
[TE-6386] Support split-by-example with selector splitting

meghan-kradolfer:
### Description

Preserve split-by-example behavior for Playwright and pytest when selector splitting is enabled. Slow files are expanded into individual examples while remaining files continue to be sent as selectors.

RSpec and Cucumber continue to expand selector-backed files for skipped-test handling, and Playwright/pytest continue to send selectors directly when split-by-example is disabled.

### Context

- [TE-6386](https://linear.app/buildkite/issue/TE-6386)
- [PR #599 review comment](https://github.com/buildkite/test-engine-client/pull/599#discussion_r3592253070)

### Changes

- Expand selector-backed files when either skipped-test handling or split-by-example requires it.
- Add Playwright and pytest coverage for enabled and disabled split-by-example behavior.
- Verify slow files become examples while other files remain selectors.

### Testing

- `go test ./internal/command -run 'TestShouldExpandSelectorFiles|TestCreateRequestParams_SelectorSplittingExpandsSlowFilesForSplitByExampleRunners|TestCreateRequestParams_RSpecSelectorSplitting' -count=1`
- `go test ./internal/api ./internal/config`

The full `go test ./internal/command` run also requires the Buildkite pytest collector locally for its existing tag-filter tests; the focused request-path tests pass.

PR #602, opened 2026-07-16
Support tag filters with selector splitting

nprizal:
### Description

Today `shouldUseSelectorSplitting` returns false whenever tag filters are set, so a pytest job with both `BUILDKITE_TEST_ENGINE_TAG_FILTERS` and `BUILDKITE_TEST_ENGINE_SELECTOR_SPLITTING=true` falls back to the file/example path and gets no selector timings.

We can't just drop that guard. In the selector path, `filterAndSplitSelectorFiles` only expands the filtered (slow or skipped) files into examples and sends the rest as raw selectors. Tag filtering is applied during local example collection (`pytest --collect-only --tag-filters`), so any file sent as a raw selector would run all its tests and ignore the tag filter.

So when tag filters are set, the selector path now expands every file into tag-filtered examples so nothing goes out as a raw unfiltered selector, and I removed the tag-filter guard in `shouldUseSelectorSplitting`. Only pytest supports tag filters today.

This builds on the split-by-example-in-selector-mode work from #600.

### Context

[TE-6418](https://linear.app/buildkite/issue/TE-6418)

### Changes

- Removed the tag-filter guard in `shouldUseSelectorSplitting` so selector splitting runs when tag filters are set.
- Added a tag-filter branch to the selector path that expands all files into tag-filtered examples (`splitAllSelectorFiles`), so no file goes out as a raw selector that would ignore the filter.
- Pulled the selector-branch logic into a `selectorTestParams` helper. The two selector returns shared an identical `api.TestPlanParams` envelope, so folding them into the helper collapsed that duplicate envelope into a single return in `createRequestParam`.

### Testing

Covered by unit tests. Renamed the old `SelectorOptInIgnoredForTagFilters` test to `SelectorSplittingWithTagFilters` since the opt-in is no longer ignored, and it asserts every file expands to tag-filtered examples with no raw selectors.

### DIFF
--- a/internal/command/request_param.go
+++ b/internal/command/request_param.go
@@ -25,29 +25,10 @@ import (
 // Currently only the Pytest runner supports tag filtering.
 func createRequestParam(ctx context.Context, cfg *config.Config, testTargets []string, client api.Client, runner runner.TestRunner) (api.TestPlanParams, error) {
 	if shouldUseSelectorSplitting(cfg, runner) {
-		if shouldFilterAndSplitSelectorFiles(cfg, runner) {
-			testParams, err := filterAndSplitSelectorFiles(ctx, cfg, client, testTargets, runner)
-			if err != nil {
-				return api.TestPlanParams{}, err
-			}
-
-			return api.TestPlanParams{
-				Identifier:     cfg.Identifier,
-				Parallelism:    cfg.Parallelism,
-				MaxParallelism: cfg.MaxParallelism,
-				TargetTime:     cfg.TargetTime.Seconds(),
-				Branch:         cfg.Branch,
-				LocationPrefix: cfg.LocationPrefix,
-				Selection:      buildSelectionParams(cfg.SelectionStrategy, cfg.SelectionParams),
-				Metadata:       cfg.Metadata,
-				Runner:         cfg.TestRunner,
-				Tests:          testParams,
-			}, nil
+		testParams, err := selectorTestParams(ctx, cfg, client, testTargets, runner)
+		if err != nil {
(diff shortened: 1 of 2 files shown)


internal/command/request_param.go
```diff
@@ -129,14 +108,32 @@ func createRequestParam(ctx context.Context, cfg *config.Config, testTargets []s
...
+// When tag filters are set, every file is expanded into tag-filtered examples so that
+// nothing goes out as a raw selector, which would run all of the file's tests and
+// ignore the tag filter. Tag filtering is currently only supported for pytest.
+//
+// Otherwise, slow or skipped files are expanded into examples and the rest are sent as
+// raw selectors, or everything is sent as raw selectors when the runner doesn't need
+// filtered files.
+func selectorTestParams(ctx context.Context, cfg *config.Config, client api.Client, testTargets []string, runner runner.TestRunner) (api.TestPlanParamsTest, error) {
+	if cfg.TagFilters != "" && runner.Name() == "pytest" {
+		return splitAllSelectorFiles(testTargets, runner)
```
mttns, line 125: I understand that we can’t send raw selectors here, since they would run tests that don’t match the tag filter. Since that leaves only tests.examples in the payload, does this path intentionally give up selector timings?

nprizal: Yes, this is intentional . For pytest `--tag-filters`, we filter at individual tests/examples (i.e. tags are test level instead of file level). This means we need to _split_ all files/selector into tests/examples.

meghan-kradolfer (approved):

PR #603, opened 2026-07-17
Send `target_time` and `max_parallelism` on filter_tests

nprizal:
### Description

Right now slow-file filtering runs before the server calculates dynamic parallelism, so the plan job reports a parallelism of 1 and almost nothing gets flagged as slow. The server side of `filter_tests` is being taught to accept top-level `target_time` (numeric seconds) and `max_parallelism` params, and to use them to derive the run's real parallelism for slow file filtering. That server change ships alongside this one.

This PR teaches bktec to include `target_time` and `max_parallelism` in the `filter_tests` request body, using the same values it already sends to the `create` (plan) endpoint. Both are optional on the server, so nothing breaks for older clients.

### Context

- Linear ticket [TE-6423](https://linear.app/buildkite/issue/TE-6423)
- Parent [TE-6422](https://linear.app/buildkite/issue/TE-6422)

### Changes

- Added `target_time` (numeric seconds) and `max_parallelism` (integer) to the `filter_tests` request body, forwarding the same config values already sent to `create`.

### Testing

Covered by unit tests. Extended the `filter_tests` API test to assert the new params appear in the request body, and the request-param test to confirm the config values are forwarded end-to-end.

### DIFF
--- a/internal/api/filter_tests.go
+++ b/internal/api/filter_tests.go
@@ -10,8 +10,10 @@ import (
 )
 
 type FilterTestsParams struct {
-	Files []plan.TestCase   `json:"files"`
-	Env   config.EnvPayload `json:"env"`
+	Files          []plan.TestCase   `json:"files"`
+	Env            config.EnvPayload `json:"env"`
+	MaxParallelism int               `json:"max_parallelism,omitempty"`
+	TargetTime     float64           `json:"target_time,omitempty"`
 }
 
 type FilteredTest struct {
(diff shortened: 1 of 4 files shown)


gchan (approved):
LGTM, are you planning to fetch timings on the filter_tests request or will you simply split files slower than the target time?

nprizal:
The filter tests endpoint already fetch timing for all files, and return files slower than "node average time".
So my plan is to send `target_time` and `max_parallelism` to calculate the "node average time" for dynamic parallelism.

PR #608, opened 2026-07-22
Introduce focused test plan API request types

meghan-kradolfer:
### Description

Introduce focused request types for test plan files and examples, keeping the API request model separate from the broader response and execution-side `plan.TestCase` model.

This is the second reviewable increment of TE-6249. It preserves mixed payloads (files plus examples or selectors plus examples) and existing wire behaviour while preventing file requests from carrying response-only test case fields.

### Context

- [TE-6249](https://linear.app/buildkite/issue/TE-6249/tidy-up-split-by-file-vs-split-by-selector-code-paths)
- Follows [#607](https://github.com/buildkite/test-engine-client/pull/607)

### Changes

- Add focused `TestPlanFile` and `TestPlanExample` API request types.
- Use `TestPlanFile` for `filter_tests` requests.
- Convert runner-produced `plan.TestCase` examples explicitly at the command/API boundary.
- Update request and serialization tests for the focused types, including example `format`.

### Testing

- `go test ./internal/api`
- Focused request construction coverage: `go test ./internal/command -run 'TestCreateRequestParams_(NoFilteredFiles|WithLocationPrefix|RSpecSelectorSplittingExpandsFilteredFiles)|TestCreateRequestParams$'`
- `go test ./internal/api ./internal/command` was also attempted. The API package passed

### DIFF
--- a/internal/api/create_test_plan.go
+++ b/internal/api/create_test_plan.go
@@ -10,9 +10,21 @@ import (
 )
 
 type TestPlanParamsTest struct {
-	Files     []plan.TestCase          `json:"files,omitempty"`
-	Examples  []plan.TestCase          `json:"examples,omitempty"`
+	Files     []TestPlanFile           `json:"files,omitempty"`
+	Examples  []TestPlanExample        `json:"examples,omitempty"`
 	Selectors []TestPlanParamsSelector `json:"selectors,omitempty"`
+}
+
+type TestPlanFile struct {
+	Path string `json:"path"`
+}
+
+type TestPlanExample struct {
+	Format     plan.TestCaseFormat `json:"format,omitempty"`
+	Identifier string              `json:"identifier,omitempty"`
+	Name       string              `json:"name,omitempty"`
+	Path       string              `json:"path"`
+	Scope      string              `json:"scope,omitempty"`
 }
 
 type TestPlanParamsSelector struct {
--- a/internal/api/create_test_plan_test.go
+++ b/internal/api/create_test_plan_test.go
@@ -29,7 +29,7 @@ func TestCreateTestPlan(t *testing.T) {
 			"git_diff": "line1\nline2",
 		},
 		Tests: TestPlanParamsTest{
-			Files: []plan.TestCase{
+			Files: []TestPlanFile{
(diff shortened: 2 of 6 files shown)


pda:
🤖 Review by [Amp](https://ampcode.com) (model: Fable 5)

**Verdict: good change, safe to merge.**

This does what `plan.TestCase`'s own TODO has been asking for ("it's best if we split this into two types") — but at the request boundary, which is the right place to start: response/execution code keeps using `plan.TestCase`, and requests get focused types that can't accidentally carry response-only fields.

**Wire compatibility — verified byte-identical:**

- `TestPlanFile{Path}`: previously files serialized as `plan.TestCase` with only `Path` set; every other field is `omitempty`, so the payload was already just `{"path": "..."}`. No change.
- `TestPlanExample`: keeps `format`, `identifier`, `name`, `path`, `scope` with JSON tags identical to `plan.TestCase` (including `omitempty` placement, and `path` staying required). The dropped fields (`estimated_duration`, `value`, `timing_sample_size`) are never set by any runner's `GetExamples` (checked rspec, pytest, gotest, cucumber, custom) — they only appear on the API *response* side, so they were never serialized in requests anyway.
- All `Examples` construction in non-test code routes through `exampleParamsFromTestCases`; no path bypasses the conversion.

**Testing:** ran `go test ./internal/api ./internal/command ./internal/plan` on the branch. `api` and `plan` pass. The two `internal/command` failures (`TestCreateRequestParams_WithTagFilters`, `..._SelectorSplittingWithTagFilters`) are a local environment issue (`pytest: error: unrecognized arguments: --tag-filters` — missing pytest plugin) and fail identically on `main`; CI is green.

**Minor notes (non-blocking):**

- `TestPlanExample.Format` reuses `plan.TestCaseFormat`, keeping a small `plan` dependency in the request types. Fine for now; worth revisiting if the split ever moves `format` semantics apart.
- The `TestCreateTestPlan_SplitByExample` test now sets/asserts `"format": "example"` — this matches what runners actually produce (they all set `Format`), so it makes the fixture more realistic rather than changing behavior.

pda (approved):
🤖 says it's good 😁

PR #609, opened 2026-07-22
Unify file and selector target processing

meghan-kradolfer:
### Description

Replace the parallel file and selector request transformations with one target-processing pipeline while preserving existing request behaviour.

This is the third reviewable increment of TE-6249 and is stacked on #608. Each target retains both its original runnable value and its prefixed file representation, allowing filtering and example expansion to be shared before the remaining targets are encoded for the active mode.

### Context

- [TE-6249](https://linear.app/buildkite/issue/TE-6249/tidy-up-split-by-file-vs-split-by-selector-code-paths)
- Stacked on [#608](https://github.com/buildkite/test-engine-client/pull/608)
- Follows [#607](https://github.com/buildkite/test-engine-client/pull/607)

### Changes

- Add an internal request target representation containing the raw target and prefixed file path.
- Unify file and selector filtering, example expansion, ordering, and final request encoding.
- Remove the duplicated file/selector splitting helper pairs.
- Preserve raw selector values, prefixed file paths, pytest tag filtering, skipped-test expansion, and file-mode filtering behaviour.
- Add focused coverage for mode encoding, filtered-target ordering, prefix round trips, and Playwright file-mode filtering.

### Testing

- Targeted `internal/command` request construction tests
- `go test ./internal/api ./internal/plan`
- `go vet ./internal/command ./internal/api ./internal/plan`

### DIFF
--- a/internal/command/request_param.go
+++ b/internal/command/request_param.go
@@ -12,6 +12,18 @@ import (
 	"github.com/buildkite/test-engine-client/v2/internal/runner"
 )
 
+type requestMode uint8
+
+const (
+	requestByFile requestMode = iota
+	requestBySelector
+)
+
+type requestTarget struct {
+	value string
+	file  api.TestPlanFile
+}
+
 // createRequestParam generates the parameters needed for a test plan request.
 //
 // For the Rspec, Cucumber, Pytest, and Playwright runners, it fetches test files through the Test Engine API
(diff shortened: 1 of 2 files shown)


pda (commented):
🤖 Review by agent (Amp), model (Fable 5)

**Verdict: good change, safe.** I traced both old code paths against the unified pipeline and believe it's behavior-preserving.

**Why it's safe:**
- File mode flow is unchanged: prefix → `!SplitByExample` short-circuit → pytest tag-filter expansion → filter-and-split, with the same debug output and "files" label.
- Selector mode still sends raw selector values while filtering/expansion operates on prefixed paths — `requestTarget{value, file}` carries both through the pipeline, and `TestEncodeRequestTargets` / `TestFilterAndExpandTargets` cover the round trip explicitly.
- Remaining-target selection switched from index-based matching (`selectors[i]`) to a map lookup on `target.file.Path`; equivalent, since the old code keyed `filteredFilesMap` by the same path.

**One real reordering (benign):** in selector mode, the `!features.SplitByExample` short-circuit now runs *before* the pytest tag-filter check, whereas `selectorTestParams` checked tag filters first. This is only reachable by a selector-capable runner named "pytest" without `SplitByExample` — impossible today (pytest has both) — and the new ordering is arguably more correct, since example expansion requires split-by-example support anyway.

**Verified:** ran `go test ./internal/command ./internal/api ./internal/plan` on the PR head locally. Two failures (`TestCreateRequestParams_WithTagFilters`, `TestCreateRequestParams_SelectorSplittingWithTagFilters`) reproduce identically on the base branch and are local-environment issues (system pytest lacks `--tag-filters`); CI is green.

**Nit (non-blocking):** the `label` argument to `filterAndExpandTargets` is fully derivable from `mode` — could be computed inside the function to shrink the signature.

pda (approved):
🤖 says it's good 😁

PR #613, opened 2026-07-27
Default bktec parallelism to one

pda:
### Description

Default bktec to serial execution when Buildkite does not provide parallel-job environment variables, matching Buildkite's command-step behavior.

### Context

- Investigation: https://ampcode.com/threads/T-019fa5ba-fe24-7309-a1e6-197062d9451b
- The independent invalid-configuration error improvement was split into #614.

### Changes

- Default static parallelism to 1 while preserving explicit flag and environment values.
- Document serial execution as the default.

### Testing

- `go test .`
- `go test ./internal/config`
- `go test ./internal/command` was attempted but cannot complete locally because pytest does not recognize the repository's `--tag-filters` option.

### Deployment

Low risk; no migration. Pipelines that omit `parallelism` begin running serially instead of failing validation.

### Rollback

Revert `f42ee8f` to restore required parallelism.

### Related

- #614


jameshill (approved):
agree defaulting to 1

PR #632, opened 2026-09-03
Clarify usage of bktec and test collectors for uploading

omehegan:
### Description

Clarify docs around the use of bktec vs. a test collector to upload results.

### Context

SUP-8013

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ Its main responsibilities are discovering runnable tests, requesting a test plan
 
 Use `bktec` to split test work across parallel Buildkite jobs using timing data from Test Engine. It can also retry failed tests and mute or skip known failures for runners that support those features, while leaving you in control of the command your test runner executes.
 
-For runner-specific setup, see the [runner guides](#runner-guides).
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -76,6 +76,8 @@ steps:
       BUILDKITE_TEST_ENGINE_UPLOAD_RESULTS: "true"
 ```
 
+This setup does not require a Buildkite Test Collector. If your test suite already uses a collector, remove or turn off its reporter, plugin, or test integration before enabling built-in upload. Using both upload methods records duplicate test executions.
```
nprizal:
```suggestion
This setup does not require a Buildkite Test Collector. If your test suite already uses a collector, set `BUILDKITE_TEST_ENGINE_UPLOAD_RESULTS: false` to avoid double uploads. See Upload test results to Test Engine section below.
```

What do you think of asking folks to disable bktec upload if they already use test collector, instead?

omehegan: Yeah, that seems to make more sense!

nprizal (approved):

PR #633, opened 2026-09-06
Add --plan-out to bktec run to write the executed test plan

pda:
### Description

Let consumers save the plan `bktec run` actually uses instead of fetching it again. bk/bk's plan-aware RSpec verification currently re-requests the cached plan on shard 0 with a fresh OIDC token; saving it in bktec removes that extra request and token handling, and records local fallback runs accurately.

```sh
bktec run --plan-out tmp/test-engine/plan.json
# Or:
BUILDKITE_TEST_ENGINE_PLAN_OUT=tmp/test-engine/plan.json bktec run
```

### Context

- [bk/bk plan-aware RSpec verification](https://github.com/buildkite/buildkite/pull/33801)
- Related to [TE-6949](https://linear.app/buildkite/issue/TE-6949) (this does not complete its broader Austral rollout/evaluation scope).

### Changes

- Add the run-only env binding `BUILDKITE_TEST_ENGINE_PLAN_OUT`; the CLI flag takes precedence. Create parent directories and overwrite the destination before executing tests. File open/write/close failures stop the run with exit 16; otherwise runner exit semantics are unchanged.
- Write the full cached or freshly created API response as indented JSON with a trailing newline, before node-specific prefix trimming or retries. Preserve every server field, including `selection`, `skipped_reason`, and fields not modeled by the execution structs. Every node writes the entire plan; no extra network requests are added.
- For local full-suite fallback splitting, write the actual tasks, configured identifier and parallelism, and explicit `"fallback": true`. Empty server error plans are replaced by the fallback bktec actually executes, not emitted as empty plans.
- `run --plan-out` is file-only: `-` is a literal filename, not stdout, which already carries test output. Existing `plan --plan-out` behavior is unchanged.
- Document the contract in help, README, and CHANGELOG.

Example server-plan contents (all additional API fields are retained):

```json
{
  "identifier": "build-id/step-id",
  "parallelism": 1,
  "tasks": {
    "0": {
      "node_number": 0,
      "tests": [{"format": "selector", "path": "spec/apple_spec.rb", "identifier": "apple-id", "value": "spec/apple_spec.rb"}]
    }
  },
  "selection": {"strategy": "austral", "applied": true, "skipped_reason": null}
}
```

Local fallback uses the same node/task shape, with path-only tests:

```json
{"identifier":"build-id/step-id","parallelism":1,"tasks":{"0":{"node_number":0,"tests":[{"path":"spec/apple_spec.rb"}]}},"fallback":true}
```

### Testing

Set up fixtures with `bin/setup`; Python tests use the worktree's `.venv` and `CI=true`. Tests cover cache hits/misses, both empty-plan fallback paths, API-error fallback, env/flag precedence, both nodes retaining the full response, prefix trimming, write errors preventing execution, and preserving test exit status 7.

Addressed both [independent review findings](https://github.com/buildkite/test-engine-client/pull/633#issuecomment-5557597546): empty HTTP 200 bodies now leave a zero plan on both GET and POST, restoring local fallback with or without `--plan-out`. New API and run-level tests failed before the fix and pass after it. Server-plan output is compared byte-for-byte with the re-indented original, including `1.0`, `1e3`, `-0.0`, a 23-digit integer, `\u00e9`, nulls, unknown fields, key order, and the trailing newline.

- `go build ./...`: exit 0, no output.
- `go vet ./...`: exit 0, no output.
- `golangci-lint fmt --diff` on changed Go files: exit 0, no output (gofmt/goimports).
- `golangci-lint run` v2.5.0 (matching CI): `0 issues.`
- `git diff --check`: exit 0, no output.
- `go run . run --help`: confirmed `--plan-out PATH`, env var, fallback marker, and write-failure behavior.
- `./bin/test`: `DONE 613 tests in 34.775s`

`source .venv/bin/activate; export CI=true; go test ./...`:

```text
ok  	github.com/buildkite/test-engine-client/v3	0.300s
ok  	github.com/buildkite/test-engine-client/v3/internal/api	6.810s
ok  	github.com/buildkite/test-engine-client/v3/internal/command	33.754s
ok  	github.com/buildkite/test-engine-client/v3/internal/config	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/debug	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/git	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/otlprelay	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/packaging	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/plan	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/runner	38.945s
ok  	github.com/buildkite/test-engine-client/v3/internal/upload	(cached)
?   	github.com/buildkite/test-engine-client/v3/internal/version	[no test files]
?   	github.com/buildkite/test-engine-client/v3/util/supported_features	[no test files]
```

Full-suite race check, `source .venv/bin/activate; export CI=true; go test -race ./...`:

```text
ok  	github.com/buildkite/test-engine-client/v3	1.601s
ok  	github.com/buildkite/test-engine-client/v3/internal/api	8.131s
ok  	github.com/buildkite/test-engine-client/v3/internal/command	34.289s
ok  	github.com/buildkite/test-engine-client/v3/internal/config	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/debug	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/git	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/otlprelay	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/packaging	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/plan	(cached)
ok  	github.com/buildkite/test-engine-client/v3/internal/runner	36.377s
ok  	github.com/buildkite/test-engine-client/v3/internal/upload	(cached)
?   	github.com/buildkite/test-engine-client/v3/internal/version	[no test files]
?   	github.com/buildkite/test-engine-client/v3/util/supported_features	[no test files]
```

Opt-in, with no migration required. Rollback: remove the flag/env setting, or revert this change; consumers must retain their previous plan-fetching path until they adopt a release containing it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,7 @@
 # Changelog
+## Unreleased
+- Add `bktec run --plan-out <path>` (env: `BUILDKITE_TEST_ENGINE_PLAN_OUT`) to save the full cached or freshly created API plan before running tests, retaining selection and server-only fields without extra requests. Local fallback output includes the actual tasks and `fallback: true`. Parent directories are created; write failures stop the run.
+
 ## 3.0.0 - 2026-07-31
 - ⚠️ **BREAKING:** Selector-based splitting now replaces file-based splitting for supported runners. The deprecated `--selector-splitting` flag and `BUILDKITE_TEST_ENGINE_SELECTOR_SPLITTING` environment variable are still accepted for upgrade compatibility, but their values no longer affect behavior. Existing bktec v2 releases continue to send file-based requests and remain compatible with file-based splitting; remain on v2 if file-based requests are required. See [Migrating from bktec v2 to v3](./docs/migrating-to-v3.md).
 - The Go module path is now `github.com/buildkite/test-engine-client/v3`.
--- a/README.md
+++ b/README.md
@@ -255,6 +255,50 @@ printed to stderr). Only when the server cannot be reached at all does `bktec`
 fall back to a minimal locally-generated plan; this carries the identifier and
 parallelism but no tasks (it is not a computed split), and is noted on stderr.
 
+### Saving the plan used by a run
+
+`bktec run --plan-out <path>` writes the full plan before running tests, so
+verification tools can compare results against the actual split and selection
+without fetching the plan again or managing another API token:
+
+```sh
+bktec run --plan-out tmp/test-engine/plan.json
+# Or configure the output path through the environment:
+BUILDKITE_TEST_ENGINE_PLAN_OUT=tmp/test-engine/plan.json bktec run
+```
+
+The flag takes precedence over `BUILDKITE_TEST_ENGINE_PLAN_OUT`. Parent
+directories are created and existing files are overwritten. If the file cannot
+be created, written, or closed, the run stops before executing tests (exit 16).
+Otherwise test execution and exit statuses are unchanged. Unlike `bktec plan`,
+`run --plan-out` accepts only file paths: `-` names a literal file, not stdout.
+
+For cached and freshly created plans, the file is the API's full `test_plan`
+JSON response, indented with a trailing newline. All server fields are retained,
+including `identifier`, `parallelism`, all nodes in `tasks`, test `format`,
+`path`, `identifier`, and `value`, and `selection`, `skipped_reason`, and timing
+metadata when present. It is saved before node-specific location-prefix
+adjustments or retries. Every parallel node writes the entire plan without any
+additional network requests; consumers can upload the file from node 0.
+
+When bktec falls back to local splitting (including an empty server error plan),
+the file instead records the actual local tasks with an explicit `fallback: true`:
+
+```json
+{
+  "identifier": "build-id/step-id",
+  "parallelism": 2,
+  "tasks": {
+    "0": {"node_number": 0, "tests": [{"path": "spec/apple_spec.rb"}]},
+    "1": {"node_number": 1, "tests": [{"path": "spec/banana_spec.rb"}]}
+  },
+  "fallback": true
+}
+```
+
+This is a local full-suite split, not a server selection. Verification tools
+should check `fallback` before interpreting the document as a server plan.
+
 ### Selector-based test splitting
 
 By default, `bktec` discovers tests and requests a plan by sending runner-specific **selectors**, the values Test Engine looks up when computing a split for the current job. For every runner except gotest, the selector is the same file path that bktec discovers, so selector splitting doesn't change which tests run where, only how that path is reported and matched. gotest is the exception: its selector is a Go package import path from `go list`, and selector splitting replaces the legacy even-count package split with duration-aware splitting, so Go users may see a different (and better balanced) split once historical timing data is available.
--- a/cli.go
+++ b/cli.go
@@ -464,6 +464,13 @@ var planIdentifierFlag = &cli.StringFlag{
 	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_PLAN_IDENTIFIER"),
 }
 
+var runPlanOutFlag = &cli.StringFlag{
+	Name:        "plan-out",
+	Usage:       "write the full test plan as JSON to file `PATH` before running tests, creating parent directories; local fallback plans include fallback: true. Write failures stop the run. Unlike 'plan --plan-out', '-' is a literal file path",
+	Destination: &cfg.PlanOut,
+	Sources:     cli.EnvVars("BUILDKITE_TEST_ENGINE_PLAN_OUT"),
+}
+
 // `plan` command flags
 var maxParallelismFlag = &cli.IntFlag{
 	Name:        "max-parallelism",
@@ -684,6 +691,7 @@ func runCommandFlags() []cli.Flag {
 		filesFlag,
 		tagFiltersFlag,
 		planIdentifierFlag,
+		runPlanOutFlag,
 	}
 	flags = append(flags, buildEnvironmentFlags...)
 	flags = append(flags, testEngineFlags...)
--- a/cli_test.go
+++ b/cli_test.go
@@ -1,7 +1,9 @@
 package main
 
 import (
+	"bytes"
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"net/http"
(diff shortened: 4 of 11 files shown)


pda:
🤖 Independent review by Amp thread https://ampcode.com/threads/T-01a07576-924c-7628-80ca-34f453b180ef (dispatched by Paul via manager thread T-01a06527-1550-73c3-b5f1-223ec56d298f)

**Verdict: needs changes — one small, contained fix; otherwise ship.** The `--plan-out` contract itself holds up: I confirmed byte-level preservation of the API response, the fallback shape, placement before trimming/retries, flag-over-env precedence, `-` as a literal filename, and exit 16 on write failure, all against the built binary and a fake server. The one material finding is a behaviour change on the `run` path when `--plan-out` is *not* set.

### Findings

**1. Medium — `bktec run` now fails fatally (exit 16) on a `200` with an empty body; previously it warned and fell back.** Flag-independent regression.

- `internal/api/client.go:354` only decodes when `len(responseBody) > 0`, so an empty body leaves the target untouched. Before this PR `FetchTestPlan` decoded straight into a `plan.TestPlan`, so an empty body yielded a zero plan → `len(Tasks)==0` → `warnErrorPlan()` → local fallback. `run` used `CreateTestPlan` for the POST path, same behaviour.
- Now `FetchTestPlanRaw` (`internal/api/fetch_test_plan.go:25,41`) decodes into a nil `json.RawMessage` and then `json.Unmarshal(raw, &testPlan)` fails with `unexpected end of JSON input`. `run` also switched to `CreateTestPlanRaw` (`internal/command/run.go:590`), which has the same shape (`internal/api/create_test_plan.go:79,90`). `handleError` doesn't classify that error, so it becomes `❌ Unexpected error` → exit 16, and tests never run.
- Verified with binaries built from the merge-base (`647b6ae`) and the PR head against a server returning `200` + empty body, no `--plan-out`:
  - base: `⚠️ Error Plan … Falling back to non-intelligent splitting` → `Running tests` → **exit 0**
  - head: `❌ Unexpected error: parsing test plan: unexpected end of JSON input` → **exit 16**, no tests run
- Likelihood is low (the API shouldn't send an empty 200), but the PR's stated contract is zero behaviour change without the flag, and this turns a soft fallback into a hard job failure for every `bktec run` user.
- **Fix:** in both `FetchTestPlanRaw` and `CreateTestPlanRaw`, mirror the client's guard — only `json.Unmarshal` when `len(raw) > 0`, otherwise return a zero `plan.TestPlan` (and the empty `raw`). `run` then hits the existing empty-tasks fallback, and `--plan-out` correctly writes the fallback document. Add a unit test in `internal/api` for the empty-body case on both functions. (`plan --json`/`plan --plan-out` already had this behaviour via `CreateTestPlanRaw`, so the fix also tightens those.)

**2. Low (test gap) — `TestRunPlanOut` can't prove the "unmodified JSON" claim.** `cli_test.go:138,140` use `assert.JSONEq`, which compares parsed values, so it would pass even if numbers were re-formatted (`1.0`→`1`, big ints → float64), `null`s dropped, or key order changed. The implementation is right (`json.RawMessage` + `json.Indent` copy literals verbatim — I verified `1.0`, `1e3`, `-0.0`, a 23-digit int, `\u00e9`, `null` and a duplicate key all survive byte-for-byte), but nothing pins it. **Fix:** in the cached/created cases, also assert `string(contents)` equals `json.Indent`ed `serverPlan` + `"\n"` (or include a literal like `"fut":1.0` in `server_only` and `assert.Contains` on the raw bytes).

Nothing else material. Specifically checked and found fine:
- **Write point** (`run.go:77–90`): before `printSplitSummary`, node task lookup, `trimTaskLocationPrefix` (which mutates the shared `*plan.Task`), and `runTestsWithRetry`. Server plans use the raw bytes so later mutation can't leak in; fallback marshals `testPlan.Tasks` before any mutation and trimming is skipped for fallback anyway. A node cannot write a plan different from what it executes.
- **Fallback path** (`run.go:612`): all three routes (cached error plan, created error plan, API error/timeout/billing) produce `raw == nil` + `Fallback: true` and write `{identifier, parallelism, tasks, "fallback": true}` with path-only tests, matching README/PR examples. Uses `cfg.Identifier`/`cfg.Parallelism`, the same inputs `CreateFallbackPlan` used.
- **Errors**: `writeRunPlan` errors aren't `*exec.ExitError`, so `logErrorAndExit` maps them to 16; write/close errors joined; run aborts before tests. Partial file may remain on a mid-write failure, but the exit code makes that safe for consumers.
- **Flag/env** matches `planIdentifierFlag`'s pattern exactly; `plan`'s `planOutFlag` has no env source and is a separate flag object, so `BUILDKITE_TEST_ENGINE_PLAN_OUT` in a pipeline env doesn't leak into `bktec plan`. Relative paths → `MkdirAll(".")` is a no-op. `os.Create` follows symlinks and truncates, same as `plan --plan-out`.
- **Help/README/CHANGELOG** accurate; `## Unreleased` matches prior convention.
- **No extra network calls**: `planRequests` assertion in the test, and confirmed 1 GET (cached) / 1 GET + 1 POST (miss) with the fake server.

### Checks run (worktree at 9c83962)

- `go build ./...` — exit 0
- `go vet ./...` — exit 0
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.5.0 run` — `0 issues.` (locally installed v2.0.2 refuses go1.25 configs)
- `CI=true go test -count=1 . ./internal/api ./internal/command ./internal/config ./internal/plan` (with the `.venv` on PATH for pytest) — all `ok`
- `go test -race -count=1 . ./internal/api ./internal/command` — all `ok`, no data races
- Manual binary runs against a fake Test Engine server: cached plan (byte-identical after de-indenting), error-plan fallback with `--plan-out -` (wrote a file literally named `-`), `--plan-out <dir>` (exit 16, tests not run), flag-over-env (env path never created), empty-body regression above.

### Not verified

- `internal/runner` package tests (need `bin/setup` fixtures: node_modules, Playwright, Cypress, bundler) — untouched by this PR.
- Whether the POST response and the subsequent cached GET response are byte-identical on the server side; bktec writes whatever it receives, so "every node writes the same file" holds only if the API serialises both identically.
- The bk/bk consumer (buildkite/buildkite#33801) honouring `fallback: true`.

pda:
🤖 Fixed both findings from the [independent review](https://github.com/buildkite/test-engine-client/pull/633#issuecomment-5557597546) in [b600f02dd7f60b878617fee429391a1d1adc903f](https://github.com/buildkite/test-engine-client/commit/b600f02dd7f60b878617fee429391a1d1adc903f): empty GET/POST 200 responses again trigger local fallback, with regression coverage both with and without --plan-out. Server-plan tests now assert exact re-indented bytes for numeric spellings, Unicode escapes, nulls, key order, and unknown fields. All 613 tests, full race tests, build, vet, and lint pass.

pda:
🤖 @codex review

Head: b600f02dd7f60b878617fee429391a1d1adc903f

meghan-kradolfer (approved):

PR #640, opened 2026-09-16
Show how bktec works with a Mermaid flow diagram

ashmckenzie:
### Description

Replace the README’s linear text flow with a compact six-node Mermaid overview and a collapsed detailed diagram immediately below it. The detailed flow covers client/server responsibilities, fallback planning, optional uploads, and retries without crowding the default view.

### Context

The initial diagram was too large for a README overview. Separate follow-up commits simplify the default view and restore the original detail behind a clearly labelled disclosure.

### Testing

Both Mermaid diagrams were rendered and visually inspected. Verified that the detailed diagram exactly matches the original commit. Rendered the updated section through GitHub’s Markdown API and checked in Chromium that the disclosure starts collapsed and expands on click with its diagram present. `git diff --check` passes. No runtime code changed.


ashmckenzie:
Hmm, c9e3e08cb55c8de38d06f4265827b1ecd362ac3a is probably a little too big, despite containing more detail?

malclocke (commented):
Looks good. It does push the fold down quite a bit, I wonder if we could move this section further down the page?

ashmckenzie:
> Looks good. It does push the fold down quite a bit, I wonder if we could move this section further down the page?

Thanks @malclocke, that's fair :+1: I updated the initial/visible chart to now flow horizontally which reduces the fold impact. WDYT?

malclocke (approved):

